PR #7, opened 2025-03-20
docs: flesh out documentation

james-garner-canonical:
Add docstrings to public methods and classes and generate reference documentation. Users are expected to use the `PathProtocol` type annotations when writing substrate agnostic code -- this means they'll be seeing the `PathProtocol` docstrings in their editors a fair bit. `ContainerPath` docstrings are also likely to be surfaced in editors, and probably especially when debugging -- same for `LocalPath` but it mostly just inherits from `pathlib.PosixPath` so there's docstrings there.

Make some docs configuration changes to support this.

See the reference docs build [here](https://canonical-pathops--7.com.readthedocs.build/reference/pathops/).

There's one runtime change, where documenting a method revealed that one error case hadn't been handled, so this PR also fixes that.

The PR is already a bit long and wordy, so cleaning up the comments documenting the reasons for not including several methods in the protocol will be left till the next PR.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -385,7 +385,7 @@ autodoc_default_options = {
         '__weakref__,'
         # string methods
         '__repr__,'
-        '__str__',
+        '__str__,'
         # comparison methods
         '__eq__,'
         '__ge__,'
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -5,9 +5,6 @@ import sys
 
 sys.path.insert(0, str(pathlib.Path(__file__).parent.parent / 'src/charmlibs'))  # So that sphinx.ext.autodoc can find code.
 
-def setup(app):
-    app.add_css_file('project_specific.css')
-
 # don't automatically add parentheses after function and method references
 add_function_parentheses = False
 
@@ -183,6 +180,7 @@ html_context = {
 # Template and asset locations
 
 html_static_path = [".sphinx/_static"]
+html_css_files = ["project_specific.css"]
 templates_path = [".sphinx/_templates"]
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -370,6 +370,7 @@ autodoc_default_options = {
         '__init__,'
         '__module__,'
         '__parameters__,'
+        '__repr__,'
         '__subclasshook__,'
         '__weakref__,'
     ),
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -5,6 +5,9 @@ import sys
 
 sys.path.insert(0, str(pathlib.Path(__file__).parent.parent / 'src/charmlibs'))  # So that sphinx.ext.autodoc can find code.
 
+def setup(app):
+    app.add_css_file('project_specific.css')
+
 # don't automatically add parentheses after function and method references
 add_function_parentheses = False
 
@@ -180,7 +183,6 @@ html_context = {
 # Template and asset locations
 
 html_static_path = [".sphinx/_static"]
-html_css_files = ["project_specific.css"]
 templates_path = [".sphinx/_templates"]
 
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -364,6 +364,11 @@ autoclass_content = 'class'
 #autodoc_member_order = 'groupwise'
 autodoc_member_order = 'bysource'
 
+# don't evaluate default values
+# autodoc_preserve_defaults = True
+
+add_module_names = False
+
 autodoc_default_options = {
     'members': None,  # None here means "yes"
     'special-members': None,  # meaning all
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -8,6 +8,9 @@ sys.path.insert(0, str(pathlib.Path(__file__).parent.parent / 'src/charmlibs')) 
 def setup(app):
     app.add_css_file('project_specific.css')
 
+# don't automatically add parentheses after function and method references
+add_function_parentheses = False
+
  # not worthwhile because it doesn't replace references in function signatures
  # open issue: https://github.com/sphinx-doc/sphinx/issues/10785
  # autodoc_type_aliases = {'StrPathLike': 'str | os.PathLike[str]'}
@@ -357,9 +360,9 @@ autoclass_content = 'class'
 # alphabetical (value 'alphabetical'), by member type (value
 # 'groupwise') or by source order (value 'bysource'). The default is
 # alphabetical.
-autodoc_member_order = 'alphabetical'
+#autodoc_member_order = 'alphabetical'
 #autodoc_member_order = 'groupwise'
-#autodoc_member_order = 'bysource'
+autodoc_member_order = 'bysource'
 
 autodoc_default_options = {
     'members': None,  # None here means "yes"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -381,6 +381,8 @@ autodoc_default_options = {
         '__repr__,'
         '__subclasshook__,'
         '__weakref__,'
+        '__annotations__,'
+        '__protocol_attrs__,'
     ),
     'undoc-members': None,
     'show-inheritance': None,
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -357,9 +357,9 @@ autoclass_content = 'class'
 # alphabetical (value 'alphabetical'), by member type (value
 # 'groupwise') or by source order (value 'bysource'). The default is
 # alphabetical.
-#autodoc_member_order = 'alphabetical'
+autodoc_member_order = 'alphabetical'
 #autodoc_member_order = 'groupwise'
-autodoc_member_order = 'bysource'
+#autodoc_member_order = 'bysource'
 
 autodoc_default_options = {
     'members': None,  # None here means "yes"
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -373,16 +373,25 @@ autodoc_default_options = {
     'members': None,  # None here means "yes"
     'special-members': None,  # meaning all
     'exclude-members': (
+        '__annotations__,'
         '__abstractmethods__,'
         '__dict__,'
+        '__hash__,'
         '__init__,'
         '__module__,'
         '__parameters__,'
-        '__repr__,'
+        '__protocol_attrs__,'
         '__subclasshook__,'
         '__weakref__,'
-        '__annotations__,'
-        '__protocol_attrs__,'
+        # string methods
+        '__repr__,'
+        '__str__',
+        # comparison methods
+        '__eq__,'
+        '__ge__,'
+        '__gt__,'
+        '__le__,'
+        '__lt__,'
     ),
     'undoc-members': None,
     'show-inheritance': None,
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -15,6 +15,7 @@
 r""":mod:`pathlib`-like interface for local and :class:`ops.Container` filesystem paths.
 
 The ``pathops`` charm library provides:
+
 - :class:`PathProtocol`: defines the interface of methods common to both local and container paths.
   Use this to type annotate code designed to work on both Kubernetes and machine charms.
 - :class:`ContainerPath`: the concrete implementation of the interface for remote paths in the
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -15,13 +15,13 @@
 r""":mod:`pathlib`-like interface for local and :class:`ops.Container` filesystem paths.
 
 The ``pathops`` charm library provides:
-- :class:`PathProtocol`: defines the interface of methods common to both local and container paths.
+* :class:`PathProtocol`: defines the interface of methods common to both local and container paths.
     Use this to type annotate code designed to work on both Kubernetes and machine charms.
-- :class:`ContainerPath`: the concrete implementation of the interface for remote paths in the
+* :class:`ContainerPath`: the concrete implementation of the interface for remote paths in the
     workload container of Kubernetes charms. Operations are implemented using the Pebble file API.
-- :class:`LocalPath`: the concrete implementation of the interface for local paths, which includes
+* :class:`LocalPath`: the concrete implementation of the interface for local paths, which includes
     both machine charms and the charm container of Kubernetes charms.
-- Top-level helper functions such as :func:``ensure_contents``, which operate on both container
+* Top-level helper functions such as :func:``ensure_contents``, which operate on both container
     and local paths.
 
 :class:`ContainerPath` methods that interact with the remote filesystem will raise a
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -42,6 +42,12 @@ This library also provides the following functions:
     ``StrPathLike`` is a type alias for :class:`str` | :class:`os.PathLike`\[:class:`str`].
     This allows path arguments to be specified as strings or as path-like objects, such as
     :class:`pathlib.PurePath`. :class:`ContainerPath` is **not** :class:`os.PathLike`.
+
+The constants defining the default file and directory permissions are not exported by this package,
+but are documented here to explain the otherwise opaque numbers appearing as default arguments.
+
+.. autodata:: pathops._constants.DEFAULT_MKDIR_MODE
+.. autodata:: pathops._constants.DEFAULT_WRITE_MODE
 """
 
 from __future__ import annotations
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -39,7 +39,6 @@ This library also provides the following functions:
   or :class:`os.PathLike` object, and ensures that the requested contents are available at that
   path, writing and setting ownership and permissions as needed.
 
-
 .. note::
     ``StrPathLike`` is a type alias for :class:`str` | :class:`os.PathLike`\[:class:`str`].
     This allows path arguments to be specified as strings or as path-like objects, such as
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -15,14 +15,14 @@
 r""":mod:`pathlib`-like interface for local and :class:`ops.Container` filesystem paths.
 
 The ``pathops`` charm library provides:
-* :class:`PathProtocol`: defines the interface of methods common to both local and container paths.
-    Use this to type annotate code designed to work on both Kubernetes and machine charms.
-* :class:`ContainerPath`: the concrete implementation of the interface for remote paths in the
-    workload container of Kubernetes charms. Operations are implemented using the Pebble file API.
-* :class:`LocalPath`: the concrete implementation of the interface for local paths, which includes
-    both machine charms and the charm container of Kubernetes charms.
-* Top-level helper functions such as :func:``ensure_contents``, which operate on both container
-    and local paths.
+- :class:`PathProtocol`: defines the interface of methods common to both local and container paths.
+  Use this to type annotate code designed to work on both Kubernetes and machine charms.
+- :class:`ContainerPath`: the concrete implementation of the interface for remote paths in the
+  workload container of Kubernetes charms. Operations are implemented using the Pebble file API.
+- :class:`LocalPath`: the concrete implementation of the interface for local paths, which includes
+  both machine charms and the charm container of Kubernetes charms.
+- Top-level helper functions such as :func:``ensure_contents``, which operate on both container
+  and local paths.
 
 :class:`ContainerPath` methods that interact with the remote filesystem will raise a
 :class:`PebbleConnectionError` if the workload container isn't reachable.
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -17,7 +17,7 @@ r""":mod:`pathlib`-like interface for local and :class:`ops.Container` filesyste
 :class:`ContainerPath` implements a ``pathlib.Path`` style interface for path operations on
 a Juju charm's workload container.
 This interface is defined in :class:`PathProtocol`, which defines a subset of ``pathlib.Path``
-methods that can be implemented via Pebble. The file creation methods are extended with ownership
+methods that Pebble supports. The file creation methods are extended with ownership
 and permissions arguments, as Pebble sets these on file creation.
 :class:`LocalPath` is a subclass of ``pathlib.Path`` that provides these extended method
 signatures for local filesystem operations.
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -14,40 +14,20 @@
 
 r""":mod:`pathlib`-like interface for local and :class:`ops.Container` filesystem paths.
 
-:class:`ContainerPath` implements a ``pathlib.Path`` style interface for path operations on
-a Juju charm's workload container.
-This interface is defined in :class:`PathProtocol`, which defines a subset of ``pathlib.Path``
-methods that Pebble supports. The file creation methods are extended with ownership
-and permissions arguments, as Pebble sets these on file creation.
-:class:`LocalPath` is a subclass of ``pathlib.Path`` that provides these extended method
-signatures for local filesystem operations.
+The ``pathops`` charm library provides:
+- :class:`PathProtocol`: defines the interface of methods common to both local and container paths.
+    Use this to type annotate code designed to work on both Kubernetes and machine charms.
+- :class:`ContainerPath`: the concrete implementation of the interface for remote paths in the
+    workload container of Kubernetes charms. Operations are implemented using the Pebble file API.
+- :class:`LocalPath`: the concrete implementation of the interface for local paths, which includes
+    both machine charms and the charm container of Kubernetes charms.
+- Top-level helper functions such as :func:``ensure_contents``, which operate on both container
+    and local paths.
 
-.. tip::
-    When writing substrate-agnostic code, use :class:`PathProtocol` in your type annotations
-    for accurate type checking and useful autocompletions.
-    Calling code can then supply a :class:`ContainerPath` or :class:`LocalPath` as needed.
-
-.. warning::
-    :class:`ContainerPath` may raise a :class:`PebbleConnectionError` if the workload container is
-    unreachable. Subtrate-agnostic code may choose to be aware of this, or may leave handling this
-    up to substrate-aware calling code, documenting this if so.
-
-This library also provides the following functions:
-
-- :func:`ensure_contents` operates on a :class:`ContainerPath` or any local filesystem :class:`str`
-  or :class:`os.PathLike` object, and ensures that the requested contents are available at that
-  path, writing and setting ownership and permissions as needed.
-
-.. note::
-    ``StrPathLike`` is a type alias for :class:`str` | :class:`os.PathLike`\[:class:`str`].
-    This allows path arguments to be specified as strings or as path-like objects, such as
-    :class:`pathlib.PurePath`. :class:`ContainerPath` is **not** :class:`os.PathLike`.
-
-The constants defining the default file and directory permissions are not exported by this package,
-but are documented here to explain the otherwise opaque numbers appearing as default arguments.
-
-.. autodata:: pathops._constants.DEFAULT_MKDIR_MODE
-.. autodata:: pathops._constants.DEFAULT_WRITE_MODE
+:class:`ContainerPath` methods that interact with the remote filesystem will raise a
+:class:`PebbleConnectionError` if the workload container isn't reachable.
+Methods designed to work with both local and container paths may handle this error internally,
+or they may leave handling connection errors to the caller, documenting this if so.
 """
 
 from __future__ import annotations
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -23,7 +23,7 @@ and permissions arguments, as Pebble sets these on file creation.
 signatures for local filesystem operations.
 
 .. tip::
-    When writing substrate-agnostic code, use :class:`PathProtocol` in your type annotations,
+    When writing substrate-agnostic code, use :class:`PathProtocol` in your type annotations
     for accurate type checking and useful autocompletions.
     Calling code can then supply a :class:`ContainerPath` or :class:`LocalPath` as needed.
 
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -22,7 +22,7 @@ The ``pathops`` charm library provides:
   workload container of Kubernetes charms. Operations are implemented using the Pebble file API.
 - :class:`LocalPath`: the concrete implementation of the interface for local paths, which includes
   both machine charms and the charm container of Kubernetes charms.
-- Top-level helper functions such as :func:``ensure_contents``, which operate on both container
+- Top-level helper functions such as :func:`ensure_contents`, which operate on both container
   and local paths.
 
 :class:`ContainerPath` methods that interact with the remote filesystem will raise a
--- a/src/charmlibs/pathops/__init__.py
+++ b/src/charmlibs/pathops/__init__.py
@@ -22,11 +22,10 @@ and permissions arguments, as Pebble sets these on file creation.
 :class:`LocalPath` is a subclass of ``pathlib.Path`` that provides these extended method
 signatures for local filesystem operations.
 
-When writing substrate-agnostic code, you can use :class:`PathProtocol` in your type annotations.
-Calling code can then supply a :class:`ContainerPath` or :class:`LocalPath` as needed.
-Alternatively, user-facing substrate-agnostic code might accept a union of :class:`ContainerPath`
-and :class:`str` or :class:`os.PathLike` objects, converting these to :class:`LocalPath` objects
-internally if needed.
+.. tip::
+    When writing substrate-agnostic code, use :class:`PathProtocol` in your type annotations,
+    for accurate type checking and useful autocompletions.
+    Calling code can then supply a :class:`ContainerPath` or :class:`LocalPath` as needed.
 
 .. warning::
     :class:`ContainerPath` may raise a :class:`PebbleConnectionError` if the workload container is
--- a/src/charmlibs/pathops/_container_path.py
+++ b/src/charmlibs/pathops/_container_path.py
@@ -143,9 +143,8 @@ class ContainerPath:
     def is_absolute(self) -> bool:
         """Return True if the path is absolute (has a root), which is always the case.
 
-        .. note::
-            Always ``True``, since initialising a :class:`ContainerPath` with a non-absolute
-            path will result in a :class:`RelativePathError`.
+        Always ``True``, since initialising a :class:`ContainerPath` with a non-absolute
+        path will result in a :class:`RelativePathError`.
         """
         return self._path.is_absolute()
 
--- a/src/charmlibs/pathops/_container_path.py
+++ b/src/charmlibs/pathops/_container_path.py
@@ -649,7 +649,10 @@ class ContainerPath:
     #############################
 
     def with_segments(self, *pathsegments: str | os.PathLike[str]) -> Self:
-        """Return a new ContainerPath with the same container, with its entire path replaced.
+        """Construct a new ``ContainerPath`` (with the same container) from path-like objects.
+
+        You can think of this like a copy of the current :class:`ContainerPath`, with its path
+        replaced by ``pathlib.Path(*pathsegments)``.
 
         This method is used internally by all :class:`ContainerPath` methods that return new
         :class:`ContainerPath` instances, including :meth:`parent` and :meth:`parents`. Therefore,
--- a/src/charmlibs/pathops/_container_path.py
+++ b/src/charmlibs/pathops/_container_path.py
@@ -58,6 +58,10 @@ class ContainerPath:
 
     Raises:
         RelativePathError: If instantiated with a relative path.
+
+    Comparison methods compare by path. A :class:`ContainerPath` is only comparable to another
+    object if it is also a :class:`ContainerPath` on the same :class:`ops.Container`. If this is
+    not the case, then equality is ``False`` and other comparisons are :class:`NotImplemented`.
     """
 
     def __init__(self, *parts: str | os.PathLike[str], container: ops.Container) -> None:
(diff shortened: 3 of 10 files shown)


docs/conf.py
```diff
@@ -5,6 +5,15 @@
 
 sys.path.insert(0, str(pathlib.Path(__file__).parent.parent / 'src/charmlibs'))  # So that sphinx.ext.autodoc can find code.
 
+def setup(app):
+    app.add_css_file('project_specific.css')
```
dwilding, line 9: Yesterday I discovered that conf.py has a variable called `html_css_files`, which you should be able to use to achieve the same thing (I haven't tested it yet).

docs/index.md
```diff
@@ -1,8 +1,8 @@
 # Pathops
 
 ```{toctree}
-:maxdepth: 2
-:hidden: true
+:maxdepth: 3
+:hidden: false
```
dwilding, line 5: Does this actually make a difference to anything?

src/charmlibs/pathops/__init__.py
```diff
@@ -12,7 +12,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Pathlib-like interface for local and ops.Container filesystem paths."""
+r""":mod:`pathlib`-like interface for local and :class:`ops.Container` filesystem paths.
+
+:class:`ContainerPath` implements a ``pathlib.Path`` style interface for path operations on
+a Juju charm's workload container.
+This interface is defined in :class:`PathProtocol`, which defines a subset of ``pathlib.Path``
+methods that can be implemented via Pebble. The file creation methods are extended with ownership
```
dwilding, line 20: Suggest rewording to avoid "via":

```suggestion
methods that Pebble supports. The file creation methods are extended with ownership
```

src/charmlibs/pathops/__init__.py
```diff
@@ -12,7 +12,43 @@
...
+.. note::
+    ``StrPathLike`` is a type alias for :class:`str` | :class:`os.PathLike`\[:class:`str`].
+    This allows path arguments to be specified as strings or as path-like objects, such as
+    :class:`pathlib.PurePath`. :class:`ContainerPath` is **not** :class:`os.PathLike`.
+
+The constants defining the default file and directory permissions are not exported by this package,
+but are documented here to explain the otherwise opaque numbers appearing as default arguments.
+
+.. autodata:: pathops._constants.DEFAULT_MKDIR_MODE
+.. autodata:: pathops._constants.DEFAULT_WRITE_MODE
```
dwilding, line 50: I found this a bit difficult to read, mainly because of the various admonitions. How about restructuring it like below. I suggest removing the sentence about the constants and explaining this in the docstrings of the constants (there are only two constants, so I'm OK with the repetition, and I think it would help make the intro more focused).

```suggestion
When writing substrate-agnostic code, use :class:`PathProtocol` in your type annotations
for accurate type checking and useful autocompletions.
Calling code can then supply a :class:`ContainerPath` or :class:`LocalPath` as needed.

.. warning::
    :class:`ContainerPath` may raise a :class:`PebbleConnectionError` if the workload container is
    unreachable. Subtrate-agnostic code may choose to be aware of this, or may leave handling this
    up to substrate-aware calling code, documenting this if so.

This library also provides the following functions:

- :func:`ensure_contents` operates on a :class:`ContainerPath` or any local filesystem :class:`str`
  or :class:`os.PathLike` object, and ensures that the requested contents are available at that
  path, writing and setting ownership and permissions as needed.

``StrPathLike`` is a type alias for :class:`str` | :class:`os.PathLike`\[:class:`str`].
This allows path arguments to be specified as strings or as path-like objects, such as
:class:`pathlib.PurePath`.

:class:`ContainerPath` is not :class:`os.PathLike`.

.. autodata:: pathops._constants.DEFAULT_MKDIR_MODE
.. autodata:: pathops._constants.DEFAULT_WRITE_MODE
```

dwilding: Here's how I'd do the docstrings for the constants:

```
"""Pebble's default of 0o755 (493) 'drwxr-xr-x'.

Pathlib's default value is 0o777 (511) 'drwxrwxrwx'.

The package doesn't export this constant.
It's included here to make the documentation clearer.
"""
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -57,34 +74,73 @@ def __init__(self, *parts: StrPathLike, container: ops.Container) -> None:
     #############################
 
     def __hash__(self) -> int:
+        """Hash on (container-name, path) for efficiency."""
```
dwilding, line 77:
```suggestion
        """Hash the tuple (container-name, path) for efficiency."""
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -57,34 +74,73 @@ def __init__(self, *parts: StrPathLike, container: ops.Container) -> None:
...
+        :class:`ContainerPath` was instantiated with.
+        """
         return self._path.__str__()
 
     def as_posix(self) -> str:
+        """Return the string representation of the path in the container."""
         return self._path.__str__()
 
     def __lt__(self, other: Self) -> bool:
+        """Compare paths only if other is a ContainerPath on the same container.
```
dwilding, line 105:
```suggestion
        """Compare paths only if other is a :class:`ContainerPath` on the same container.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -57,34 +74,73 @@ def __init__(self, *parts: StrPathLike, container: ops.Container) -> None:
...
+        To sort collections of mixed paths, consider using :class:`str` as the key to sort
+        purely by the path, or :func:`repr` to sort primarily by path class, secondarily by path,
+        and thirdly by container name.
+        """
         if not self._can_compare(other):
             return NotImplemented
         return self._path < other._path
 
     def __le__(self, other: Self) -> bool:
+        """Compare paths only if other is a ContainerPath on the same container.
```
dwilding, line 116:
```suggestion
        """Compare paths only if other is a :class:`ContainerPath` on the same container.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -57,34 +74,73 @@ def __init__(self, *parts: StrPathLike, container: ops.Container) -> None:
...
+        To sort collections of mixed paths, consider using :class:`str` as the key to sort
+        purely by the path, or :func:`repr` to sort primarily by path class, secondarily by path,
+        and thirdly by container name.
+        """
         if not self._can_compare(other):
             return NotImplemented
         return self._path <= other._path
 
     def __gt__(self, other: Self) -> bool:
+        """Compare paths only if other is a ContainerPath on the same container.
```
dwilding, line 127:
```suggestion
        """Compare paths only if other is a :class:`ContainerPath` on the same container.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
         return isinstance(other, ContainerPath) and other._container.name == self._container.name
 
     def __eq__(self, other: object, /) -> bool:
+        """Compare paths if other is a ContainerPath on the same container, else return False.
```
dwilding, line 152:
```suggestion
        """Compare paths if other is a :class:`ContainerPath` on the same container, else return False.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
...
+
+        To explicitly compare paths, consider using :class:`str` to get the canonical path string,
+        or create a :class:`ContainerPath` with the same container using :meth:`with_segments`.
+        """
         if not isinstance(other, ContainerPath) or self._container.name != other._container.name:
             return False
         return self._path == other._path
 
     def __truediv__(self, key: StrPathLike) -> Self:
         """Return a new ContainerPath with the same container and the joined path.
```
dwilding, line 162:
```suggestion
        """Return a new :class:`ContainerPath` with the same container and the joined path.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
...
-        If the pattern is relative, matching is done from the right, otherwise the
-        entire path is matched. The recursive wildcard '**' is not supported.
+        .. warning::
+            Only the path is matched against, the container is not considered.
         """
         return self._path.match(path_pattern)
 
     def with_name(self, name: str) -> Self:
-        """Return a new ContainerPath with the same container and the filename changed."""
+        """Return a new ContainerPath, with the same container, but with the path name replaced.
```
dwilding, line 194:
```suggestion
        """Return a new :class:`ContainerPath`, with the same container, but with the path name replaced.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -132,52 +210,82 @@ def with_suffix(self, suffix: str) -> Self:
...
+            path = ContainerPath('/', 'foo', 'bar.txt.zip', container=container)
+            repr(path.with_suffix('.tar.gz'))
+            # "ContainerPath('/foo/bar.txt.tar.gz', container=<ops.Container 'c'>)"
+            repr(path.with_suffix(''))
+            # "ContainerPath('/foo/bar.txt', container=<ops.Container 'c'>)"
         """
         return self.with_segments(self._path.with_suffix(suffix))
 
     def joinpath(self, *other: StrPathLike) -> Self:
         r"""Return a new ContainerPath with the same container and the new args joined to its path.
```
dwilding, line 226:
```suggestion
        r"""Return a new :class:`ContainerPath` with the same container and the new args joined to its path.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
...
 
-        Note that the right hand operand here cannot be a ContainerPath.
-        This is because ContainerPath objects only support absolute paths currently,
-        so using one as the right hand side is likely to be an error.
-        For the same reason, __rtruediv__ is undefined, meaning a ContainerPath cannot
-        be the right hand side operand for arbitrary string or PathLike objects either.
+        .. warning::
+            ``__rtruediv__`` is not supported, as :class:`ContainerPath` only supports absolute
+            paths. Since an absolute path as the right hand operand will completely replace the
+            left hand path, this is mostly likely to be a user error, so it isn't supported.
```
dwilding, line 169: Maybe rephrase the second sentence to be something like "You likely wouldn't want to provide an absolute path as the right-hand operand, because the absolute path would completely replace the left-hand path."

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
...
+
+        .. note::
+            Always ``True``, since initialising a :class:`ContainerPath` with a non-absolute
+            path will result in a :class:`RelativePathError`.
+        """
         return self._path.is_absolute()
 
     def match(self, path_pattern: str) -> bool:
-        """Whether the patch matches the given pattern.
+        """Return true if this path matches the given pattern.
```
dwilding, line 183:
```suggestion
        """Return True if this path matches the given pattern.
```

dwilding: Similarly for `match` in `PathProtocol`

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
...
         """
         return self.with_segments(self._path, key)
 
     def is_absolute(self) -> bool:
-        """Whether the path is absolute. Will always be True, as relative paths error on init."""
+        """Return True if the path is absolute (has a root), which is always the case.
+
+        .. note::
+            Always ``True``, since initialising a :class:`ContainerPath` with a non-absolute
+            path will result in a :class:`RelativePathError`.
```
dwilding, line 178: I'd probably simplify to:

```suggestion
       Initialising a :class:`ContainerPath` with a non-absolute
       path will result in a :class:`RelativePathError`.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
...
-        """Whether the patch matches the given pattern.
+        """Return true if this path matches the given pattern.
+
+        If the pattern is relative, matching is done from the right; otherwise, the entire
+        path is matched. The recursive wildcard ``'**'`` is **not** supported by this method.
 
-        If the pattern is relative, matching is done from the right, otherwise the
-        entire path is matched. The recursive wildcard '**' is not supported.
+        .. warning::
+            Only the path is matched against, the container is not considered.
```
dwilding, line 189:
```suggestion
           This method only matches the path. The container is not considered.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -459,4 +683,11 @@ def mkdir(
     #############################
 
     def with_segments(self, *pathsegments: StrPathLike) -> Self:
+        """Return a new ContainerPath with the same container, with its entire path replaced.
+
+        All :class:`ContainerPath` methods returning new :class:`ContainerPath` instances do so by
+        calling this method, including :meth:`parent` and :meth:`parents`. Subclasses can therefore
+        customise the behaviour of all these methods by overriding only this method. The same is
+        true of :class:`pathlib.Path` in Python 3.12+.
```
dwilding, line 691:
```suggestion
        This method is used internally by all :class:`ContainerPath` methods that return new
        :class:`ContainerPath` instances, including :meth:`parent` and :meth:`parents`. Therefore,
        subclasses can customise the behaviour of all such methods by overriding only this method.
        The same is true of :class:`pathlib.Path` in Python 3.12+.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -286,26 +430,74 @@ def _glob(self, pattern: StrPathLike, skip_is_dir: bool = False) -> Generator[Se
...
+            FileNotFoundError: If the path does not exist.
+            ops.pebble.ConnectionError: If the remote container cannot be reached.
+        """
         info = _fileinfo.from_container_path(self)  # FileNotFoundError if path doesn't exist
         return info.group or ''
 
     def exists(self) -> bool:
+        """Whether this path exists.
+
+        Will follow symlinks to determine if the symlink target exists. This means that exists
```
dwilding, line 455:
```suggestion
        Will follow symlinks to determine if the symlink target exists. This means that :meth:`exists`
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -57,34 +74,73 @@ def __init__(self, *parts: StrPathLike, container: ops.Container) -> None:
...
     def as_posix(self) -> str:
+        """Return the string representation of the path in the container."""
         return self._path.__str__()
 
     def __lt__(self, other: Self) -> bool:
+        """Compare paths only if other is a ContainerPath on the same container.
+
+        To sort collections of mixed paths, consider using :class:`str` as the key to sort
+        purely by the path, or :func:`repr` to sort primarily by path class, secondarily by path,
+        and thirdly by container name.
```
dwilding, line 109: It might be good to simplify the sentence structure here, for easier reading (and similar sentences elsewhere). How about:

```suggestion
        To sort collections of mixed paths, consider using :class:`str` as the key. This will sort
        purely by the path. Alternatively, if you use :func:`repr` as the key, the paths will be sorted
        by path class, then by path, then by container name.
```

dwilding (approved):
This is really great work! I'm not sure I read every single word, but I'm finishing the review because my suggestions are getting pretty nitpicky! Most of my suggestions are minor - I don't think there was anything that I struggled to understand. I think my most significant comment is the one about simplifying the structure of the introduction.

benhoyt (commented):
I started reviewing yesterday, but I see David shares the same concern about all the "admonition" boxes in the intro, so I'm going to post my suggestion anyway and see what you and @dwilding thinks of that approach.

src/charmlibs/pathops/__init__.py
```diff
@@ -12,7 +12,43 @@
...
+- :func:`ensure_contents` operates on a :class:`ContainerPath` or any local filesystem :class:`str`
+  or :class:`os.PathLike` object, and ensures that the requested contents are available at that
+  path, writing and setting ownership and permissions as needed.
+
+.. note::
+    ``StrPathLike`` is a type alias for :class:`str` | :class:`os.PathLike`\[:class:`str`].
+    This allows path arguments to be specified as strings or as path-like objects, such as
+    :class:`pathlib.PurePath`. :class:`ContainerPath` is **not** :class:`os.PathLike`.
+
+The constants defining the default file and directory permissions are not exported by this package,
```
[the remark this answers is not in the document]
james-garner-canonical: Unfortunately `autodoc_preserve_defaults = True` gives just the text `_constants.DEFAULT_...` while `False` gives the base ten representation of the integer.

Subclassing `int` sounds like it would probably work, but seems like a bit much.

I'll go with the default showing in base 10, and just manually add the relevant information to the docstrings (e.g. `"defaults to 0o755 (drwxr-xr-x)"`), dropping the inclusion of the constants in the reference docs.

src/charmlibs/pathops/__init__.py
```diff
@@ -12,7 +12,43 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Pathlib-like interface for local and ops.Container filesystem paths."""
+r""":mod:`pathlib`-like interface for local and :class:`ops.Container` filesystem paths.
+
+:class:`ContainerPath` implements a ``pathlib.Path`` style interface for path operations on
```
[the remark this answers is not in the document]
james-garner-canonical: Good call, thanks, this streamlines things a lot. The original (bad) ordering was just following the (alphabetical) order that the classes and functions appear in the reference. This is better, thanks.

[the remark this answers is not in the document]
dwilding: I also think this would be good :+1:

src/charmlibs/pathops/_types.py
```diff
@@ -34,138 +34,450 @@
...
     # comparison methods
-    def __lt__(self, other: Self) -> bool: ...
-    def __le__(self, other: Self) -> bool: ...
-    def __gt__(self, other: Self) -> bool: ...
-    def __ge__(self, other: Self) -> bool: ...
-    def __eq__(self, other: object, /) -> bool: ...
+    def __lt__(self, other: Self) -> bool:
+        """Comparison methods other than equality are only guaranteed for members of the same type.
+
+        Specifically, :class:`ContainerPath` objects are not comparable to other types of objects.
```
benhoyt, line 69: Similar here, I think we should probably just drop this. Or mention it once on the first comparison method or at the class level.

src/charmlibs/pathops/_types.py
```diff
@@ -34,138 +34,450 @@
...
-    def glob(
-        self,
-        pattern: str,  # support for _StrPath added in 3.13
-        # *,
-        # # TODO: move this information to the docstring
-        # case_sensitive: bool = False,  # added in 3.12
-        # recurse_symlinks: bool = False,  # added in 3.13
-    ) -> Generator[Self]: ...
+        .. note::
+            In Python 3.13, :meth:`pathlib.Path.glob` added support for ``pattern`` to be an
```
benhoyt, line 314: I think we should drop these Python 3.13 and similar comments -- here and below. I think the fact that it's not mentioned/specified in the protocol is enough.

src/charmlibs/pathops/_types.py
```diff
@@ -34,138 +34,450 @@
...
+            :class:`os.PathLike`\[:class:`str`] instead of just a :class:`str`. This is not
+            required by this protocol.
+
+            The ``case_sensitive`` argument, added in Python 3.12, is also not required -- the
+            default behaviour is case-insensitive matching.
```
benhoyt, line 319: Are you sure? I would think the default is case-sensitive on Unix-like OSes. Per the stdlib pathlib docs:

> By default, or when the case_sensitive keyword-only argument is set to None, this method matches paths using platform-specific casing rules: typically, *case-sensitive on POSIX,* and case-insensitive on Windows. Set case_sensitive to True or False to override this behaviour.

src/charmlibs/pathops/_types.py
```diff
@@ -34,138 +34,450 @@
...
+            group: The name of the group to set for the file.
+
+        Returns: The number of bytes written.
+
+        Raises:
+            FileNotFoundError: if the parent directory does not exist.
+            LookupError: if the user or group is unknown.
+            NotADirectoryError: if the parent exists as a non-directory file.
+            PermissionError: if the user does not have permissions for the operation.
+            UnicodeError: if the provided data is not valid utf-8.
```
benhoyt, line 476: I reckon less is more for these lists. In particular, I don't think UnicodeError needs to be mentioned. Even as is, the list is probably not exhaustive.

src/charmlibs/pathops/_types.py
james-garner-canonical: Good catch, I'll investigate this. Looks like all my test filenames and test match strings were all lowercase ...

src/charmlibs/pathops/_container_path.py
```diff
@@ -202,7 +310,7 @@ def read_text(self, *, newline: str | None = None) -> str:
             FileNotFoundError: if the parent directory does not exist.
             IsADirectoryError: if the target is a directory.
             PermissionError: if the Pebble user does not have permissions for the operation.
-            PebbleConnectionError: if the remote Pebble client cannot be reached.
+            ops.pebble.ConnectionError: if the remote Pebble client cannot be reached.
```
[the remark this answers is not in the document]
james-garner-canonical: > given this is mentioned once at the top, do we need to note this exception on every method?

I thought about this, and figured that it is useful to know which methods might raise this -- all the pure path ones won't. I think it's worth documenting which ones will work with or without an active pebble connection, but maybe it would be better as a line in the docstring than in `Raises:`?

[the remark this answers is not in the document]
james-garner-canonical: > Don't we want to encourage pathops.PebbleConnectionError, seeing we're making it public, and they're importing pathops already?

Good point. I changed it to `ops.pebble.ConnectionError` in the docstring because sphinx wasn't linking to `PebbleConnectionError`, but you've convinced me that I should go back to `PebbleConnectionError` even if I can't get the linking working, since that's the error we want users to catch.

src/charmlibs/pathops/_container_path.py
```diff
@@ -57,34 +74,73 @@ def __init__(self, *parts: StrPathLike, container: ops.Container) -> None:
...
         return self._path.__str__()
 
     def as_posix(self) -> str:
+        """Return the string representation of the path in the container."""
         return self._path.__str__()
 
     def __lt__(self, other: Self) -> bool:
+        """Compare paths only if other is a ContainerPath on the same container.
+
+        To sort collections of mixed paths, consider using :class:`str` as the key to sort
```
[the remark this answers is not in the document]
james-garner-canonical: Would it be too odd if the reference docs included just one of the comparison methods (e.g. `__eq__` and `__lt__`) mentioning that the other methods are supported too?

docs/index.md
james-garner-canonical: I don't think `maxdepth` does in the end, due to to toc not showing more levels, but I'm pretty sure that `:hidden: false` makss the left hand table of contents start expanded when you first visit the page.

src/charmlibs/pathops/_container_path.py
```diff
@@ -57,34 +74,73 @@ def __init__(self, *parts: StrPathLike, container: ops.Container) -> None:
...
+        and thirdly by container name.
+        """
         if not self._can_compare(other):
             return NotImplemented
         return self._path > other._path
 
     def __ge__(self, other: Self) -> bool:
+        """Compare paths only if other is a ContainerPath on the same container.
+
+        To sort collections of mixed paths, consider using the :func:`str` as the key to sort
```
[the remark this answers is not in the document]
james-garner-canonical: Ah yeah it does, thanks -- I caught this in some cases, but clearly not in all

src/charmlibs/pathops/_container_path.py
james-garner-canonical: I'm not sure about links in the first docstring line. Do we have any general guidance here?

In this specific case it also results in the line being too long =(

src/charmlibs/pathops/_container_path.py
james-garner-canonical: While I forgot to capitalise `True`, this is the first sentence of the `pathlib.Path.match` docstring -- no 'False otherwise', which I switched to instead of 'whether ...'.

If it's not worth following pathlib's docstring here, how about 'Return whether the patch matches the given pattern.'? (I think an opening verb of `Return` is more Python style than 'Report'.)

src/charmlibs/pathops/_container_path.py
```diff
@@ -187,13 +295,13 @@ def stem(self) -> str:
     def read_text(self, *, newline: str | None = None) -> str:
         r"""Read a remote file as text and return the contents as a string.
 
-        ..note::
+        .. note::
             Compared to pathlib.Path.read_text, this method drops the encoding and errors args.
-            The encoding is assumed to be 'utf-8', and any errors encountered will be raised.
+            The encoding is assumed to be utf-8, and any errors encountered will be raised.
```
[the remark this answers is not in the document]
james-garner-canonical: Thanks, I'll fix these. If only `str.encode` didn't have a default `encoding` value of `'utf-8'` (which turns out to be completely case insensitive -- `'uTf-8'` is fine too) ...

src/charmlibs/pathops/_container_path.py
james-garner-canonical: What do you think about 'this method' instead of linking to this method? 'this method will return False ...'

src/charmlibs/pathops/_container_path.py
```diff
@@ -422,6 +620,32 @@ def mkdir(
...
+            will not have their ownership set.
+
+        Args:
+            mode: The permissions to set on the created directory. Any parents created will have
+                their permissions set to the default of :data:`_constants.DEFAULT_MKDIR_MODE`.
+            parents: Whether to create any missing parent directories as well. If ``False``
+                (default) and a parent directory does not exist, a :class:`FileNotFound` error will
+                be raised.
+            exist_ok: Whether to error if the directory already exists. If ``False`` (default) and
+                the directory already exists, a :class:`FileExistsError` will be raised.
```
james-garner-canonical, line 637:
```suggestion
            exist_ok: Whether to raise an error if the directory already exists.
                If ``False`` (default) and the directory already exists,
                a :class:`FileExistsError` will be raised.
```

src/charmlibs/pathops/_container_path.py
```diff
@@ -422,6 +620,32 @@ def mkdir(
...
+            These are used to set the ownership of the created directory. Any created parents
+            will not have their ownership set.
+
+        Args:
+            mode: The permissions to set on the created directory. Any parents created will have
+                their permissions set to the default of :data:`_constants.DEFAULT_MKDIR_MODE`.
+            parents: Whether to create any missing parent directories as well. If ``False``
+                (default) and a parent directory does not exist, a :class:`FileNotFound` error will
+                be raised.
+            exist_ok: Whether to error if the directory already exists. If ``False`` (default) and
```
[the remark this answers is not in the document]
james-garner-canonical: Thanks, fixed wording, but needed to adjust linebreaks as this makes the line too long.

src/charmlibs/pathops/_local_path.py
```diff
@@ -110,6 +139,33 @@ def mkdir(
...
+            will not have their ownership set.
+
+        Args:
+            mode: The permissions to set on the created directory. Any parents created will have
+                their permissions set to the default value for this argument.
+            parents: Whether to create any missing parent directories as well. If ``False``
+                (default) and a parent directory does not exist, a :class:`FileNotFound` error will
+                be raised.
+            exist_ok: Whether to error if the directory already exists. If ``False`` (default) and
+                the directory already exists, a :class:`FileExistsError` will be raised.
```
james-garner-canonical, line 156:
```suggestion
            exist_ok: Whether to raise an error if the directory already exists.
                If ``False`` (default) and the directory already exists,
                a :class:`FileExistsError` will be raised.
```

src/charmlibs/pathops/_local_path.py
```diff
@@ -110,6 +139,33 @@ def mkdir(
...
+            These are used to set the ownership of the created directory. Any created parents
+            will not have their ownership set.
+
+        Args:
+            mode: The permissions to set on the created directory. Any parents created will have
+                their permissions set to the default value for this argument.
+            parents: Whether to create any missing parent directories as well. If ``False``
+                (default) and a parent directory does not exist, a :class:`FileNotFound` error will
+                be raised.
+            exist_ok: Whether to error if the directory already exists. If ``False`` (default) and
```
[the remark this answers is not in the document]
james-garner-canonical: Same as other `exist_ok`, thanks

src/charmlibs/pathops/_container_path.py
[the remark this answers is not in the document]
benhoyt: I think that's a good idea.

src/charmlibs/pathops/_container_path.py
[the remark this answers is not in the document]
benhoyt: A "Raises:" definitely seems better than just in the docstring. I still lean towards only documenting it at the top or class level. But I'll leave the final call to your judgement.

james-garner-canonical (commented):
Just batching a few comments.

src/charmlibs/pathops/_container_path.py
```diff
@@ -251,6 +372,29 @@ def iterdir(self) -> typing.Generator[Self]:
...
+        r"""Iterate over this directory and yield all paths matching the provided pattern.
+
+        For example, ``path.glob('*.txt')``, ``path.glob('*/foo.txt')``.
+
+        .. warning::
+            Recursive matching using the ``'**'`` pattern is not supported.
+
+        Args:
+            pattern: The :class:`str` or :class:`os.PathLike` pattern to match against. It must be
+                a relative pattern, that is it cannot begin with ``'/'``.
```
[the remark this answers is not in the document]
james-garner-canonical:
```suggestion
            pattern: The :class:`str` or :class:`os.PathLike` pattern to match against.
                The pattern must be relative, meaning it cannot begin with ``'/'``.
```

docs/conf.py
james-garner-canonical: I looked at it in the docs, seemed like it would work, tried it, it doesn't work.

src/charmlibs/pathops/__init__.py
```diff
@@ -12,7 +12,43 @@
...
+    unreachable. Subtrate-agnostic code may choose to be aware of this, or may leave handling this
+    up to substrate-aware calling code, documenting this if so.
+
+This library also provides the following functions:
+
+- :func:`ensure_contents` operates on a :class:`ContainerPath` or any local filesystem :class:`str`
+  or :class:`os.PathLike` object, and ensures that the requested contents are available at that
+  path, writing and setting ownership and permissions as needed.
+
+.. note::
```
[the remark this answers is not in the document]
james-garner-canonical: The problem here is that sphinx apparently handles type aliases horribly. I'd be happy for it to either expand the alias wherever used (I can't get it to do this in function/method signatures), or to treat it as a class, indexing it in the reference and linking to it from type signatures, but that also seem difficult/impossible.

It's not actually an exported type or anything. You know, the way `StrPathLike` shows in vscode when autocompleting methods isn't great either. I think the best user experience would actually be if I removed the alias altogether and just inlined `str | os.PathLike[str]` everywhere.

src/charmlibs/pathops/__init__.py
james-garner-canonical: I think this is resolved now -- dropped the constants and all the boxes.

src/charmlibs/pathops/__init__.py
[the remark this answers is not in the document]
benhoyt: Agreed, thanks

src/charmlibs/pathops/_container_path.py
dwilding: I'd be totally good with that

src/charmlibs/pathops/_container_path.py
dwilding: Good point. After doing some research, I don't see any standard that advises against links (or markup generally) in the first line. I notice that Ben has a link in [`jubilant.WaitError`](https://canonical-jubilant.readthedocs-hosted.com/reference/jubilant/#jubilant.WaitError), which renders like this in VS Code:

vscode

It's not clickable, but I don't think it looks weird.

However, including links is going to increase the line length, as you say. I think in your case, I would probably go with:

```
Compare paths if other is a ``ContainerPath`` on the same container, else return False.
```

Which should squeeze inside the limit (I hope). In your case, I don't actually see huge value in making `ContainerPath` a link anyway, as `ContainerPath` is so central to pathops.

docs/conf.py
dwilding: Oh :confused:

BTW, later we should upgrade to version 1.0.0 of the starter pack, which moves some things around. Not urgent. I've created https://github.com/canonical/charmlibs-pathops/issues/11 to track.

src/charmlibs/pathops/_container_path.py
james-garner-canonical: Per Ben's advice I'll drop this paragraph.

src/charmlibs/pathops/_container_path.py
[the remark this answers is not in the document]
james-garner-canonical: Good point about `repr` having things in the wrong order for useful sorting. I'm going to drop the paragraph about sorting. I feel like sorting won't come up very much, but if it does we can consider adding some recipes or methods for helping with that. I think ideally you'd want to sort on (class-name, container-name or None, path), but there isn't a public way to get the container name out of a `ContainerPath` at the moment.

src/charmlibs/pathops/_types.py
```diff
@@ -34,138 +34,450 @@
...
-    def __str__(self) -> str: ...
+        .. warning::
+            Python 3.12's :class:`pathlib.Path` adds the ``case_sensitive`` keyword argument
+            (default False), which is not part of this protocol. The default behaviour is always
+            case-insensitive matching.
 
-    # like __str__
-    def as_posix(self) -> str: ...
+        .. note::
+            :class:`ContainerPath` only matches against its path, the container is not considered.
```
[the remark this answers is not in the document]
james-garner-canonical: Thanks, I've aligned these docstrings

src/charmlibs/pathops/_types.py
james-garner-canonical: You're correct that it's case-sensitive by default, thanks for catching this.

src/charmlibs/pathops/_container_path.py
james-garner-canonical: Thanks, I've gone with 'Return whether ...'

src/charmlibs/pathops/_container_path.py
dwilding: I agree - what you have now (below) looks better than what I suggested.

```
"""Return whether the path is absolute (has a root), which is always the case.

Always ``True``, since initialising a :class:`ContainerPath` with a non-absolute
path will result in a :class:`RelativePathError`.
"""
```

src/charmlibs/pathops/_container_path.py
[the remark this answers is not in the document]
james-garner-canonical: I've switched back to `PebbleConnectionError` for the docs. It's a shame but it doesn't automatically link to the exception alias, even though it automatically documents it as an alias. Manually linking to ':class:\`PebbleConnectionError\`' makes no difference. I'm reluctant to try a more heavyweight hyperlinking solution as it will look ugly in the docstrings, so this will do for now.

I'm inclined to leave the exception lists as they are for now, and leave that as the place to document whether a method may raise a `ConnectionError`.

src/charmlibs/pathops/_container_path.py
```diff
@@ -44,6 +44,23 @@ class RelativePathError(ValueError):
 
 
 class ContainerPath:
+    r"""Implementation of :class:`PathProtocol` for Juju Charm workload containers.
+
+    Args:
+        \*parts: :class:`str` or :class:`os.PathLike`.
+        container: :class:`ops.Container` used to communicate with workload container.
+            Required, keyword only argument.
```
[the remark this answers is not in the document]
james-garner-canonical: > With type annotations in the signature, we don't need to repeat the types here.

Thanks, I've dropped the type annotations in the signature.

> Do we need the "required keyword-only part"? I would lean no as that's implied by the signature, though I can see why making it explicit could be helpful.

It is technically redundant information isn't it, but I'm going to leave it in as I'm hoping it's helpful for people getting started with the library.

src/charmlibs/pathops/_container_path.py
james-garner-canonical: Thanks for checking how the hyperlinks look in the editor. I'm going to err on the side of not linking to `ContainerPath` in the initial line of `ContainerPath` methods specifically, for brevity, as it's often already challenging fitting the required information on one line.

src/charmlibs/pathops/_container_path.py
james-garner-canonical:
```
        r"""Return a new ContainerPath with the same container and the new args joined to its path.
```
is at the line length limit already, any suggestions for shortening it?

src/charmlibs/pathops/_container_path.py
james-garner-canonical:
```
        """Return a new ContainerPath, with the same container, but with the path name replaced.
```
can only fit 3 more characters before the line length limit. Any suggestions for rephrasing it?

src/charmlibs/pathops/_container_path.py
```diff
@@ -93,37 +149,59 @@ def _can_compare(self, other: object) -> bool:
...
+        """Return a new ContainerPath, with the same container, but with the path name replaced.
+
+        The name is the last component of the path, including any suffixes.
+
+        ::
+
+            container = self.unit.get_container('c')
+            path = ContainerPath('/', 'foo', 'bar.txt', container=container)
+            repr(path.with_name('baz.bin'))
+            # ContainerPath('/foo/baz.bin', container=<ops.Container 'c'>)"
```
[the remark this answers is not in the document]
james-garner-canonical: Created an issue to track
- #12

src/charmlibs/pathops/_types.py
james-garner-canonical: I've converted all these notes to comments above the relevant `PathProtocol` methods, which I think turned out quite tidy. Having the unsupported argument names and Python versions in the source code should be helpful when bumping our minimum Python version, or aiming for a closer match between `PathProtocol` and newer `pathlib` versions.

src/charmlibs/pathops/_types.py
james-garner-canonical: I guess I would quite like these lists to be at least exhaustive of the known errors that relate to the state of the system with respect to the operation being performed -- the user's arguments, the remote filesystem, permissions, and connections. That way if a user does get some other error type, they know that it's really unexpected.

src/charmlibs/pathops/_container_path.py
dwilding: I think this would work:

```
Return a new ``ContainerPath`` with the same container and the provided path name.
```

src/charmlibs/pathops/_container_path.py
dwilding: How about dropping the second "new":

```
Return a new ``ContainerPath`` with the same container and the args joined to its path.
```

benhoyt (approved):
Looks excellent, thanks.

src/charmlibs/pathops/_types.py
james-garner-canonical: I guess one massive category of exceptions neglected by `Raises` are all the ones arising from incorrect argument types, where we're currently just assuming that the type checker warnings are sufficient

benhoyt: Yes, we've made the conscious in Ops to *not* include "raises TypeError" in the docs. Almost every function can potentially raise a type error, so it just adds noise.

james-garner-canonical: Yeah I agree that's the right call.

PR #8, opened 2025-03-24
chore: restructure repository as a monorepo

james-garner-canonical:
This PR moves the pathops `src` and `tests` directories to `pathops/`. Package specific metadata is moved from the base `pyproject.toml` to a new `pyproject.toml` in `pathops/`, along with the current package specific `README.md`. The location of the `py.typed` file is also corrected to `pathops/src/charmlibs/pathops/py.typed`. The project specific `pyproject.toml` extends the repo level one explicitly to specify files for `ruff` and `pyright`, as well as to set the `pyright` python version to `3.8`.

Repo wide `pyproject.toml` configuration and the `tox.ini` file remain at the top level. The `tox.ini` file has been minimally modified with hardcoded paths to continuing to work for `pathops`, as that is the only package in the monorepo currently. [This separate PR](https://github.com/james-garner-canonical/charmlibs-pathops/pull/2/files) sketches how CI could work for multiple packages, and can be reviewed separately.

`docs` remain at the top level, but renamed to `_docs` to separate and distinguish the directory from package directories like `pathops`, and have been minimally modified to continue to build the `pathops` reference docs. It should be possible to add reference docs for further packages the same way. Subsequent PRs will add repo-level documentation.


benhoyt (commented):
Per discussion, how about we drop "packages", and put other things like "docs" with underscore prefixes, like so:

```
charmtech-charmlibs/_charmlibs/
charmtech-charmlibs/_docs/
charmtech-charmlibs/_templates/
charmtech-charmlibs/pathops/pyproject.toml
charmtech-charmlibs/pathops/[src/charmlibs/pathops]/_constants.py
```

pathops/py.typed
benhoyt, line 1: Does py.typed work here? I would have thought it needed to live in src/charmlibs/pathops?

benhoyt (approved):
Looks good. Just one question about py.typed location.

pathops/py.typed
james-garner-canonical: Ah, looks like you're right!, [PEP 561](https://peps.python.org/pep-0561/) says: "For namespace packages (see [PEP 420](https://peps.python.org/pep-0420/)), the py.typed file should be in the submodules of the namespace, to avoid conflicts and for clarity."

Issue #10, opened 2025-03-26
Support Python 3.10+ `newline` argument for `LocalPath.write_text`


james-garner-canonical:
When type checking `pathops` with Python 3.10+, `LocalPath` is detected as incompatibly overriding the `write_text` method of `pathlib.Path`, because it doesn't support the `newline` argument.

The `newline` argument controls how and whether `'\n'` is translated on write. By default (`None`), it is translated to the platform's line ending, which we can assume is `'\n'` since we inherit from `PosixPath`. The values `''` and `'\n'` explicitly do nothing.

Since translating explicitly to the other permitted values (`\r`, `\r\n`) should be very rare, it should be ok to implement this in a straightforward way that translates the entire string in memory before writing.

```diff
 class LocalPath(pathlib.PosixPath):
     ...
     def write_text(
         self,
         data: str,
         encoding: str | None = None,
         errors: str | None = None,
+        newline: str | None = None,
         *,
         mode: int = _constants.DEFAULT_WRITE_MODE,
         user: str | None = None,
         group: str | None = None,
     ) -> int:
         _validate_user_and_group(user=user, group=group)
+        if newline in ('\r', '\r\n'):
+           data = data.replace('\n', newline)
         bytes_written = super().write_text(data, encoding=encoding, errors=errors)
         _chown_if_needed(self, user=user, group=group)
         self.chmod(mode)
         return bytes_written
```

PR #13, opened 2025-03-28
ci: update task runners and workflows to handle the monorepo structure dynamically

james-garner-canonical:
This PR replaces the current `tox.ini`, which is hardcoded to run the tests for the `pathops` package, with a `justfile` that can be parametrized by package name.

The `justfile` provides two public variables, `python` for setting the Python version, and `package` for specifying which package to run tests for. The monorepo's top-level `pyproject.toml` is updated to specify the development dependencies, allowing `uv run` to install these for linting and tests. To install the package itself for testing and static analysis, an optional dependency group is specified for each package, instructing `uv` to install the package from the local files.

The github `tests` workflow is updated to invoke `just` instead of `tox`, passing the appropriate arguments. It also features a new job which checks which packages have changes, and which test suites they have, so that only tests for the changed packages are run in PRs. If global files like the `justfile` or `pyproject.toml` are changed, then tests are run for all packages regardless of other changes.

Where simple commands are not possible in `tests.yaml` and the `justfile`, scripts are used, in-line bash scripts in the `justfile`, and a separate Python script in `.github` for `tests.yaml`.

Static tests for `pathops` are expected to fail with higher Python versions until the following issue is resolved:
- #10

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -115,20 +115,11 @@ jobs:
       - name: Install Pebble
         run: go install github.com/canonical/pebble/cmd/pebble@${{ matrix.pebble-version }}
 
-      - name: Start Pebble
-        run: |
-          umask 0
-          $HOME/go/bin/pebble run --create-dirs &
-        env:
-          PEBBLE: /tmp/pebble-test
-
       - name: Install uv
         uses: astral-sh/setup-uv@v5
 
       - name: Run Pebble integration tests
         run: uvx --from rust-just just package=${{ matrix.package }} python=${{ matrix.python-version }} pebble
-        env:
-          PEBBLE: /tmp/pebble-test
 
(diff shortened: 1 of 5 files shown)


.github/workflows/tests.yaml
```diff
@@ -8,28 +8,96 @@ on:
...
-        uses: actions/setup-python@v5
         with:
-          python-version: '3.12'
+          fetch-depth: 0
 
-      - name: Install tox
-        run: pip install tox~=4.2
+      - name: Collect changed packages and record which test suites they have
+        id: packages
+        run: |
```
james-garner-canonical, line 25: I don't love this inline bash script. I wonder if it would be better as a separate Python script called by this job.

benhoyt (approved):
Looks like a good start. Can always make further improvements/changes!

PR #14, opened 2025-04-02
docs: Add a security policy

tonyandrewmeyer:
I've made some assumptions here - let me know if you want any of these changed:

* The GitHub security reporting functionality will get turned on (it's not perfect, but it's the best choice for projects hosted on GitHub in my opinion).
* Email to the Canonical security team is also ok (I'm quite biased towards email, but I do feel it's good to have as a backup that doesn't rely on third-party tooling and is very low-tech).
* Security support will be provided for all major versions that are active, where "active" means have had a release in the previous 12 months. This is low burden as long as you're not going to be releasing a lot of major versions. It seems reasonable for users if you're expecting (and supporting) people to update to the latest major version quickly.
* The same 3 day response and 3 month fix timelines that Charm-Tech is using elsewhere.
* There's one security policy that applies to all libs in the monorepo.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -2,7 +2,7 @@
 
 ## Supported versions
 
-Security updates will be released for all major versions that have had releases in the last year.
+Security updates will be released for all major versions from 1.0 onwards that have had releases in the last year.
 
 ## Reporting a vulnerability
 
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,36 @@
+# Security policy
+
+## Supported versions
+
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,36 @@
+# Security policy
+
+## Supported versions
+
+Security updates will be released for all major versions that have had releases in the last year.
```
james-garner-canonical, line 5:
```suggestion
Security updates will be released for all major versions from 1.0 onwards that have had releases in the last year.
```

james-garner-canonical (approved):
Thanks for this!

PR #16, opened 2025-04-03
tests: add Juju integrations tests

james-garner-canonical:
This PR adds simple Juju integration tests for pathops. To test with the in-development version of `pathops`, the `pack.sh` script copies the `pathops` library into the charm before packing.

This PR updates the `justfile` to have separate `pack-k8s`, `pack-vm`, `juju-k8s` and `juju-vm` recipes. These pack the kubernetes or machine charm respectively, and execute the integration tests for the respective substrate.


pathops/tests/integration/juju/utils.py
```diff
@@ -0,0 +1,41 @@
...
+    assert ret.is_file()
+    return ret
+
+
+def deploy(juju: jubilant.Juju, substrate: str) -> str:
+    charm_name = f'test-{substrate}'
+    if substrate == 'kubernetes':
+        juju.deploy(
+            get_packed_charm_path(charm_name),
+            resources={'workload': 'ubuntu:latest'},
```
james-garner-canonical, line 35: Is there a lighter weight image I could use here?

pathops/tests/integration/juju/charms/test-kubernetes/src/common.py
```diff
@@ -0,0 +1 @@
+../../common/src/common.py
```
james-garner-canonical, line 1: These symlinks are replaced with the actual files at packing time. They're included to allow the charm code to type check successfully.

justfile
```diff
@@ -28,7 +28,18 @@ format:
 
 [doc('Run `pyright`, e.g. `just python=3.8 static pathops`.')]
 static package *pyright_args:
-    uv run --group='{{package}}' pyright --pythonversion='{{python}}' {{pyright_args}} '{{package}}'
+    #!/usr/bin/env bash
+    set -xueo pipefail
+    uv pip install packaging
+    if uv run python -c 'from packaging.version import Version; from sys import exit; exit(0 if (Version("{{python}}") < Version("3.10")) else 1)'
```
james-garner-canonical, line 34: Is there a tidier way to compare Python version on the commandline?

tonyandrewmeyer: What about:

```bash
uv run python -c "import sys; exit(0 if sys.version_info >= (3, 10) else 1)"
```

james-garner-canonical: Good idea, but the python version being compared against isn't the running version of python, but rather the version of python that we're asking `pyright` to check compatibility with. Nice that you don't need to import `exit` though, I've updated it with that improvement.

justfile
```diff
@@ -28,7 +28,26 @@ format:
 
 [doc('Run `pyright`, e.g. `just python=3.8 static pathops`.')]
 static package *pyright_args:
-    uv run --group='{{package}}' pyright --pythonversion='{{python}}' {{pyright_args}} '{{package}}'
+    #!/usr/bin/env bash
+    set -xueo pipefail
+    touch '{{package}}'/pyproject.toml  # force uv to use latest changes
```
james-garner-canonical, line 33: This line is essentially a hacky bugfix/workaround due to specifying the local package as a dependency in the monorepo's pyproject.toml --- `uv` will only update the install if the `pyproject.toml` modified time is newer. Irrelevant for CI on github, but very important locally. This is probably a sign that I'm holding `uv` wrong in this project. Certainly it wouldn't be needed if we did an editable install with `uv pip install -e`, and now that we've had to put `uv pip install jubilant` in the `justfile`, maybe the package install could get the same treatment ...

All the other lines of the static recipe that have changed are because of adding `jubilant` and it not supporting Python < 3.10. We want to know if we're type checking 3.10+, and if so, install jubilant and type check everything -- but if not, we need to exclude tests/integration/juju (well only the tests therein, not the charms, but ...) but `pyright` doesn't accept `exclude` or `ignore` args on the commandline.

justfile
```diff
@@ -48,14 +67,11 @@ pebble package +flags='-rA':
...
 
-
-[doc("Run juju integration tests. Requires `juju`.")]
-juju package +flags='-rA': (_coverage package 'integration/juju' flags)
-
 [doc("Use uv to install and run coverage for the specified package's tests.")]
 _coverage package test_subdir +flags='-rA':
     #!/usr/bin/env bash
     set -xueo pipefail
+    touch '{{package}}'/pyproject.toml  # force uv to use latest changes
```
james-garner-canonical, line 74: See above -- doing an editable install might be the correct fix.

benhoyt (changes requested):

pathops/tests/integration/juju/test_pathops.py
```diff
@@ -0,0 +1,48 @@
...
+        tb = traceback.format_exc()
+        pytest.exit(f'Deployment failed due to {e!r}\nTraceback:\n{tb}', returncode=1)
+
+
+def test_ensure_contents(juju: jubilant.Juju, substrate: str):
+    _run(juju, substrate, 'ensure_contents')
+
+
+def test_iterdir(juju: jubilant.Juju, substrate: str):
+    _run(juju, substrate, 'iterdir')
```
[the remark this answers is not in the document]
james-garner-canonical: We should, I haven't really fleshed out the tests themselves yet.

pathops/tests/integration/juju/utils.py
james-garner-canonical: With `hello-world` as the workload, tests (sometimes) fail with inability to connect to pebble -- I guess `hello-world` exits after doing it's thing? I'll just stay on `ubuntu:latest` for now, but a minimal image for this kind of thing would be nice.

justfile
james-garner-canonical: The need for this has been eliminated by Jubilant now supporting Python 3.8.

pathops/tests/integration/juju/test_pathops.py
```diff
@@ -16,33 +16,26 @@
...
-    except BaseException as e:
-        tb = traceback.format_exc()
-        pytest.exit(f'Deployment failed due to {e!r}\nTraceback:\n{tb}', returncode=1)
 
+def test_ensure_contents(juju: jubilant.Juju, charm: str):
+    contents = 'Hello world!'
+    result = juju.run(
+        f'{charm}/0', 'ensure-contents', params={'path': 'file.txt', 'contents': contents}
+    )
+    print(result)
```
benhoyt, line 35: Leftover debugging?

pathops/tests/integration/juju/test_pathops.py
```diff
@@ -16,33 +16,26 @@
...
-
-
-def _run(juju: jubilant.Juju, substrate: str, test_case: str):
-    name = utils.charm_name(substrate)
-    result = juju.run(f'{name}/0', 'test', params={'case': test_case})
-    assert result.success
-    return result
+def test_iterdir(juju: jubilant.Juju, charm: str):
+    result = juju.run(f'{charm}/0', 'iterdir', params={'path': '/'})
+    print(result)
```
benhoyt, line 41: Shouldn't we be asserting *something* on the results?

benhoyt (approved):
Couple of minor questions about the tests. But I like the structure of this now. Looks good!

pathops/tests/integration/juju/utils.py
benhoyt: Ah yes, you're right though -- `hello-world` will print a message and then exit...

PR #19, opened 2025-04-03
feat: support Python 3.10+ newline argument of pathlib.Path.write_text

### LINKED ISSUE #10: Support Python 3.10+ `newline` argument for `LocalPath.write_text`

james-garner-canonical:
Resolves #10

### DIFF
--- a/pathops/src/charmlibs/pathops/_local_path.py
+++ b/pathops/src/charmlibs/pathops/_local_path.py
@@ -134,7 +134,7 @@ class LocalPath(pathlib.PosixPath):
         if newline in ('\r', '\r\n'):
             data = data.replace('\n', newline)
         elif newline not in ('', '\n', None):
-            raise ValueError(f'illegal newline value: {newline}')
+            raise ValueError(f'illegal newline value: {newline!r}')
         bytes_written = super().write_text(data, encoding=encoding, errors=errors)
         _chown_if_needed(self, user=user, group=group)
         self.chmod(mode)
--- a/pathops/src/charmlibs/pathops/_local_path.py
+++ b/pathops/src/charmlibs/pathops/_local_path.py
@@ -87,6 +87,7 @@ class LocalPath(pathlib.PosixPath):
         data: str,
         encoding: str | None = None,
         errors: str | None = None,
+        newline: str | None = None,
         *,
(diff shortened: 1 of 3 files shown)


pathops/src/charmlibs/pathops/_local_path.py
```diff
@@ -124,8 +128,13 @@ def write_text(
             LookupError: if the user or group is unknown.
             NotADirectoryError: if the parent exists as a non-directory file.
             PermissionError: if the local user does not have permissions for the operation.
+            ValueError: if ``newline`` is any value other than those documented above.
         """
         _validate_user_and_group(user=user, group=group)
+        if newline in ('\r', '\r\n'):
+            data = data.replace('\n', newline)
+        elif newline not in ('', '\n', None):
+            raise ValueError(f'illegal newline value: {newline}')
```
benhoyt, line 137:
```suggestion
            raise ValueError(f'illegal newline value: {newline!r}')
```

benhoyt (approved):

PR #27, opened 2025-04-07
feat: add charmlibs package and publication actions

james-garner-canonical:
This PR adds the `charmlibs` package to this repository. This is supposed to be a minimal package that just claims the name on PyPI. It shouldn't be installed, so it emits a warning at import time.

This PR also adds the workflows to publish the package to PyPI and test PyPI. Due to the monorepo structure, PyPI publication is triggered by tags being pushed, while publication to test PyPI is triggered manually.


benhoyt (approved):
Looks reasonable to me.

Issue #11, opened 2025-03-26
Upgrade documentation starter pack to version 1.0.0


dwilding:
Tracking this for later, as a copy of https://github.com/canonical/jubilant/issues/70 (probably best if I do that first).

Version 1.0.0 of the starter pack includes upgrade functionality, which should make it easier for us to stay on the latest version in the future.

See https://github.com/canonical/sphinx-docs-starter-pack/releases/tag/1.0.0

PR #35, opened 2025-04-07
docs: enable spellchecking in ci

james-garner-canonical:


james-garner-canonical:
We'll revisit this when we upgrade to the latest docs starter pack next cycle (see #11).

PR #38, opened 2025-04-11
feat: LocalPath also sets group (if not specified) based on user

### LINKED ISSUE #39: What to do about Pebble's handling of user/group for chown?
There are two issues here:
1. If only the group is provided, Pebble will error
2. If only the user is provided, Pebble will set the group as well, like `chown user:`

Currently, `ContainerPath` but not `LocalPath` will error in case (1), and `ContainerPath` and `LocalPath` will behave differently in case (2).

I think the cleanest solution would be to drop the `group` argument from the API to avoid (1), and have `LocalPath` always set the group based on the user (following Pebble's behaviour) to resolve (2).

---

It's worth noting here too that the API currently doesn't support using ids instead of names for user and group, because it's only in a recent version of Pebble/Juju that passing the user solely by id isn't an error (in older versions, the username is required).

james-garner-canonical:
This PR aligns the behaviour of the file creation methods (`mkdir`, `write_bytes`, `write_text`) when the user argument is provided but the group argument is not.

In this case, Pebble sets the group based on the user (like `chown $user: $path`), while `shutil.chown` by default does not (like `chown $user $path`). Since changing the Pebble behaviour to match `shutil` here would require additional API calls (and possibly creating the directory twice?), and since the `user` and `group` arguments for the file creation methods are an API design that follows Pebble rather than `pathlib`, `LocalPath` is updated to follow Pebble's behaviour here.

This PR updates the unit tests to reflect this change, and adds a collection of integration tests for file ownership.

Documentation of the error case when `group` is provided without `user` for `ContainerPath` has also been added.

Resolves #39

### DIFF
--- a/justfile
+++ b/justfile
@@ -61,7 +61,7 @@ _coverage package test_subdir +flags='-rA':
     export COVERAGE_RCFILE=../pyproject.toml
     DATA_FILE=".report/coverage-$(basename {{test_subdir}})-{{python}}.db"
     uv run --active coverage run --data-file="$DATA_FILE" --source='src' \
-        -m pytest --tb=native -vv '{{flags}}' 'tests/{{test_subdir}}'
+        -m pytest --tb=native -vv {{flags}} 'tests/{{test_subdir}}'
     uv run --active coverage report --data-file="$DATA_FILE"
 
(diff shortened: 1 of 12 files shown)


benhoyt (approved):
Thanks for this, looks good to me.

PR #45, opened 2025-04-14
tests: validate that Pebble and LocalPath.mkdir behaviour aligns

### LINKED ISSUE #40: Consistent behaviour for permissions of directory that already exists
EDIT: It turns out that the Pebble behaviour and the `pathlib` behaviour already match -- if the directory already exists, then neither will update the permissions on `mkdir`. Tests added in #45 to validate this. Original issue below.

---

I believe that currently Pebble will set the permissions regardless of whether or not the file already exists, while `LocalPath` follows `pathlib` in only updating the permissions if the directory is actually created. The behaviour should match.

Option 1: Update `ContainerPath` to check if the directory exists first, and use the existing permissions if so.

Options 2: Update `LocalPath` to manually `chmod` after `mkdir`.

EDIT: see below

~I think Option 1 is the correct choice here because we should try to follow `pathlib`'s behaviour.~

james-garner-canonical:
Also make some driveby improvements to `ContainerPath.mkdir`

Resolves #40

### DIFF
--- a/pathops/src/charmlibs/pathops/_container_path.py
+++ b/pathops/src/charmlibs/pathops/_container_path.py
@@ -607,23 +607,39 @@ class ContainerPath:
             PebbleConnectionError: if the remote Pebble client cannot be reached.
         """
         if parents and not exist_ok and self.exists():
-            raise _errors.raise_file_exists(repr(self))
-        elif not parents and exist_ok and not self.parent.exists():
+            _errors.raise_file_exists(repr(self))
+        if exist_ok and not parents and not self.parent.exists():
             _errors.raise_file_not_found(repr(self.parent))
-        if parents:
+        if parents and mode != _constants.DEFAULT_MKDIR_MODE:
             # create parents with default permissions, following pathlib
-            self._container.make_dir(
+            self._mkdir(
                 path=self._path.parent,
                 make_parents=True,
                 permissions=_constants.DEFAULT_MKDIR_MODE,
             )
+        self._mkdir(
+            path=self._path,
+            make_parents=exist_ok or parents,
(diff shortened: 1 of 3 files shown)


pathops/src/charmlibs/pathops/_container_path.py
```diff
@@ -607,23 +607,39 @@ def mkdir(
             PebbleConnectionError: if the remote Pebble client cannot be reached.
         """
         if parents and not exist_ok and self.exists():
-            raise _errors.raise_file_exists(repr(self))
-        elif not parents and exist_ok and not self.parent.exists():
+            _errors.raise_file_exists(repr(self))
+        if exist_ok and not parents and not self.parent.exists():
             _errors.raise_file_not_found(repr(self.parent))
```
james-garner-canonical, line 612: Just trying to improve readability. Though the `raise _errors.raise...` line was a genuine mistake that needed fixing.

pathops/src/charmlibs/pathops/_container_path.py
```diff
@@ -607,23 +607,39 @@ def mkdir(
...
-                make_parents=exist_ok,  # parents created separately above
-                permissions=mode,
-                user=user if isinstance(user, str) else None,
-                group=group if isinstance(group, str) else None,
+                path=path,
+                make_parents=make_parents,
+                permissions=permissions,
+                user=user,
+                group=group,
             )
```
james-garner-canonical, line 643: This is also a kind of unrelated efficiency + correctness improvement.
1. If we're creating a directory with the default permissions, we don't need a separate API call to make the parents
2. When we create the parents, we need to catch all the same Pebble errors

PR #47, opened 2025-04-15
fix: write methods no longer change existing mode unless it's passed explicitly

### LINKED ISSUE #46: write methods shouldn't change permissions of existing files when mode argument isn't passed explicitly
As was brought up in #40, the current behaviour of both `ContainerPath` and `LocalPath` on `write_{bytes,text}` is to always set the mode -- even setting the default mode if none is provided. I think overriding any existing file permissions with the default when none is provided is a mistake, as `pathlib.Path.write_{bytes,text}(data)` doesn't change any existing file permissions.

james-garner-canonical:
The current behaviour of both `ContainerPath` and `LocalPath` for `write_{bytes,text}` is to always set the mode -- even setting the default mode if none is provided. I think overriding any existing file permissions with the default when none is provided is a mistake, as (e.g.) `pathlib.Path.write_bytes(data)` doesn't change any existing file permissions. This PR modifies these methods so that the permissions are only changed if the `mode` argument is not `None`.

Tests have been expanded to cover the case of writing when the target file already exists, and updated to reflect this new behaviour.

Resolves #46

### DIFF
--- a/pathops/src/charmlibs/pathops/_container_path.py
+++ b/pathops/src/charmlibs/pathops/_container_path.py
@@ -520,6 +520,14 @@ class ContainerPath:
         if isinstance(data, (bytearray, memoryview)):
             # TODO: update ops to correctly test for bytearray and memoryview in push
             data = bytes(data)
+        if mode is None:
+            # if the file already exists, then don't change permissions unless explicitly requested
+            try:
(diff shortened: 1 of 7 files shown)


benhoyt (approved):
This seems reasonable to me, thanks

Issue #24, opened 2025-04-03
Add a way to remove files


james-garner-canonical:
We'd planned to implement `unlink` and `rmdir`, but Pebble is currently incapable of implementing the same semantics. Rather than providing these methods but with different semantics, I'd like to wait and see if we can add the functionality to Pebble, and add some charmcraft or other feature that lets a new library version the depends on Pebble having these semantics declare that it needs a certain Juju version.

In the meantime, a `remove_path` method following the semantics of `ops.Container.remove_path` would be possible. It seems like it will be useful for at least this project's tests -- so it could start out as a private internal helper function and get exercised that way.

james-garner-canonical:
> for older versions of pebble, could you fall back to pebble exec & try to use coreutils in the image if they exist

I don't think this is really an option for us. I've yet to add some bare rock images for integration tests, but ideally `pathops` should continue to work in such a case.

What I've done for now in my integration test charms is implement a substrate aware `remove_path` method on the charms themselves, with Pebble-like semantics (though silently returning if the path already didn't exist, as that fit the needs of test cleanup nicely) -- [kubernetes](https://github.com/canonical/charmtech-charmlibs/blob/main/pathops/tests/integration/juju/charms/kubernetes/src/charm.py#L52), [machine](https://github.com/canonical/charmtech-charmlibs/blob/main/pathops/tests/integration/juju/charms/machine/src/charm.py#L51).

An alternative approach would be for users to provide their own `unlink` and `rmdir` methods instead, with the 'best effort' semantics that best fit their needs (e.g. failing to `unlink` symlinks to directories, or allowing `unlink` to operate on empty directories). This could be done either by subclassing `ContainerPath` and `LocalPath` (and `PathProtocol` too if needed for type annotations), or perhaps providing them on the charm class instead (cleaner imo, as you can continue to use `PathProtocol` unmodified).

I'd prefer not to do 'best effort' `unlink` and `rmdir` in `pathops` itself yet, as I suspect it will make adding `unlink` and `rmdir` with correct semantics tricky if/when Pebble supports it. But I could definitely add a `remove_path` helper with Pebble semantics if that would be useful -- do let me know if you think it would be, or if you think that rolling their own `unlink` and `rmdir` would be better for charms.

james-garner-canonical:
Good news! Turns out we can figure out if things are symlinks after all, so `rmdir` and `unlink` are in the works here #52

PR #50, opened 2025-04-15
feat: add remove_path helper function

james-garner-canonical:
This PR adds a new helper function: `remove_path`. This removes a local or container path following Pebble's file removal semantics. This is necessary because we can't currently get the information necessary to implement `pathlib`'s file removal methods (`unlink` and `rmdir`) with the correct semantics (see #24).

Currently the helper is implemented in a private module and isn't exported publicly. It's used in the machine and Kubernetes charms for the Juju integration tests, which both exercises the function and eliminates the need for the charms to provide their own substrate specific file removal methods.

I believe that this helper function would be worth providing publicly, but we can defer that to a future PR if need be.

The main value that it adds over charms simply rolling their own file removal methods is that it converts Pebble errors to the correct native Python exception types. Charms could call `remove_path` directly, or trivially implement their own `pathlib`-like `unlink` and `rmdir` with the 'best effort' semantics that fit their needs. At it's simplest, they could have both simply call `remove_path`, which will remove anything that those methods should remove -- it'll just also remove things they strictly shouldn't (e.g. `unlink` could remove a(n empty) directory).


james-garner-canonical:
Closing in favour of #52

PR #52, opened 2025-04-17
feat: add ContainerPath.rmdir and unlink methods

### LINKED ISSUE #24: Add a way to remove files

james-garner-canonical:
It turns out that we actually can figure out if a file is a symlink via Pebble (thanks @benhoyt!). This PR uses that knowledge to add the `ContainerPath.rmdir`, `ContainerPath.unlink`, and `ContainerPath.is_symlink` methods.

Resolves #24


pathops/src/charmlibs/pathops/_fileinfo.py
```diff
@@ -56,8 +62,22 @@ def from_container_path(path: ContainerPath) -> pebble.FileInfo:
...
+        info_list = path._container.list_files(path._path.parent, pattern=path.name)
+    except (pebble.APIError, pebble.PathError) as e:
+        msg = repr(path)
+        _errors.raise_if_matches_file_not_found(e, msg=msg)
+        _errors.raise_if_matches_permission(e, msg=msg)
+        raise
+    if not info_list:
+        _errors.raise_file_not_found(repr(path))
+    (info,) = info_list
+    return info
```
benhoyt, line 76: Nit, so take it or leave it, but the intent seems clearer with:

```python
assert len(info_list) == 1  # bad bug in Pebble
return info_list[0]
```

benhoyt (approved):
Nice work, thanks!

pathops/src/charmlibs/pathops/_fileinfo.py
james-garner-canonical: Makes sense, I've done the same for the `_get_fileinfo_directly` method as well.

PR #55, opened 2025-04-24
docs(pathops): extended readme

james-garner-canonical:
This PR extends the readme with getting started information. This will show up in both the repository and on PyPI.

### DIFF
--- a/_docs/index.md
+++ b/_docs/index.md
@@ -16,6 +16,6 @@ If you're new to charms, see {ref}`Juju | Charm <juju:charm>`.
 
 ## Pathops
 
-Pathops is a Python package providing a {doc}`pathlib <python:library/pathlib>`-like interface for both local filesystem and Kubernetes workload container paths. Charms can use [ContainerPath](pathops.ContainerPath) to interact with files in the workload container, or [LocalPath](pathops.LocalPath) to interact with local files using the same API. Code designed to work for both cases can use [PathProtocol](pathops.PathProtocol) in type annotations.
+Pathops is a Python package that provides a {doc}`pathlib <python:library/pathlib>`-like interface for both local filesystem and Kubernetes workload container paths. Charms can use [ContainerPath](pathops.ContainerPath) to interact with files in the workload container, or [LocalPath](pathops.LocalPath) to interact with local files using the same API. Code designed to work for both cases can use [PathProtocol](pathops.PathProtocol) in type annotations.
 
 Pathops is [available on PyPI](https://pypi.org/project/charmlibs-pathops). Read the [Pathops reference documentation](pathops).
--- a/_docs/index.md
+++ b/_docs/index.md
@@ -14,7 +14,7 @@ as well as general documentation about charm libraries.
 
 You can also read our {ref}`guidance on distributing charm libraries as Python packages <how-to-python-package>`.
 
-If you're new charms, see {ref}`Juju | Charm <juju:charm>`.
+If you're new to charms, see {ref}`Juju | Charm <juju:charm>`.
 
 ## Pathops
 
--- a/_docs/index.md
+++ b/_docs/index.md
@@ -18,4 +18,4 @@ If you're new to charms, see {ref}`Juju | Charm <juju:charm>`.
 
 Pathops is a Python package that provides a {doc}`pathlib <python:library/pathlib>`-like interface for both local filesystem and Kubernetes workload container paths. Charms can use [ContainerPath](pathops.ContainerPath) to interact with files in the workload container, or [LocalPath](pathops.LocalPath) to interact with local files using the same API. Code designed to work for both cases can use [PathProtocol](pathops.PathProtocol) in type annotations.
 
-Pathops is [available on PyPI](https://pypi.org/project/charmlibs-pathops). Read the [Pathops reference documentation](pathops).
+Pathops is [available on PyPI](https://pypi.org/project/charmlibs-pathops) with examples of how to get started. You can also read the [full reference documentation](pathops).
--- a/_docs/index.md
+++ b/_docs/index.md
@@ -8,9 +8,7 @@ how-to/index
 reference/index
 ```
 
-This is the Charm Tech team at Canonical's documentation website for charm libraries.
-It hosts the documentation for the team's charm libraries that are distributed as Python packages,
-as well as general documentation about charm libraries.
+This is the Charm Tech team at Canonical's documentation website for charm libraries. It hosts the documentation for the team's charm libraries that are distributed as Python packages, as well as general documentation about charm libraries.
 
 You can also read our {ref}`guidance on distributing charm libraries as Python packages <how-to-python-package>`.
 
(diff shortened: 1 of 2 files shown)


pathops/README.md
```diff
@@ -9,3 +9,80 @@ from charmlibs import pathops
...
+            source=(self.charm_root / 'foo/bar').read_bytes(),
+        )
+        if changed:
+            # e.g. make the workload reload its configuration
+            ...
+```
+
+`pathops.PathProtocol` provides a subset of the `pathlib.Path` API -- relative paths, hardlink and symlink manipulation, and `open` are unsupported.
+
+A separate `chmod` method is also unsupported, as Pebble does not currently support this. `mkdir`, `write_bytes` and `write_text` instead provide the arguments `mode`, `user`, and `group` to set directory or file permissions and ownership. `ensure_contents` also supports these arguments.
```
dwilding, line 88: Some small suggestions for readability:

```suggestion
`pathops` doesn't provide a separate `chmod` method, as Pebble doesn't currently support this. Instead, `mkdir`, `write_bytes`, and `write_text` have arguments `mode`, `user`, and `group` to set directory or file permissions and ownership. `ensure_contents` also has these arguments.
```

_docs/index.md
```diff
@@ -8,22 +8,14 @@ how-to/index
...
 ## Pathops
 
-Pathops is a Python library providing
-a {doc}`pathlib <python:library/pathlib>`-like interface
-for Kubernetes charms
-to interact with files in their workload container.
-Charms can use [ContainerPath](pathops.ContainerPath) to interact with files in the workload container,
-or [LocalPath](pathops.LocalPath) to interact with local files using the same API.
-Code designed to work for both cases can use [PathProtocol](pathops.PathProtocol) in type annotations.
+Pathops is a Python package providing a {doc}`pathlib <python:library/pathlib>`-like interface for both local filesystem and Kubernetes workload container paths. Charms can use [ContainerPath](pathops.ContainerPath) to interact with files in the workload container, or [LocalPath](pathops.LocalPath) to interact with local files using the same API. Code designed to work for both cases can use [PathProtocol](pathops.PathProtocol) in type annotations.
```
dwilding, line 19: Very minor - I'd probably say "that provides" instead of providing:

```suggestion
Pathops is a Python package that provides a {doc}`pathlib <python:library/pathlib>`-like interface for both local filesystem and Kubernetes workload container paths. Charms can use [ContainerPath](pathops.ContainerPath) to interact with files in the workload container, or [LocalPath](pathops.LocalPath) to interact with local files using the same API. Code designed to work for both cases can use [PathProtocol](pathops.PathProtocol) in type annotations.
```

_docs/index.md
```diff
@@ -8,22 +8,14 @@ how-to/index
...
 
 Pathops is [available on PyPI](https://pypi.org/project/charmlibs-pathops). Read the [Pathops reference documentation](pathops).
```
dwilding, line 21: How about adding a reference to the README:

```suggestion
Pathops is [available on PyPI](https://pypi.org/project/charmlibs-pathops). Read the [Pathops reference documentation](pathops) and [examples of how to get started](https://github.com/canonical/charmtech-charmlibs/blob/main/pathops/README.md#getting-started).
```

dwilding (approved):
Nice one, these are great updates!

_docs/index.md
james-garner-canonical: I was wondering if there was a nice way to integrate the readme content into the docs site. Do you know any? Presumably I could include the full readme file somewhere as a page if that would be a good idea.

pathops/README.md
```diff
@@ -9,3 +9,80 @@ from charmlibs import pathops
 ```
 
 Check out the reference docs [here](https://canonical-charmlibs.readthedocs-hosted.com/reference/pathops/).
+
+# Getting started
+
+To get started, consider creating an attribute with the root directory for the local or remote paths you'll be working with.
```
james-garner-canonical, line 15: Thanks, I'll tweak your suggestion to reflect that this is not the only way.
```suggestion
To get started, you can create an attribute `self.root` with the root directory for the local or remote paths you'll be working with.
```

pathops/README.md
```diff
@@ -9,3 +9,80 @@ from charmlibs import pathops
...
+
+# or
+
+class MachineCharm(ops.CharmBase):
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        self.root = pathops.LocalPath('/')
+```
+
+You can then write code that could be used by either charm, using the `pathops.PathProtocol` type, which describes the operations supported by both local and container paths.
```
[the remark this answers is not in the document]
james-garner-canonical: I think I'll stick with what I have for this one, as it hopefully makes the reason to use `pathops.PathProtocol` over `LocalPath | ContainerPath` more clear (though I don't think drilling into this in detail is needed here).

_docs/index.md
dwilding: I'm not aware of a way. I haven't looked in great detail though. Maybe a good approach would be to create a how-to guide (later) based on similar examples. Called something like "How to use pathops to write configuration files"

pathops/README.md
```diff
@@ -9,3 +9,80 @@ from charmlibs import pathops
...
+
+```py
+CONFIG_FILE_LOCATION = '/foo/bar'
+CONFIG_FILE_CONTENTS = '...'
+
+def write_config_file(root: pathops.PathProtocol) -> None:
+    path = root / CONFIG_FILE_LOCATION
+    path.write_text(CONFIG_FILE_CONTENTS)
```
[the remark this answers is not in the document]
james-garner-canonical: I think I'll keep this one as the somewhat more concise original version. I guess the helper function could just be in-lined into `_on_config_changed`, but I figure maybe you'd do this on pebble-ready or some other startup event (the config file doesn't necessarily reflect Juju user exposed config, right?).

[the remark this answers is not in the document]
dwilding: Makes sense. A dedicated how-to guide could cover a more specific/detailed scenario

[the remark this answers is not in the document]
dwilding: Or, conversely, we find the relevant places in the Ops docs and promote pathops there :wink:

_docs/index.md
tonyandrewmeyer: In theory you can use literal-include, like:

```
.. literalinclude:: readme.md
        :language: markdown
        :start-at: foo:
        :end-before: bar:
```

However, (a) I don't know if you can do this in myst-markdown or only in ReST, and (b) I'm not sure if you can reference files that are outside of the root docs directory.

*But*, I don't think it's great to do this anyway. A little repetition or a link is fine, I think, and longer term I would expect that there would be more detailed examples in how-to/reference content with the README only having minimal/short ones (because the purpose of the docs are different).

pathops/README.md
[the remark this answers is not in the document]
james-garner-canonical: Yeah that would be nice!

_docs/index.md
james-garner-canonical: Good point @tonyandrewmeyer , and thanks for the info. I'm thinking I'll go with this:
```suggestion
Pathops is [available on PyPI](https://pypi.org/project/charmlibs-pathops) with examples of how to get started. You can also read the [full reference documentation](pathops).
```

PR #56, opened 2025-04-24
ci: multi-base packing

james-garner-canonical:
This PR adds the option to specify the base when packing. Currently the 20.04 and 24.04 bases are supported and exercised in CI. The charm + `pathops` code is therefore now exercised on both bases in the integration tests run in CI. Most significantly this means that `pathops` is tested with multiple Python versions in integration tests (3.8 and 3.12).


benhoyt (approved):
Thanks. This seems reasonable to me. The duplication is a bit annoying, and I'm sure there are other ways we could do this, but this is direct and simple enough.

james-garner-canonical:
> The duplication is a bit annoying, and I'm sure there are other ways we could do this, but this is direct and simple enough.

Agreed. Maybe in future some templating would be worthwhile to factor out all the repetition across the four `charmcraft.yaml` files, but for now at least you can read the whole `charmcraft.yaml` file for a given substrate and base in one place, and the tests in `test_meta.py` ensure that they're kept in sync.

PR #73, opened 2025-06-10
docs: library listing pages and infra

james-garner-canonical:
This PR adds pages to the charmlibs docs to list charm libraries.

The libraries and their metadata are recorded in `csv` files. This PR currently contains a limited subset of libraries from the following sources (subsequent PRs will add all the relevant libraries from them):
1. running [charm-analysis/tools/summarise_deps.py](https://github.com/tonyandrewmeyer/charm-analysis/blob/main/tools/summarise_dependencies.py)
2. searching for 'charm' on PyPI (all results were manually filtered to charming related libs only, and then further filtered to libraries intended to be used in charms, rather than for testing, managing deployments, etc)
3. the [charmcraft docs](https://canonical-charmcraft.readthedocs-hosted.com/stable/reference/files/libname-py-file/#popular-libraries)
4. running [charm-analysis/tools/summarise_libs.py](https://github.com/tonyandrewmeyer/charm-analysis/blob/main/tools/summarise_libs.py)

This PR adds a local `sphinx` extension, `generate`, which consumes these `csv` files and produces `rst` files with the information in table form. This includes writing some raw html to faciliate sorting cells logically, wrapping links in a class that disables spellchecking, and adding tooltips to the icons used to indicate status (which requires some custom CSS to show).

This PR also adds the `sphinx_datatables` extension, which provides searching and sorting for the generated table.

To support the development and maintenance of this local sphinx extension, some unit tests have also been added for it, and the docs CI has been extended to run unit and static tests for the extensions directory. The repo's linting config has also been updated with more selective ignores so that extensions are covered by linting as well.


dwilding (commented):
I like this! I think my preference would be to try to collapse some of the info in the table, to make it easier to understand at a glance. You can judge whether that's desirable, but here's a simpler model I have in mind:

- Remove ↪️. There's only one lib in this category, and the description already says it's a dependency of other libs.

- Combine 🪦 and 🚫 into a single category with symbol 🚫 and tooltip "Not recommended". I don't find the symbol 🪦 especially intuitive, and in practice is there so much difference between the two categories? The descriptions already say things like legacy / deprecated / used internally, which I think provides sufficient info about the reason a lib is not recommended.

BTW, I haven't finished reviewing yet - I wanted to send my initial comments for you to look at.

_docs/.gitignore
```diff
@@ -23,3 +23,6 @@ package*.json
 __pycache__
 .idea/
 .vscode/
+
+# generated docs
+**/generated/*
```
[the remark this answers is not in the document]
james-garner-canonical: I've gone with `generated/`, which will ignore the `generated` directory as intended, thanks for suggesting simplification here.

_docs/extensions/generate.py
```diff
@@ -0,0 +1,267 @@
...
+# Sphinx extension #
+####################
+
+if typing.TYPE_CHECKING:
+    from typing import Iterable
+
+    import sphinx.application
+
+
+def setup(app: sphinx.application.Sphinx) -> dict[str, str | bool]:
```
dimaqq, line 33: There's a lot of code in this file.
How do you test it?

dimaqq (commented):
Coming up nicely :)

james-garner-canonical:
`Docs` CI is currently failing due to the charmhub.io links for integrations being flaky.

.github/workflows/docs.yaml
```diff
@@ -49,3 +49,11 @@ jobs:
         uses: DavidAnson/markdownlint-cli2-action@v20
         with:
           config: "_docs/.sphinx/.markdownlint.json"
+      - name: Local Sphinx extensions static analysis
+        id: ext-static-step
+        if: success() || failure()
+        run: uvx --from=rust-just just docs ext-static
+      - name: Local Sphinx extensions unit tests
+        id: ext-unit-step
```
dimaqq, line 57: **nitpick:** the id's are redunant.

.github/workflows/docs.yaml
```diff
@@ -49,3 +49,11 @@ jobs:
...
+        if: success() || failure()
```
dimaqq, line 58: The proliferation of `if: success() || failure()` suggests that there ought to be a better way to distribute these commands.

Perhaps separate job would do the trick?

_docs/explanation/libraries.rst
```diff
@@ -0,0 +1,16 @@
+Libraries
+=========
+
+There are two kinds of two kinds of charm library.
```
dimaqq, line 4:
```suggestion
There are two kinds of charm libraries.
```

_docs/explanation/libraries.rst
```diff
@@ -0,0 +1,16 @@
...
+
+#. A library may either be a relation library or not.
+
+    * Relation libraries allow charms to interact with other charms over a defined Juju relation. Using a relation library will therefore require adding a specific relation to your ``charmcraft.yaml`` and integrating your charm with another charm after deploying them both.
+
+    * Non-relation libraries cover all the other use cases for libraries, from general purpose charming helpers to sharing team specific code between charms. These libraries may work with your charm alone, or they may provide functionality that relies on being connected to another charm via a relation library.
+
+#. A library may either be distributed as a Python package, or as a single file module via Charmhub.
```
dimaqq, line 12:
```suggestion
#. A library may be distributed either as a Python package or as a single file module on Charmhub.
```

_docs/explanation/libraries.rst
```diff
@@ -0,0 +1,16 @@
+Libraries
+=========
+
+There are two kinds of two kinds of charm library.
+
+#. A library may either be a relation library or not.
+
+    * Relation libraries allow charms to interact with other charms over a defined Juju relation. Using a relation library will therefore require adding a specific relation to your ``charmcraft.yaml`` and integrating your charm with another charm after deploying them both.
+
+    * Non-relation libraries cover all the other use cases for libraries, from general purpose charming helpers to sharing team specific code between charms. These libraries may work with your charm alone, or they may provide functionality that relies on being connected to another charm via a relation library.
```
dimaqq, line 10: **fixes:** general-purpose, team-specific, Charmhub-hosted

_docs/extensions/generate.py
```diff
@@ -0,0 +1,284 @@
...
+    key_table = _get_status_key_table()
+    msg = ' Library status is shown in the left column. See tooltips, or click here for a key.'
+    content = f'.. dropdown::{msg}\n\n' + '\n'.join('   ' + line for line in key_table.split('\n'))
+    _write_if_needed(path=(gen_dir / 'status-key-table.rst'), content=content)
+
+
+def _write_if_needed(path: pathlib.Path, content: str) -> None:
+    to_write = _FILE_HEADER + content
+    if not path.exists() or path.read_text() != to_write:
+        path.write_text(to_write)
```
dimaqq, line 138: **question:** why not blindly overwrite the file if they are the same? Does some step in the flow depend on file timestamps?

_docs/extensions/generate.py
```diff
@@ -0,0 +1,284 @@
...
+    return f'{main_link} ({extra_links})'
+
+
+def _kind(entry: _CSVRow) -> str:
+    kind = entry['kind']
+    prefix = _hidden_text(_KIND_SORTKEYS[kind])
+    kind_str = _EMOJIS.get(kind, '') + kind
+    if not kind_str:
+        return prefix
+    return f'{prefix}       | {kind_str}'
```
dimaqq, line 213: Is the number of spaces meaningful or important here?

_docs/extensions/test_generate.py
```diff
@@ -0,0 +1,85 @@
...
+from docutils import core
+
+
+def rst_to_html(rst: str) -> str:
+    return core.publish_parts(rst, writer_name='html')['html_body']  # type: ignore
+
+
+##########
+# tables #
+##########
```
dimaqq, line 37: **minor:** I don't think these markers are needed. The test names are sufficient.

_docs/explanation/libraries.rst
james-garner-canonical: It's actually intentional -- but maybe more accurate as 'two kinds of kinds of charm library'? In both cases I'm just trying to be playful. But there's the relation vs non-relation distinction, and the PyPI vs Charmhub distinction. A better way to phrase this?

_docs/.sphinx/_static/project_specific.css
```diff
@@ -6,3 +6,28 @@
...
+  background-color: #EBEBEB;
+  color: #000000;
+  padding: 5px;
+  border-radius: 5px;
+  white-space: nowrap;
+}
+.emoji-div:hover .emoji-tooltip {
+  display: block;
+  z-index: 9999;
+}
```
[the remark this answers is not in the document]
dimaqq: As an implementation detail, let's defer CSS changes to a separate PR.

_docs/extensions/generate.py
dimaqq: A comment would suffice for now.

Maybe a proper RST builder could be used in the future, a 3p library, or some custom code abstractions.

dwilding (approved):
I need to stop for today. Preemptively approving, but I'll continue working through the rest first thing tomorrow if you're willing to wait!

dwilding (approved):
I finished reviewing. It all works well, thanks! I didn't review the content of the lists - saving that for later, as we discussed.

_docs/explanation/libraries.rst
```diff
@@ -0,0 +1,16 @@
+Libraries
+=========
+
+There are two kinds of two kinds of charm library.
+
+#. A library may either be a relation library or not.
+
+    * Relation libraries allow charms to interact with other charms over a defined Juju relation. Using a relation library will therefore require adding a specific relation to your ``charmcraft.yaml`` and integrating your charm with another charm after deploying them both.
```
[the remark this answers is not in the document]
james-garner-canonical: Yes, I greatly dislike how this looks, but I don't know how to change it.

_docs/.gitignore
```diff
@@ -23,3 +23,7 @@ package*.json
 __pycache__
 .idea/
 .vscode/
+
+# generated docs
+generated/
+.~lock.*.csv#
```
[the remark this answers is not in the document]
james-garner-canonical: That's actually what the lockfile created by LibreOffice Calc looks like, but I wonder how `git` parses it. Maybe I should go with a more generic `~lock.*`

_docs/extensions/generate.py
```diff
@@ -0,0 +1,284 @@
...
+
+def _rst_link(name: str, url: str) -> str:
+    return f'`{name.strip()} <{url.strip()}>`__'
+
+
+def _hidden_text(msg: str | int) -> str:
+    return f"""
+       .. raw:: html
+
+          <span style="display:none;" class="hidden-sortkey-text">{msg}</span>
```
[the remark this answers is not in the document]
james-garner-canonical: I definitely see why you're suggesting this. I indeed only added the `hidden-sortkey-text` key to disable the spell checker. I wonder if I should change the ignored class to something like `no-spellcheck` and apply it to fields like the library name ... that would probably reduce the custom wordlist by a lot.

---

The reason I went with inline "display:none;" instead of having that be set via the custom css, is that I'd like the page to render nicely even if the custom css and datatables extension failed to load. For example, currently the custom css only controls whether the tooltips appear on hover. If it's removed, the tooltips are just always hidden. I'm not sure how important this consideration is really, but I also like the locality of setting it inline, so I'll stick with this for now.

---

I'll look into making the unit test a bit more robust ... maybe it I switch to using an `html` parser for that part I can `assert` on the value of `'display'` cleanly instead of matching the string.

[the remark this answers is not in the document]
dwilding: > The reason I went with inline "display:none;" instead of having that be set via the custom css, is that I'd like the page to render nicely even if the custom css and datatables extension failed to load.

That seems like a reasonable justification

+1 on adding a general `no-spellcheck` class that can be applied anywhere it's needed

_docs/explanation/libraries.rst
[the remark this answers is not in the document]
dwilding: Ah, it turns out you need exactly 3 spaces before the `*`s. To make the `*`s exactly line up with the text after `#.` RIP my tab key

.github/workflows/docs.yaml
james-garner-canonical: I'll leave this as is for now since it's incrementally building on the workflow from the docs starterpack.

_docs/extensions/generate.py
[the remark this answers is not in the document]
james-garner-canonical: `no-spellcheck` is proliferated and custom wordlist is pruned. Html is used a bit more, but I'm sticking with checking for the `display:none;` string in unit tests as it doesn't seem trivial to get cleaner than that with the stdlib.

PR #79, opened 2025-06-27
docs: add remaining general libraries to listing

james-garner-canonical:
This PR adds the other known general libraries to the library listing docs. The libraries were found using the methodology described in #73, where only a small number of the known libraries were added to simplify review.

The emojis for the statuses are updated based on discussion, as are status tooltips and library sorting. Key tables for the emojis now only include the emojis used on that page. Libraries with the `'unlisted'` status are filtered out of the table. Charmcraft-hosted libs are prefixed with their charm name, and linebreaks are controlled with injected html. 

A small number of words are added to the custom wordlist.


_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+awsjuju,legacy,https://pypi.org/project/awsjuju/,,https://github.com/kapilt/awsjuju,PyPI,,,Legacy library for managing AWS resources.
 charm-api,experimental,https://pypi.org/project/charm-api/,,https://github.com/canonical/charm-api,PyPI,y,y,Experimental API for writing charms.
+charm-json,legacy,https://pypi.org/project/charm-json/,,https://github.com/canonical/charm-json,PyPI,y,y,JSON typed relation data – unnecessary with ops 2.23+ due to the addition of typed relation data.
+charm-refresh-build-version,team,https://pypi.org/project/charm-refresh-build-version/,,https://github.com/canonical/charm-refresh-build-version,PyPI,y,y,Workaround for a charmcraft issue.
+charm-refresh,team,https://pypi.org/project/charm-refresh/,https://canonical-charm-refresh.readthedocs-hosted.com/latest/,https://github.com/canonical/charm-refresh,PyPI,y,y,In-place rolling refreshes of stateful charmed applications.
+charm-upgrade,legacy,https://pypi.org/project/charm-upgrade/,,https://github.com/canonical/charm-upgrade,PyPI,y,y,Alias for charm-refresh? (Repo URL redirects.)
```
tonyandrewmeyer, line 25: Pretty certain that this is indeed just an earlier name, and so shouldn't have a row.

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+charms.reconciler,experimental,https://pypi.org/project/charms.reconciler/,,https://github.com/charmed-kubernetes/charm-lib-reconciler,PyPI,y,y,Handle all Juju events in an ops charm with a single method.
+charms.templating.jinja2,legacy,https://pypi.org/project/charms.templating.jinja2/,https://pythonhosted.org/charms.templating.jinja2/,https://github.com/juju-solutions/charms.templating.jinja2,PyPI,,,Legacy library for Jinja templating in reactive charms.
+coordinated-workers,team,https://pypi.org/project/coordinated-workers/,,https://github.com/canonical/cos-coordinated-workers,PyPI,y,y,"Abstractions for charms following the coordinator-worker pattern, used by Observability Team."
+jujubigdata,legacy,https://pypi.org/project/jujubigdata/,http://pythonhosted.org/jujubigdata/,https://github.com/juju-solutions/jujubigdata,PyPI,,,Legacy library for developing Big Data charms.
+jujuresources,legacy,https://pypi.org/project/jujuresources/,http://pythonhosted.org/jujuresources/,https://github.com/juju-solutions/jujuresources,PyPI,y,y,Legacy library for loading binary resources.
```
tonyandrewmeyer, line 34: Maybe add the same sort of "use the feature built in to ops" text you had above?

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+ops_reactive_interface,,https://pypi.org/project/ops-reactive-interface/,https://github.com/canonical/ops-reactive-interface/tree/main/docs,https://github.com/juju-solutions/ops-reactive-interface,PyPI,y,y,"Helper for relation library developers, to allow an ops based relation library to interact with legacy, reactive charms."
```
tonyandrewmeyer, line 35: I get that this is for migrating from legacy charms, but I think it should get the legacy label too.

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+ops.manifest,,https://pypi.org/project/ops.manifest/,,https://github.com/canonical/ops-lib-manifest,PyPI,,y,Work with Kubernetes manifests.
+sborl,legacy,https://pypi.org/project/sborl/,,https://github.com/canonical/sborl,PyPI,y,y,Legacy library for implementing relation libraries.
```
tonyandrewmeyer, line 37: I totally missed this one when doing my research!

I think here we also want to say to use the functionality built in to ops.

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+juju_topology,legacy,https://charmhub.io/observability-libs/libraries/juju_topology,,https://github.com/canonical/observability-libs,Charmhub,y,y,Deprecated in favour of ``cosl.juju_topology.JujuTopology``.
+kubernetes_service_patch,legacy,https://charmhub.io/observability-libs/libraries/kubernetes_service_patch,,https://github.com/canonical/observability-libs,Charmhub,,y,Deprecated in favour of ops.Unit.set_ports.
+charm_tracing,legacy,https://charmhub.io/tempo-coordinator-k8s/libraries/charm_tracing,,https://github.com/canonical/tempo-coordinator-k8s-operator,Charmhub,y,y,Consider using ops[tracing] instead.
+data_models,legacy,https://charmhub.io/data-platform-libs/libraries/data_models,,https://github.com/canonical/data-platform-libs,Charmhub,y,y,``pydantic``-based typed relation data – unnecessary with ops 2.23+ due to the addition of typed relation data.
+upgrade,,https://charmhub.io/data-platform-libs/libraries/upgrade,,https://github.com/canonical/data-platform-libs,Charmhub,y,y,Manage in-place upgrades.
```
tonyandrewmeyer, line 42: This is legacy, it's the first version of the upgrade/refresh system, of which the latest (v3) is charm-refresh.

Probably should have a "use charm-refresh".

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+kubernetes_compute_resources_patch,,https://charmhub.io/observability-libs/libraries/kubernetes_compute_resources_patch,,https://github.com/canonical/observability-libs,Charmhub,,y,Patch Kubernetes compute resource limits.
+rollingops,,https://charmhub.io/rolling-ops/libraries/rollingops,,https://github.com/canonical/charm-rolling-ops,Charmhub,y,y,"Perform ""rolling"" operations across units, for example rolling restarts."
```
tonyandrewmeyer, line 44: I'm looking into this one in my secondment. It's flawed, and there isn't a replacement version. I'd leave it as-is for now, but likely we'll want to add a `legacy` label to it in the future.

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+client,,https://charmhub.io/zookeeper/libraries/client,,http://github.com/canonical/zookeeper-operator,Charmhub,y,y,Perform ``zookeeper`` operations.
```
tonyandrewmeyer, line 45: This is a pretty terrible name when it's not seen in the charm name context. Could it be `[Zookeeper] client` or something?

_docs/.custom_wordlist.txt
```diff
@@ -2,6 +2,7 @@
 autocomplete
 autocompletions
 args
+CNI
```
tonyandrewmeyer, line 5: I'm not convinced that CNI is a well-known enough term that we're ok using it as an abbreviation. If we spell it out, we also avoid having to have it here.

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+capture_events,legacy,https://charmhub.io/harness-extensions/libraries/capture_events,,https://github.com/PietroPasotti/harness-extensions,Charmhub,y,y,Helper for legacy harness tests. New charms should write scenario tests with ``ops[testing]`` instead.
+postgresql,team,https://charmhub.io/postgresql-k8s/libraries/postgresql,,https://github.com/canonical/postgresql-k8s-operator,Charmhub,y,y,Shared code between the ``posgresql`` and ``postgresql-k8s`` charms.
+multus,,https://charmhub.io/kubernetes-charm-libraries/libraries/multus,,https://github.com/canonical/kubernetes-charm-libraries,Charmhub,,y,Use the Multus Kubernetes CNI.
+backups,team,https://charmhub.io/mysql/libraries/backups,,https://github.com/canonical/mysql-operator,Charmhub,y,y,Shared code between the ``mysql`` and ``mysql-k8s`` charms.
+architecture,team,https://charmhub.io/mysql/libraries/architecture,,https://github.com/canonical/mysql-operator,Charmhub,y,y,Shared code between the ``mysql`` and ``mysql-k8s`` charms.
```
tonyandrewmeyer, line 50: These are also awfully generic names. Could we prefix them?

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+async_replication,team,https://charmhub.io/mysql/libraries/async_replication,,https://github.com/canonical/mysql-operator,Charmhub,y,y,Shared code between the ``mysql`` and ``mysql-k8s`` charms.
+s3_helpers,team,https://charmhub.io/mysql/libraries/s3_helpers,,https://github.com/canonical/mysql-operator,Charmhub,y,y,Shared code between the ``mysql`` and ``mysql-k8s`` charms.
+pgb,team,https://charmhub.io/pgbouncer-k8s/libraries/pgb,,https://github.com/canonical/pgbouncer-k8s-operator,Charmhub,y,y,Shared code between ``pgbouncer`` and ``pgbouncer-k8s`` charms.
+metrics_endpoint_discovery,,https://charmhub.io/observability-libs/libraries/metrics_endpoint_discovery,,https://github.com/canonical/observability-libs,Charmhub,,,Discover metric endpoints exposed by applications deployed to a K8s cluster.
+charm_tracing,legacy,https://charmhub.io/tempo-k8s/libraries/charm_tracing,,https://github.com/canonical/tempo-k8s-operator,Charmhub,y,y,Deprecated in favour of the ``tempo-coordinator-k8s`` charm's libs. New charms should use ops[tracing] for tracing charm code instead.
```
tonyandrewmeyer, line 56: Is this the same as line 40?

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+data_secrets,team,https://charmhub.io/data-platform-libs/libraries/data_secrets,,https://github.com/canonical/data-platform-libs,Charmhub,y,y,Secrets related helpers.
```
tonyandrewmeyer, line 57: This is deprecated in favour of the stuff inside of data_interfaces (the bit in there I'm working on in my secondment, but probably at least some will stay).

So this should be `legacy`.

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
+helper_cos,team,https://charmhub.io/opensearch/libraries/helper_cos,,https://github.com/canonical/opensearch-operator,Charmhub,,,Relies on data_platform_helpers.
+evt_sequences,legacy,https://charmhub.io/harness-extensions/libraries/evt_sequences,,https://github.com/PietroPasotti/harness-extensions,Charmhub,y,y,Helper for legacy harness tests. New charms should write scenario tests with ``ops[testing]`` instead.
+charm_logging,,https://charmhub.io/loki-k8s/libraries/charm_logging,,https://github.com/canonical/loki-k8s-operator,Charmhub,,,Add charm code logging to logs sent via ``loki_push_api``.
+cert_handler,,https://charmhub.io/observability-libs/libraries/cert_handler,,https://github.com/canonical/observability-libs,Charmhub,y,y,Wraps the requirer side of the ``tls-certificates-interface`` charm's ``tls_certificates`` lib.
+tls,team,https://charmhub.io/mysql/libraries/tls,,https://github.com/canonical/mysql-operator,Charmhub,y,y,Shared code between the ``mysql`` and ``mysql-k8s`` charms.
```
tonyandrewmeyer, line 62: This is also a very generic name for something quite specific.

dwilding (approved):
I added comments for naming/formatting stuff - nothing major.

James and I had some separate discussion earlier about the categories and icons. Mainly, I find 🪦 and 🚫 somewhat unintuitive:

I interpret 🚫 as "don't go near this lib". But as James pointed out, team internal libs might still have useful code to learn from. Whereas deprecated libs (🪦) really shouldn't be used.

So I'd at least want to swap the icons round - using 🚫 for deprecated libs. For team internal libs, I might use something like 💬 instead, meaning "see the description".

james-garner-canonical:
> One other thing I forgot to mention - could you remind me how the sorting is set up by default? Are the libs within each category supposed to be in alphabetical order? Some categories look almost alphabetical, while others seem arbitrarily ordered. I'm sure we previously discussed sorting, but I can't remember the details (sorry).
> 
> _Edit:_ Ah I see that the categories are also grouped by kind, and those groupings are mostly alphabetical. But even then there are some outliers, such as `juju_topology`, `dnf`, `kubernetes_service_patch` in the deprecated category.

@dwilding: I don't know if we have discussed sorting. There are two things to cover here: how each individual column sorts, and the default sorting of the table.

- The status column is sorted in a priority order (green tick first, etc). 
- The library name column is sorted alphabetically.
- The kind column sorts PyPI -> git -> Charmhub
- The description column is special, because it includes the substrate tags. It is first sorted by substrate, (both -> machine only -> K8s only -> unknown). It is then sorted by status, then by name -- I thought this might be more useful than sorting by the description text.

The default sorting for the table is by status (green tick first, etc), then by kind (PyPI first, etc), then by description (aka by substrate, then by name).

This is just where I landed while incrementally working on the table design. I'm open to changing it. For example, maybe it's more intuitive for the default sort to be status, then name? (And presumably then kind and then substrate, but I think the names are unique enough that that won't come up).

The description column sorting could also be tweaked. I think the substrate sorting is useful, but I'm very open to changing the rest, though I think that sorting by the description itself is not that useful (with empty descriptions naturally sorting first) -- but perhaps substrate, then status, then name (if we do that for the default sort) would be good.

dwilding:
Ah, I got you, thanks for explaining how it works.

I think it's nice to be able to manually sort the description column and see the different substrate groupings. (When doing that, I don't think sorting the actual text of the description would be useful)

For the default sort order, I don't think I'd include substrate as the third level. The first two levels (status and kind) cause enough variation that the description column doesn't really _look_ obviously grouped.

Since it's possible to manually sort by substrate, I'd probably go with (status, kind, name) for the default order. Or at least, I'm curious to see how that looks!

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
 hpc-libs,team,https://github.com/charmed-hpc/hpc-libs,,,git,y,y,Used internally by HPC charms.
```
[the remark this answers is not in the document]
james-garner-canonical: Good call, I'll do this as the final step before merging to avoid messing up the review

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
 charm-helpers,legacy,https://pypi.org/project/charmhelpers/,https://charm-helpers.readthedocs.io/,https://github.com/juju/charm-helpers,PyPI,y,y,Pre-``ops`` library used by reactive charms.
 cos-lib,dep,https://pypi.org/project/cosl/,,https://github.com/canonical/cos-lib,PyPI,y,y,A dependency (via PYDEPS) of popular charm libs such as ``loki_push_api``.
+serialized-data-interface,legacy,https://pypi.org/project/serialized-data-interface/,,https://github.com/canonical/serialized-data-interface,PyPI,y,y,Relation data validation – use the features available in ops instead.
+oci-image,legacy,https://pypi.org/project/oci-image/,,https://github.com/juju-solutions/resource-oci-image,PyPI,,y,Work with OCI image resources – repo archived in March 2024.
```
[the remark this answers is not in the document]
james-garner-canonical:
```suggestion
oci-image,legacy,https://pypi.org/project/oci-image/,,https://github.com/juju-solutions/resource-oci-image,PyPI,,y,Work with OCI images in podspec charms. New Kubernetes charms should use the sidecar pattern. Repo archived in March 2024.
```

_docs/reference/libs-non-rel-raw.csv
james-garner-canonical: Sounds reasonable
```suggestion
ops_reactive_interface,legacy,https://pypi.org/project/ops-reactive-interface/,https://github.com/canonical/ops-reactive-interface/tree/main/docs,https://github.com/juju-solutions/ops-reactive-interface,PyPI,y,y,"Helper for relation library developers, to allow an ops based relation library to interact with legacy, reactive charms."
```

_docs/reference/libs-non-rel-raw.csv
tonyandrewmeyer: It's just `ops.model.resources` from what I could tell from a quick look at the lib.

_docs/reference/libs-non-rel-raw.csv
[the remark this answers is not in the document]
james-garner-canonical: Thanks, merged with other change on this line.

james-garner-canonical:
```suggestion
sborl,legacy,https://pypi.org/project/sborl/,,https://github.com/canonical/sborl,PyPI,y,y,Legacy library for implementing interface libraries. New interface libraries should use the typed relation data feature available in Ops 2.23+.
```

That raises an interesting question. Almost all interface libraries are Charmhub-hosted libs, so they don't have native support for dependency versioning. Aside from migrating to a Python package when we have a nice story for that, what's the best practice for using features available only in certain Ops versions in Charmhub-hosted interface libs?

_docs/reference/libs-non-rel-raw.csv
james-garner-canonical: No, this lib was deprecated in favour of the line 40 lib. Should be clearer when I add the charm namespace to the lib names.

_docs/reference/libs-non-rel-raw.csv
james-garner-canonical:
```suggestion
upgrade,legacy,https://charmhub.io/data-platform-libs/libraries/upgrade,,https://github.com/canonical/data-platform-libs,Charmhub,y,y,Manage in-place upgrades. Deprecated in favour of ``charm-refresh``.
```

_docs/reference/libs-non-rel-raw.csv
james-garner-canonical: I'll resolve these comments as I go through, and fix them in bulk by namespacing all the Charmhub-hosted charms.

_docs/reference/libs-non-rel-raw.csv
james-garner-canonical: Is this accurate?

```suggestion
data_secrets,legacy,https://charmhub.io/data-platform-libs/libraries/data_secrets,,https://github.com/canonical/data-platform-libs,Charmhub,y,y,Secrets-related helpers for interfaces. Deprecated in favour of the ``data-platform-libs.data_inferfaces`` interface lib.
```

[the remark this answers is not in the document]
james-garner-canonical: Merged into other change on this line.

_docs/reference/libs-non-rel-raw.csv
```diff
@@ -12,7 +12,51 @@ charmed-kubeflow-chisme,team,https://pypi.org/project/charmed-kubeflow-chisme/,,
...
 data-platform-helpers,team,https://pypi.org/project/data-platform-helpers/,,https://github.com/canonical/data-platform-helpers,PyPI,y,y,Used internally by the Data Charm Engineering team.
 mongo-charms-single-kernel,team,https://pypi.org/project/mongo-charms-single-kernel/,,https://github.com/canonical/mongo-single-kernel-library,PyPI,y,y,Used internally by the Data Charm Engineering team.
 data-platform-libs,team,https://charmhub.io/data-platform-libs/libraries,,https://github.com/canonical/data-platform-libs,Charmhub,y,y,Used internally by the Data Charm Engineering team.
```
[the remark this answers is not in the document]
james-garner-canonical: There are seven libs under this charm, three are in this table already, and the other four are interface libs and will be in the other table, so this line can go altogether.
```suggestion
```

_docs/reference/libs-non-rel-raw.csv
tonyandrewmeyer: Tricky one. "team" doesn't seem right because it's already used by other teams. "experimental" feels like it's in the other direction, going from that to supported.

I would leave it blank for now, and hopefully in a week or so I should have a better answer that can be a follow-up PR.

james-garner-canonical:
I've updated the emojis as per discussion. I've also updated the status key table generation code to only include statuses we're using on that page.

Charmhub-hosted libs now have their name prefixed with `<charm_name>.`. This makes some names very long, so I've added word break markers (`<wbr>`) to the text of the html links generated, so they can break on `.`, `-`, or `_`. This allows the column to stay in the confines allocated, and keeps the table as a whole in width we have to work with.

I've also added `style="white-space:nowrap;"` spans around the `(docs, src)` snippets so that they'll be dropped onto a newline together if needed, rather than being broken in the middle.

Please take a look and let me know what you think. Status suggestions for the remaining libraries are welcome, but not required before merging.

PR #80, opened 2025-06-30
docs: add remaining interface libraries to listing

james-garner-canonical:
This PR adds the other known interface libraries to the library listing docs. The libraries were found using the methodology described in PR #73 (where only a small number of the known libraries were added initially to simplify review). With the future organisation and review of interface libraries unclear, the status column has only been used to mark deprecated libraries (no recommended libraries yet).

The built page can be seen [here](https://canonical-charmlibs--80.com.readthedocs.build/reference/interface-libs/).

The `generate` extension is updated to wrap interface names in a `no-spellcheck` span even when no url is known for the interface (if a url is known, the link generation code already adds the `no-spellcheck` class). The extension is also updated to filter out unlisted libs for interface libraries (it already does so for non-interface libraries), and handle the 'popular' status.


_docs/.custom_wordlist.txt
```diff
@@ -5,6 +5,7 @@ args
 codebase
 databag
 databags
+datasource
```
tonyandrewmeyer, line 8: I feel like this should be "data source" in the description and so removed here.

_docs/.custom_wordlist.txt
```diff
@@ -28,8 +29,10 @@ symlinks
 templating
 tooltip
 tooltips
+TLS
 UTF
 vendored
+webui
```
tonyandrewmeyer, line 35: Similarly, I'd suggest "web UI" in the description, removing it from here.

_docs/.custom_wordlist.txt
```diff
@@ -42,8 +45,10 @@ pathlib
 pathops
 ContainerPath
 
+GRPC
```
tonyandrewmeyer, line 48:
```suggestion
gRPC
```

_docs/extensions/generate.py
```diff
@@ -205,6 +228,10 @@ def _get_status_key_table_dropdown(entries: Iterable[_CSVRow]) -> str:
     return _KEY_DROPDOWN_HEADER + _indent_lines(table, level=3)
 
 
+def _not_unlisted(row: _CSVRow) -> bool:
```
tonyandrewmeyer, line 231: I don't love a double-negative. `not_unlisted` is more technically correct, I guess, but what do you think about calling this something like `_is_listed` or `_in_table`? It does also leave it more open to other categories being excluded, although I don't know if that's likely.

_docs/extensions/generate.py
```diff
@@ -175,22 +198,22 @@ def key(row: tuple[str, ...]) -> tuple[str, ...]:
         status, _name, _kind, desc = row
         return status, desc
 
-    rows = [(_status(e), _name(e), _kind(e), _rel_description(e)) for e in entries]
-    return _LIBS_TABLE_HEADER + _rst_rows(sorted(rows, key=key))
+    rows = [
+        (_status(entry), _name(entry), _kind(entry), _rel_description(entry))
+        for entry in filter(_not_unlisted, entries)
```
tonyandrewmeyer, line 203: It seems a bit odd to use a `filter()` inside of a list comprehension. Any reason to do this over something like this?

```suggestion
        for entry in (row for row in entries if _not_unlisted(row))
```

_docs/extensions/generate.py
```diff
@@ -66,31 +67,53 @@ def _generate(app: sphinx.application.Sphinx):
...
+   :class: sphinx-datatable
+   :widths: 1, 40, 1, 60
+   :header-rows: 1
+
+   * -
+     - Library
+     - Kind
+     - Interface
+"""
+_GEN_LIBS_TABLE_HEADER = """.. list-table::
```
tonyandrewmeyer, line 101: Is the "_GEN" in this name "general"? If so, wouldn't that more be for the other table?

tonyandrewmeyer (approved):
I'm trusting that the content of the CSV is correct (I did skim over the generated page, particularly looking at the ones people shouldn't use).

For popular: I don't think this is a useful metric for choosing a charm library. It is useful for choosing what you might want to integrate your charm with ("oh, people usually provide tracing and ingress, I should do that too"), but it doesn't seem like a list of interface libs is where you would look for that. Popularity would be a signal if interfaces had multiple libraries (I might be more inclined to use the more popular one), but generally that is not the case right now. So I would suggest dropping it.

I've said this elsewhere multiple times, sorry, but I do feel that the critical feature of this list is a "Charm Tech recommends" type stamp, and that should be based on a clearly defined set of review criteria *and* that it's currently passing a set of automated tests (ideally, that would be both unit tests and results of integration tests from SolQA). It's certainly fine to not have that in this PR, but I don't think we should put in inferior substitutes for that in the meantime.

_docs/extensions/generate.py
james-garner-canonical: Good call, I've gone with `_is_listed`.

_docs/extensions/generate.py
james-garner-canonical: Good point, instead of a nested list comprehension I've added a conditional to the outer list comprehension, thanks.

_docs/extensions/generate.py
james-garner-canonical: It is 'general'. I think the table names are the right way around -- in the code snippet shown above you can see that `Interface` is a column of the other table header variable.

That said, I've taken this feedback as a good opportunity to univerally switch to using `general` and `interface` (instead of a mix of `rel`, `non-rel` and `gen`). I've also swapped out the variables `ref_dir` and `gen_dir` for `reference_dir` and `generated_dir` to remove another source of ambiguity.

james-garner-canonical:
I've updated the interface libraries table preamble with a sentence about the future plans, as an indication of why the information in the table is a bit bare bones currently.

dwilding (approved):
A couple of very minor suggestions, but all good from me.

Re Tony's comment about popular libs/interfaces, it would be nice to have that in form of that somewhere, in the future. E.g., in some kind of "what it typically takes to make a production charm" doc.

PR #84, opened 2025-08-06
ci: rename static recipe to lint and run both lint and static there

james-garner-canonical:
This PR updates the repo's CI to follow the latest charming recommendations for CI environment/recipe names.

Currently we don't actually follow the recommendations for being able to run the checks from the project subdirectories with a bare `just lint` etc. but since it's not a charm repo I think that's ok for now.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,8 +11,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  check:
-    uses: ./.github/workflows/check.yaml
+  fast-lint:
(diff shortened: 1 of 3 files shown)


tonyandrewmeyer (approved):
I don't really know Just but everything else makes sense to me.

It seems like it's "check" in most places to mean what would be "lint" in a charmcraft profile, but it's "lint" in one workflow (this seems odd), and then the justfile has "check" (which I think is "lint without type checking") and "lint". It does seem like it would be less confusing to have the same names & meanings everywhere, but I assume a bunch of this is not wanting to run type checking as often.

PR #85, opened 2025-08-06
docs: update links to Ops and Charmcraft docs

dwilding:
Now that Ops and Charmcraft docs have moved to documentation.ubuntu.com, we should update links (not time-critical, as there are redirects).


james-garner-canonical (approved):
Thanks for this!

PR #86, opened 2025-08-06
ci: unify publication and test publication logic

james-garner-canonical:
Initially, this repo just had the publication logic duplicated (4x) across the two packages, and across the workflows to publish to real `pypi.org` and to `test.pypi.org`. This PR moves all the logic to a dedicated `publish.yaml` job. Each package then has a single `publish-<package name>.yaml` workflow that run on pushing the appropriate tag and calls `publish.yaml` with the appropriate arguments.

Since we only publish to `test.pypi.org` manually, this PR avoids the need for each package to also add a `test-publish-<package name>.yaml` by adding a single `test-publish.yaml` which runs on manual `workflow_dispatch` and takes the package name as an argument.

Touches the same files as #84, whichever is updated second will need to have the changes merged with some manual conflict resolution.


.github/workflows/test-publish.yaml
```diff
@@ -0,0 +1,14 @@
+name: Publish a package to test.pypi.org
+on:
+  workflow_dispatch:
+    inputs:
+      package:
+        required: true
+        type: string
```
james-garner-canonical, line 7: It would be cool if this could be a dropdown, but it would have to be manually populated here. Then we'd probably want CI to make sure that the list here is always up to date, which seems like a pain.

PR #87, opened 2025-08-07
ci: only run tests for a package's supported Python versions

james-garner-canonical:
It seems essential that we allow individual charmlibs to set their own `requires-python`. Currently, CI hardcodes the Python versions that unit tests etc are run against. This PR adds a Python script that contains this hardcoded set of Python versions and returns only the versions that match the individual package's `requires-python` constraints.

The Python script is quite verbose, but hopefully straightforward and maintainable -- but open to alternatives that cut this down if possible.


benhoyt (approved):
Looks reasonable to me, thanks.

PR #118, opened 2025-08-19
ci: release on version change

james-garner-canonical:
This PR changes our publication workflow from running on the push event of specific tags (one per library), to running on every push to main. The `unify-publish-inputs.py` script has been modified to:
1. Get a list of the packages that changed in this event
2. Narrow that list to just those that have a different version now (new packages are included as having a different version -- the first merge to main will be a release, so they should start with a `0.X` + `.devY` / `.aY` version unless they really are ready for a true release)

The publish job is then run with just those packages (using the matrix strategy).

Narrowing the publish job to only run on packages with a changed version is mostly an efficiency and UX thing -- running the publish job with a package whose version is identical to a released version will fail for that package, but there shouldn't be any other bad effects.

This release strategy has the following benefits:
- We don't need to worry about permissions to use github releases or push tags to main
- Adding a new library won't require editing the publish job to add its tag
- `CODEOWNERS` for a library control its releases via their approval and merge
- We don't need to worry about limitations of the push tag event, like how it won't trigger at all if more than 3 tags are pushed at once

I tested this change by:
1. merging the new CI into my fork's [main](https://github.com/james-garner-canonical/charmtech-charmlibs/commit/571d344259dc056d5fe8565af39630ffac1c8f5c)
2. triggering a release by bumping two package versions in the same push
3. confirming that the publish job ran as expected for those two packages only ([failing at the trusted publisher step](https://github.com/james-garner-canonical/charmtech-charmlibs/actions/runs/17058578043/job/48361424460) as the packages on PyPI only trust `canonical/charmlibs`)


.github/unify-publish-inputs.py
```diff
@@ -43,24 +51,64 @@ def _main() -> None:
...
+    return {*(p[0] for p in diff), *('/'.join(p[:2]) for p in diff)}
+
+
+def _get_all_packages(root: pathlib.Path | str, exclude: str | None = None) -> list[str]:
+    root = pathlib.Path(root)
+    paths = [root / '.package', *root.glob(r'[a-z]*')]
+    return sorted(str(path) for path in paths if path.is_dir() and path.name != exclude)
+
+
+def _git_archive(ref: str, directory: pathlib.Path) -> None:
```
tonyandrewmeyer, line 91: A nit, but I don't love this as a name, particularly where it's used. Maybe something more like extract_git_archive?

.github/unify-publish-inputs.py
```diff
@@ -43,24 +51,64 @@ def _main() -> None:
...
+    for p in changed:
+        logger.info('%s: %s -> %s', p, old_versions[p], new_versions[p])
+    return changed
+
+
+def _get_changes(old_ref: str, new_ref: str) -> set[str]:
+    cmd = ['git', 'diff', '--name-only', old_ref, new_ref]
+    output = subprocess.check_output(cmd, text=True)
+    diff = [p.split('/') for p in output.split('\n')]
+    return {*(p[0] for p in diff), *('/'.join(p[:2]) for p in diff)}
```
tonyandrewmeyer, line 82: No need to change it, but this feels like it's tipping towards overly clever and away from readable and understandable by a casual reader. Maybe add a comment so people don't need to work their way through it? Or a docstring for the method even?

.github/unify-publish-inputs.py
```diff
@@ -43,24 +51,64 @@ def _main() -> None:
...
+            )
+        ]
+        changed_packages = sorted(changes.intersection(all_packages))
+        new_versions = {p: _get_version(new_root, p) for p in changed_packages}
+    with tempfile.TemporaryDirectory() as td2:
+        old_root = pathlib.Path(td2)
+        _git_archive(ref=old_ref, directory=old_root)
+        old_versions = {p: _get_version(old_root, p) for p in changed_packages}
+    changed = [p for p in changed_packages if old_versions[p] != new_versions[p]]
+    for p in changed:
```
tonyandrewmeyer, line 73: Nit: I feel like you probably lose the benefits of a list comp by immediately iterating over it afterwards.

tonyandrewmeyer (approved):
I like the overall approach. Like some of the other decisions, I think the trade-off of making this a bit more complex in exchange for the experience of the library maintainer being simpler is generally good.

A few small notes, but happy to approve either way.

I didn't try this out - happy to assume your test was sufficient for something that's infra.

I'll defer to Dima's comments in terms of the workflow piece, and leave it to the two of you to conclude on those.

.github/workflows/publish.yaml
```diff
@@ -24,26 +19,35 @@ jobs:
...
+      package: ${{ matrix.package }}
       skip-juju: ${{ fromJSON(needs.unify-inputs.outputs.skip-juju) }}
 
   build-n-publish:
     needs: [unify-inputs,tests]
-    name: Build and publish ${{ needs.unify-inputs.outputs.package }}
+    strategy:
+      matrix:
+        package: ${{ fromJSON(needs.unify-inputs.outputs.packages) }}
+    name: Build and publish ${{ matrix.package }}
```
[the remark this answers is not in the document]
james-garner-canonical: I think you're right that in their current form, both of these should be `fail-fast: false`.

I figured that a single failing test would block the whole publish job, so we might as well bail out of the workflow early so we get an email ASAP without waiting for any other libs to run all their slow tests. But you make a good point that it makes the results less useful, so I'm happy to change that. And I think `fail-fast: false` is just completely correct for the `build-n-publish` job, good call there.

The 'suggestion' part of this comment is very difficult to follow T_T looks like the `uses` bit for `tests` has been accidentally dedented too. Am I correct in thinking that the suggested changes are solely setting `fail-fast: false` for both `tests` and `build-n-publish`?

---

Ideally I'd want to use a reusable `test-and-publish` workflow here, which runs for a single package, and doesn't `build-n-publish` if its `tests` fail. Unfortunately, trusted publishing doesn't (yet) work with reusable workflows, so the actual publishing has to take place in the top-level triggered workflow. Here are some ideas for working around that limitation, but they add some implementation complexity so I guess not worth it right now -- I figure that most merges to main with a version bump will just be a single package anyway.

Idea 1: We indirectly trigger the publishing workflow instead of using it directly. For example, this workflow runs a matrix job that just pushes a tag. Then the actual test and publish job runs on tag push and parses the tag to figure out which single package to publish. Downside: will be annoying to debug the actual publish job not triggering when expected to, since the coupling is indirect.

Idea 2: Another option I'm thinking about is to use `continue-on-error` for `tests`, and then have `build-n-publish` check the actual result for the specific package as its first step and bail out there.

Idea 3: Unfortunately, I don't think I can use the reusable `tests` workflow as a step in a job, otherwise I'd just have a `test-n-build-n-publish` job here that runs the tests as the first step. Maybe the `tests` logic could be moved to an action instead to make this possible. I'm not totally familiar with the implications of that, but this seems like it might be the cleanest and simplest solution?

.github/workflows/publish.yaml
```diff
@@ -55,14 +59,14 @@ jobs:
...
         with:
-          subject-path: '${{ needs.unify-inputs.outputs.package }}/dist/*'
+          subject-path: '${{ matrix.package }}/dist/*'
       - name: Publish
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
-          packages-dir: ./${{ needs.unify-inputs.outputs.package }}/dist/
+          packages-dir: ./${{ matrix.package }}/dist/
           repository-url: ${{ needs.unify-inputs.outputs.repository-url }}
           verbose: true
```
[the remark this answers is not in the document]
james-garner-canonical: I don't think we can directly add something to the same package version since this will only run when the version changes. I guess users could do two PRs, one to change the version to something else, and then one to go back to the original version and publish again ... but I don't think we really want to support this use case. The docs [recommend against using skip-existing where possible](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#tolerating-release-package-file-duplicates), so I don't think it's worth using it here. What do you think?

`verbose` sounds helpful to have on by default though, thanks.

.github/unify-publish-inputs.py
james-garner-canonical: Hmm, it does run `git archive` and *then* extracts it, which I don't think `extract_git_archive` communicates. I guess the actual intention might be better, like `_snapshot` or `_snapshot_repo`?

.github/unify-publish-inputs.py
james-garner-canonical: You're right, this could be a lot more readable. How's this?
```py
def _get_changes(old_ref: str, new_ref: str) -> set[str]:
    """Return the first and first two parts of the changed paths as a set.

    e.g. If the file 'foo/bar/baz' has changed, return {'foo', 'foo/bar'}.
    """
    cmd = ['git', 'diff', '--name-only', old_ref, new_ref]
    output = subprocess.check_output(cmd, text=True)
    changes = set()
    for line in output.split('\n'):
        parts = pathlib.Path(line).parts
        changes.add(parts[0])
        changes.add(str(pathlib.Path(*parts[:2])))
    return changes
```

.github/unify-publish-inputs.py
james-garner-canonical: Ah, I understand what you mean now.

I tend to assume that iterating itself is pretty low cost compared to whatever you're doing inside the loop, such that it's not worth optimising for the number of times you iterate (e.g. 2 loops of N items with 1 operation per iteration will be similar to 1 loop of N items with 2 operations per iteration) -- with  you're repeating some time/memory expensive operation to iterate (like reading from disc).

I'll stick with the existing version, cheers.

.github/workflows/publish.yaml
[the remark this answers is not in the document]
james-garner-canonical: I'm going to merge with `fail-fast: false` for both, but get back to me if you think it's worth trying anything else in future (e.g. indirectly triggering publication by tagging in this workflow, using `continue-on-error` for tests and checking manually in `build-n-publish`).

PR #125, opened 2025-08-24
chore: add tls certificates lib with history

james-garner-canonical:
This PR is a demo of what hosting an interface library and its interface specification in the `charmlibs` monorepo might look like.

To demonstrate a simple way that a Python package might support multiple interface versions, the latest libraries supporting the `v0` and `v1` versions of the interface have both been ported, exposed as modules of the `charmlibs.interfaces.tls_certificates` package -- whether or not this is actually suitable or required for the `tls_certificates` interface library is a separate question, this is just a demo of how it could be done. This method of supporting multiple interface versions places all the responsibility for figuring out which version to use on the charm -- hopefully in reality more of this will be taken care of by the libraries, as interface versioning is figured out.

The code and tests have not been changed, aside from imports, formatting, and docstring updates. The tests pass.

The generated reference docs can be viewed [here](https://canonical-ubuntu-documentation-library--125.com.readthedocs.build/charmlibs/reference/charmlibs/interfaces/tls_certificates/).

A diff of the changes compared to `canonical/tls-certificates-interface` can be seen [here](https://github.com/james-garner-canonical/tls-certificates-interface/pull/1/files#diff-df3f8a48aced769bf005dfa87b52153adb86db1902f69a5b6d33c54fe7194948).

---

This PR additionally demonstrates a method of keeping the development history of the library and specifications in this repo. **We probably won't do this, and will just squash merge everything into main, with a link to the originating repo(s) in the PR and commit message.**

This branch contains the merged history of  history of [canonical/tls-certificates-interface](https://github.com/canonical/tls-certificates-interface) and [canonical/charm-relation-interfaces](https://github.com/canonical/charm-relation-interfaces). This PR would be merged into the `history/interfaces/tls-certificates` branch with the 'Merge' method (not the usual 'Squash'). `history/interfaces/tls-certificates` would then be 'Squash' merged into `main`. This is probably a bit overkill, and we could just include a pointer to the original repositories when porting a library.

---

The recipe for creating the merged history was:
```bash
# in my local copy of my charmlibs fork
git switch --orphan $branchname  # make a new empty branch

git remote add charm-relation-interfaces https://github.com/canonical/charm-relation-interfaces
git fetch charm-relation-interfaces
git merge charm-relation-interfaces/main --allow-unrelated-histories
# git rm everything except interfaces/tls_certificates
# git mv that to the schema subdirectory

git remote add tls-certificates-interface https://github.com/canonical/tls-certificates-interface
git fetch tls-certificates-interface
git merge tls-certificates-interface/main --allow-unrelated-histories
# git rm everything except library code and tests
# git mv it all to the interfaces subdirectory

# upstream is canonical/charmlibs
git fetch upstream
git merge upstream/main --allow-unrelated-histories  
```


.docs/.sphinx/requirements.txt
```diff
@@ -1,7 +1,7 @@
 canonical-sphinx[full]
+pydantic
```
dwilding, line 2: Are you planning to maintain this list (and autodoc_mock_imports) manually as libs are added, rather than automating somehow? Manually seems better for now, as it's clearer and it's presumably not a big job to inspect and merge the deps of each lib.

dwilding:
Minor thing: I noticed that the sorting of the reference docs is:

```
apt
tls_certificates
pathops
```

Which isn't alphabetical. Presumably because glob is seeing:

```
charmlibs/apt
charmlibs/interfaces.tls-certificates
charmblibs/pathops
```

This is almost surely a terrible idea, but how about generating `tls-certificates.interfaces` instead of `interfaces.tls-certificates`?

.docs/.sphinx/requirements.txt
james-garner-canonical: For whatever reason, `pydantic` doesn't work in `autodoc_mock_imports`. I figure it's a common dependency, so it can go in the core requirements. But ideally I'd like to just automatically use each library's real dependencies -- the issue is that they're not guaranteed to all be compatible, so some shenanigans would need to be involved to build each reference docs page one by one with the library's deps, and then combine them into the final docs build.

james-garner-canonical:
Re: sorting the table of contents, you're right that this isn't desirable. Maybe we can keep a flat TOC with good sorting by prepending `interfaces.` to the interface ones. I figure we'll ultimately have them separately under `charmlibs` and `charmlibs.interfaces` dropdowns, so the problem will go away then.

james-garner-canonical:
Closed in favour of:
- #168

PR #128, opened 2025-08-25
[WIP] feat: nginx charmlib

PietroPasotti:
This PR adds an 'nginx' charmlib that contains a few abstractions to manage nginx and nginx-prometheus-exporter sidecar containers for kubernetes charms.

TODO: 
more/better tests


nginx/pyproject.toml
```diff
@@ -0,0 +1,58 @@
...
+version = {file = ["src/charmlibs/nginx/_version.txt"]}
+
+[tool.ruff]
+extend = "../pyproject.toml"
+src = ["src", "tests/unit", "tests/functional", "tests/integration"]  # correctly sort local imports in tests
+
+[tool.pyright]
+extend = "../pyproject.toml"
+include = ["src", "tests"]
+pythonVersion = "3.8"  # check no python > 3.8 features are used
```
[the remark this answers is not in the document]
james-garner-canonical: Yeah the plan is to support Python 3.10+ from `main`, possibly with maintenance branches for older Python versions if there's a good case for it.

nginx/tests/integration/test_meta.py
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Ensure that the test charms aren't out of sync."""
+
+from __future__ import annotations
+
+import pathlib
+
+BASES = ['20.04', '24.04']
```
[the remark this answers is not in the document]
james-garner-canonical: Currently it just so happens that these are the two bases I wanted to test `pathops` with. I'll update CI to make this more general shortly. Note that this particular file is just tests to make sure the `pathops` test charms are kept in sync, that's all the `BASES` here are for -- making the integration tests more configurable is WIP.

nginx/src/charmlibs/nginx/config.py
```diff
@@ -0,0 +1,622 @@
...
+
+@dataclass
+class NginxLocationConfig:
+    """Represents a `location` block in an Nginx configuration.
+
+    For example,
+    ``NginxLocationConfig('/', 'foo', backend_url="/api/v1" headers={'a': 'b'},
+      modifier=EXACT, is_grpc=True, use_tls=True)``
+
+    would result in the nginx config:
```
james-garner-canonical, line 126: This should fix the docs build. The `::` and following empty lines before indented code sections are the important bit.

```suggestion
    For example::

        NginxLocationConfig(
            '/',
            'foo',
            backend_url="/api/v1"
            headers={'a': 'b'},
            modifier=EXACT,
            is_grpc=True,
            use_tls=True,
        )

    would result in the nginx config::

```

nginx/pyproject.toml
```diff
@@ -0,0 +1,59 @@
+[project]
+name = "charmlibs-nginx"
+description = "Nginx sidecar abstractions."
+readme = "README.md"
+requires-python = ">=3.8"
```
james-garner-canonical, line 5: We plan to not explicitly support Python 3.8 on `main` (e.g. using higher `pyright` version soon that drops 3.8 type stubs). Going to 3.10 should fix the `lint` job's type checking errors (by no longer running it with Python 3.8). If you do have the need to support 3.8 with this lib, we can talk about it though.

nginx/pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+version = {file = ["src/charmlibs/nginx/_version.txt"]}
+
+[tool.ruff]
+extend = "../pyproject.toml"
+src = ["src", "tests/unit", "tests/functional", "tests/integration"]  # correctly sort local imports in tests
+
+[tool.pyright]
+extend = "../pyproject.toml"
+include = ["src", "tests"]
+pythonVersion = "3.8"  # check no python > 3.8 features are used
```
james-garner-canonical, line 55: Change version here too if you go that route.

nginx/tests/integration/test_nginx.py
```diff
@@ -0,0 +1,32 @@
...
+
+"""Integration tests using a real Juju and charm to test ContainerPath."""
+
+from __future__ import annotations
+
+import jubilant
+import pytest
+
+from conftest import deploy
+
```
james-garner-canonical, line 12: Since it doesn't look like you're packing a machine charm, I assume that this library (or at least the integration tests) are K8s charm specific. You can skip the machine integration tests being run at all (including packing) by marking all your tests as `k8s_only`.
```suggestion

pytestmark = pytest.mark.k8s_only
```
Or if you want to future proof against forgetting to add `pytestmark` in future integration test files, consider using [pytest_collection_modifyitems](https://docs.pytest.org/en/latest/reference/reference.html#pytest.hookspec.pytest_collection_modifyitems) in your `conftest.py`.

nginx/pyproject.toml
PietroPasotti: no I don't but the requires-python in this lib needs to be compatible with the requires-python in the root package, otherwise uv gets confused and all `just` commands break.

nginx/tests/integration/test_nginx.py
PietroPasotti: yea I don't think this will ever be relevant for machine charms

james-garner-canonical:
The docs build failure is on me too, will fix.

nginx/tests/integration/charms/k8s/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[project]
+name = "test"
+version = "0.1.0"
+description = "Test charm for integration tests"
+requires-python = ">=3.8"
```
james-garner-canonical, line 5: `uv lock` of the assembled charm for integration tests will fail because of this if we drop Python 3.8.
```suggestion
requires-python = ">=3.10"
```
Other than that, dropping Python 3.8 seems to work fine in this dummy PR https://github.com/canonical/charmlibs/pull/150

nginx/tests/integration/charms/k8s/pyproject.toml
```diff
@@ -0,0 +1,12 @@
...
+dependencies = [
+    "ops==2.*",
```
james-garner-canonical, line 7:
```suggestion
    "ops==3.*",
```
Need to change this too when dropping 3.8 -- plus of course `nginx/pyproject.toml`.

nginx/tests/typecheck.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2025 Canonical Ltd.
```
james-garner-canonical, line 1: Just a note that the `typecheck.py` module is totally optional, just a handy place to put type checking statements (especially useful for `pathops` since we had a `typing.Protocol`). Feel free to retain it if it's useful for this library of course.

nginx/tests/unit/conftest.py
```diff
@@ -0,0 +1,46 @@
...
+        }
+    )
+
+
+@pytest.fixture
+def ctx() -> ops.testing.Context:
+    return ops.testing.Context(
+        ops.CharmBase,
+        meta={
+            'name': 'tony',
```
james-garner-canonical, line 40: :smile:

nginx/tests/unit/test_container_managers.py
```diff
@@ -0,0 +1,96 @@
...
+    return request.param
+
+
+@pytest.fixture(params=[3030, 5050])
+def nginx_pexp_port(request):
+    return request.param
+
+
+@pytest.fixture
+def ctx(nginx_port, nginx_insecure, nginx_pexp_port, update_cacerts):
```
james-garner-canonical, line 31:
```suggestion
def ctx(nginx_port: int, nginx_insecure: bool, nginx_pexp_port: int, update_cacerts: bool):
```

nginx/tests/unit/test_container_managers.py
```diff
@@ -0,0 +1,96 @@
...
+            )
+
+            self.nginx.reconcile(upstreams_to_addresses={})
+            self.nginx_pexp.reconcile()
+
+    return scenario.Context(MyCharm, meta=MyCharm.META)
+
+
+@pytest.fixture
+def base_state(update_cacerts):
```
james-garner-canonical, line 58:
```suggestion
def base_state(update_cacerts: bool):
```

nginx/tests/unit/test_container_managers.py
```diff
@@ -0,0 +1,96 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+import ops
+import pytest
+import scenario
```
james-garner-canonical, line 5: Let's use `ops.testing` instead if possible.

nginx/tests/unit/test_container_managers.py
```diff
@@ -0,0 +1,96 @@
...
+    return scenario.State(
+        leader=True,
+        containers={
+            scenario.Container('nginx', can_connect=True, execs=execs),
+            scenario.Container('nginx-pexp', can_connect=True, execs=execs),
+        },
+    )
+
+
+def test_nginx_container_service(ctx, base_state):
```
james-garner-canonical, line 71:
```suggestion
def test_nginx_container_service(ctx: ops.testing.Context, base_state: ops.testing.State):
```

nginx/tests/unit/test_container_managers.py
```diff
@@ -0,0 +1,96 @@
...
+    # given any event
+    state_out = ctx.run(ctx.on.update_status(), state=base_state)
+    # the services are running
+    assert state_out.get_container('nginx').services['nginx'].is_running()
+    assert state_out.get_container('nginx-pexp').services['nginx-prometheus-exporter'].is_running()
+
+
+def test_layer_commands(ctx, base_state, nginx_pexp_port, nginx_insecure, nginx_port):
```
james-garner-canonical, line 79:
```suggestion
def test_layer_commands(
    ctx: ops.testing.Context,
    base_state: ops.testing.State,
    nginx_pexp_port: int,
    nginx_insecure: bool,
    nginx_port: int,
):
```

nginx/tests/unit/test_nginx_config_write.py
```diff
@@ -0,0 +1,25 @@
+# Copyright 2025 Canonical
+# See LICENSE file for licensing details.
+import typing
+
+from scenario import CharmEvents
```
james-garner-canonical, line 5: Ditto here re: `ops.testing`.

nginx/tests/unit/test_nginx_config_write.py
```diff
@@ -0,0 +1,25 @@
...
+
+from charmlibs.nginx import Nginx, NginxConfig
+from charmlibs.nginx.nginx import NGINX_CONFIG
+
+if typing.TYPE_CHECKING:
+    import ops
+
+
+def test_nginx_config_written(ctx, null_state):
```
james-garner-canonical, line 14:
```suggestion
def test_nginx_config_written(ctx: ops.testing.Context, null_state: ops.testing.State):
```

nginx/src/charmlibs/nginx/config.py
```diff
@@ -0,0 +1,627 @@
...
+              basic Nginx metrics.
+            supported_tls_versions: list of supported tls protocol versions.
+            ssl_ciphers: ssl ciphers.
+            worker_processes: Number of nginx worker processes to spawn.
+            worker_connections: Max number of worker connections
+            proxy_read_timeout: Proxy read timeout.
+            proxy_connect_timeout: Proxy connect timeout.
+
+        Example:
+            .. code-block:: python
```
james-garner-canonical, line 221:
```suggestion
        Example::
```
I think a trailing `::` does the trick here. The indentation below needs to be fixed too tho. Not currently a docs build error, since we don't include `__init__` docstrings, only the class docstring -- so maybe move this docstring to the class?

nginx/src/charmlibs/nginx/config.py
```diff
@@ -0,0 +1,627 @@
+# Copyright 2025 Canonical
+# See LICENSE file for licensing details.
+"""NginxConfig.
```
james-garner-canonical, line 3: This nice docstring currently isn't included in the reference docs, as this module isn't re-exported in `__init__.py`. Presumably that's the solution. An alternative would be to include an `automodule` directive in the `__init__.py` docstring.

nginx/src/charmlibs/nginx/nginx_prometheus_exporter.py
```diff
@@ -0,0 +1,48 @@
...
+            'summary': 'nginx prometheus exporter layer',
+            'description': 'pebble config layer for Nginx Prometheus exporter',
+            'services': {
+                'nginx-prometheus-exporter': {
+                    'override': 'replace',
+                    'summary': 'nginx prometheus exporter',
+                    'command': f'nginx-prometheus-exporter '
+                    f'--no-nginx.ssl-verify '
+                    f'--web.listen-address=:{self._nginx_prometheus_exporter_port} '
+                    f'--nginx.scrape-uri={scheme}://127.0.0.1:{self._nginx_port}/status',
```
james-garner-canonical, line 44: Maybe just personal preference, but I find this more readable.
```suggestion
                    'command': (
                        f'nginx-prometheus-exporter '
                        f'--no-nginx.ssl-verify '
                        f'--web.listen-address=:{self._nginx_prometheus_exporter_port} '
                        f'--nginx.scrape-uri={scheme}://127.0.0.1:{self._nginx_port}/status'
                    ),
```

nginx/tests/integration/conftest.py
```diff
@@ -0,0 +1,38 @@
...
+        juju.deploy(
+            _get_packed_charm_path(charm),
+            resources={
+                'nginx': 'ghcr.io/canonical/nginx@sha256:6415a2c5f25f1d313c87315a681bdc84be80'
+                'f3c79c304c6744737f9b34207993',
+                'nginx-pexp': 'ubuntu/nginx-prometheus-exporter:1.4-24.04_stable',
+            },
+        )
+    elif charm == 'machine':
+        juju.deploy(_get_packed_charm_path(charm))
```
james-garner-canonical, line 32: Could drop this case since this library is K8s only.

nginx/tests/unit/test_container_managers.py
PietroPasotti: I'll meet you halfway and `import ops.testing as scenario` :P

nginx/pyproject.toml
```diff
@@ -0,0 +1,64 @@
+[project]
+name = "charmlibs-nginx"
```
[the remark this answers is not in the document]
james-garner-canonical: To be clear, I don't mean adding more general functionality now, but rather being open to the idea that if we wanted such functionality later, it would go in this library. I'm totally happy with `charmlibs-nginx-k8s` though if that's your preference.

nginx/src/charmlibs/nginx/_config.py
```diff
@@ -0,0 +1,638 @@
...
+
+
+class NginxConfig:
+    """NginxConfig.
+
+    To generate an Nginx configuration for a charm, instantiate the `NginxConfig` class with the
+      required inputs:
+
+    1. `server_name`: The name of the server (e.g. charm fqdn), which is used to identify the
+      server in Nginx configurations.
```
james-garner-canonical, line 114:
```suggestion
    1. `server_name`: The name of the server (e.g. charm fqdn), which is used to identify the
       server in Nginx configurations.
```
I think we need one more space to appease Sphinx.

nginx/tests/integration/pack.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+set -xueo pipefail
+cd charms
+
+cd ./k8s
+uv lock  # required by uv charm plugin
+charmcraft pack
+mv *.charm k8s.charm  # rename to a predictable name
+cd ..
```
james-garner-canonical, line 9: This fails because the `nginx` library isn't available at lock time -- and you also need to watch out for how to make it available at packing time. My usual approach is to have symlinks to the library files checked in on disk (`src`, `pyproject.toml` and `README.md`) and use `cp --recursive --dereference` to copy the charm to a temporary directory for packing and resolve the symlinks at the same time ([example](https://github.com/canonical/charmlibs/blob/main/.example/tests/integration/pack.sh)).

nginx/pyproject.toml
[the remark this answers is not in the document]
PietroPasotti: so I'm open to the idea, but also I don't want (my team) to be on the hook for gatekeeping those additions if/when they are requested.
I don't see us ever needing non-k8s nginx helpers, and if another team does, I'd rather them not be in this library. Or us needing to be involved in their development.
Does that make sense?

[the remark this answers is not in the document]
james-garner-canonical: `charmlibs-nginx-k8s` sounds like a very reasonably scope in that case, thanks

mmkay (commented):
From the product side it looks good! I've got a few remarks about things that might be either different or moving in parallel in cos-coordinated-workers. I'm however positive on where we are with this :)

nginx_k8s/README.md
```diff
@@ -0,0 +1,64 @@
...
+import ops
+
+from charmlibs import nginx_k8s
+
+
+class MyCharm(ops.CharmBase):
+    def __init__(self, framework: ops.Framework):
+        super().__init__(framework)
+        self._nginx = nginx_k8s.Nginx(
+            self.unit.get_container("nginx-container"),
```
PietroPasotti, line 29:
```suggestion
            self.unit.get_container("nginx"),  # container name as defined in charmcraft.yaml
```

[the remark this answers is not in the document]
PietroPasotti: will add a comment instead

nginx_k8s/pyproject.toml
```diff
@@ -0,0 +1,65 @@
...
+requires-python = ">=3.10"
+authors = [
+    {name="The Observability team at Canonical Ltd."},
+]
+classifiers = [
+    "Programming Language :: Python :: 3",
+    "License :: OSI Approved :: Apache Software License",
+    "Intended Audience :: Developers",
+    "Operating System :: POSIX :: Linux",
+    "Development Status :: 5 - Production/Stable",
```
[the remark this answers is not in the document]
PietroPasotti: fair point. I honestly never cared a lot about the classifiers; copied this verbatim from the template.
```suggestion
    "Development Status :: 3 - Alpha",
```

nginx_k8s/pyproject.toml
```diff
@@ -0,0 +1,65 @@
...
+integration = [
+    "jubilant",
+    "pytest-jubilant"
+]
+
+[project.urls]
+"Documentation" = "https://documentation.ubuntu.com/charmlibs"
+"Repository" = "https://github.com/canonical/charmlibs"
+"Issues" = "https://github.com/canonical/charmlibs/issues"
+"Changelog" = "https://github.com/canonical/charmlibs/blob/main/nginx/CHANGELOG.md"
```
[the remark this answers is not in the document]
PietroPasotti: mmm, will remove it altogether (I hope we're not forced to keep/use a changelog and can rely instead on the release notes)

mmkay (approved):
LGTM!

PR #132, opened 2025-08-29
docs: try building package reference docs individually

james-garner-canonical:
This PR changes the `just docs html` recipe to incrementally build the package reference docs, by running `sphinx-build` once for each package, with that package installed in the virtual environment. The local `package_docs` extension has been updated accordingly to only create the automodule rst file for the specified package, rather than for all packages at once. Together, this completely eliminates the need for package dependencies to be added to the docs `requirements.txt` or to `autodoc_mock_imports`.

As a developer convenience, `just docs html` can be invoked with specific package names to limit reference docs generation to only the specified packages.

The `just docs run` and `just docs linkcheck` recipes have not been modified. The `package_docs` extension will skip generating the package reference docs when these recipes run (since they don't set the `package` config option). This means that they work mostly as expected, but won't create (or check) package reference docs.

### DIFF
--- a/.docs/extensions/package_docs.py
+++ b/.docs/extensions/package_docs.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import pathlib
+import shutil
 import typing
 
 ####################
@@ -26,22 +27,38 @@ import typing
 if typing.TYPE_CHECKING:
     import sphinx.application
 
+COMBINE_DOCS = '__combine__'
+
 
 def setup(app: sphinx.application.Sphinx) -> dict[str, str | bool]:
     """Entrypoint for Sphinx extensions, connects generation code to Sphinx event."""
-    app.connect('builder-inited', _package_docs)
+    app.connect('builder-inited', _builder_inited)
+    app.connect('build-finished', _build_finished)
+    app.add_config_value('package', default=None, rebuild='')
+    print('setup')
     return {'version': '1.0.0', 'parallel_read_safe': False, 'parallel_write_safe': False}
 
 
-def _package_docs(app: sphinx.application.Sphinx) -> None:
-    _main(pathlib.Path(app.confdir))
+def _builder_inited(app: sphinx.application.Sphinx) -> None:
+    print('builder-inited start')
+    _package_docs_rst(docs_dir=pathlib.Path(app.confdir), package=app.config.package)
+    print('builder-inited end')
 
 
-####################
-# generation logic #
-####################
+def _build_finished(app: sphinx.application.Sphinx, exception: Exception | None) -> None:
+    print('build-finished start')
+    if exception:
+        print(exception)
+        return
+    _package_docs_html(build_dir=pathlib.Path(app.outdir), package=app.config.package)
+    print('build-finished end')
 
-AUTOMODULE_TEMPLATE = """
+
+########################
+# rst generation logic #
+########################
+
+PACKAGE_DOCS_TEMPLATE = """
 .. raw:: html
 
    <style>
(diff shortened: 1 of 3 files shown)


james-garner-canonical:
Thanks for looking at this.

To answer your question, we don't need to specify ops in the docs
dependencies because we generate the pathops reference docs in a
virtual environment with pathops installed, which means that pathops'
dependencies like ops are installed too. So we're getting ops from pathops'
dependencies rather than specifying it in the docs dependencies explicitly.

As to why running the build a second time fails, it's because on the second
run, the rst automodule files are present for all the packages due to the
first build, and we don't clean them out before running the build with just
apt installed, and because of the ops automodule file it tries to build the
ops reference docs at that point. On the first build, this wasn't an issue,
because there was no automodule file for ops when apt was being built.

But I wonder if this always happens when running just docs the second time,
or only if we touch the pathops package in between. If the former, then I
guess we should always clean out the automodule files first. If the latter,
then maybe the current implementation is OK, since we can avoid the failed
build by running clean if we touch some packages, or running just docs html
<package> explicitly if we only touched one package. I'll look into this
further.

docs.just
```diff
@@ -2,18 +2,65 @@ set working-directory := '.docs'
...
+        *sorted(p.name for p in ROOT.glob(r'[a-z]*') if p.is_dir() and p.name != 'interfaces'),
+        *sorted(f'interfaces/{p.name}' for p in (ROOT / 'interfaces').glob(r'[a-z]*') if p.is_dir()),
+    ]
+    # Suppress ref warnings on all but last package as refs can't all resolve.
+    for package in packages[: -1]:
+        build(package, suppress_warnings=True)
+    # Run with the usual warnings for the last package.
+    # Or run just once with package=None if there were no packages.
+    for package in packages[-1 :] or [None]:
+        build(package)
```
dwilding, line 51: Minor: If `packages` is empty, should that raise an error rather than trying to build? If we assume that the repo will always contain packages, I don't see how `packages` can ever be empty. (Unless there's a way I've missed that `just docs html <something>` can force it to be empty)

So, in that case, would it be possible to simplify `build()` to always expect a package?

docs.just
```diff
@@ -2,18 +2,65 @@ set working-directory := '.docs'
...
 
-[doc('Watch, build, and serve the docs.')]
-run:
+[doc('Show help for the docs recipes.')]
+help:
+    @just --list docs --unsorted
+
+[doc('Watch, build, and serve the docs -- does not include package reference docs.')]
+run: _clean-charmlibs-reference-docs
```
dwilding, line 58: I'm OK with `run` and `linkcheck` not working for package reference docs. I think that's an acceptable compromise, which we can revisit in the future if it becomes a problem

.docs/extensions/package_docs.py
```diff
@@ -30,11 +30,14 @@
...
     """Entrypoint for Sphinx extensions, connects generation code to Sphinx event."""
     app.connect('builder-inited', _package_docs)
+    app.add_config_value('package', default=None, rebuild='')
     return {'version': '1.0.0', 'parallel_read_safe': False, 'parallel_write_safe': False}
 
 
 def _package_docs(app: sphinx.application.Sphinx) -> None:
-    _main(pathlib.Path(app.confdir))
+    package = app.config.package
+    if package is not None:
```
dwilding, line 39: Minor: I think it would be helpful to add a comment about how to interpret the None case. If I understand correctly, None will only happen if we're doing a run/linkcheck build.

docs.just
```diff
@@ -2,18 +2,65 @@ set working-directory := '.docs'
...
+    def build(package: str | None, suppress_warnings: bool = False) -> None:
+        cmd = [
+            'uvx',
+            '--from', 'sphinx',
+            '--with-requirements', '.sphinx/requirements.txt',
+            *(['--with', str(ROOT / package)] if package else []),
+            'sphinx-build',
+            '-T', '-W', '--keep-going',
+            '-b', 'dirhtml',
+            '-d', '_build/doctrees',
```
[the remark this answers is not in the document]
james-garner-canonical: Is this the only thing we need to change for this change? I notice linkcheck and run don't use this argument.

james-garner-canonical, line 27:
```suggestion
            '-d', '.sphinx/.doctrees',
```

.docs/extensions/package_docs.py
james-garner-canonical:
```suggestion
    # generate reference docs for the current package
    # package is None if not set explicitly
    # e.g. when running `just docs run` or `just docs linkcheck`
    if package is not None:
```

docs.just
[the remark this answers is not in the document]
james-garner-canonical: Done =)

docs.just
```diff
@@ -2,18 +2,65 @@ set working-directory := '.docs'
...
+# this is the first recipe in the file, so it will run if just is called without a recipe
+[doc("""
+Build the docs, generating reference docs for all packages, or just those specified.
+
+`just docs` is an alias for `just docs html`. To specify packages, you must use the full recipe, e.g.
+`just docs html interfaces/foo`.
+Run `just docs clean` first to generate reference docs from scratch if you run into import errors.
+""")]
+html *packages: _clean-charmlibs-reference-docs
+    #!/usr/bin/env -S uv run --script --no-project
```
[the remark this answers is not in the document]
dwilding: Oh cool, I didn't know that, thanks :slightly_smiling_face:

PR #133, opened 2025-08-31
ci: add CODEOWNERS file

james-garner-canonical:
This PR adds the `CODEOWNERS` file that will be used to allow library owners to approve PRs that affect just their libraries. `@canonical/charm-tech` is set as the default owner for the repository (`.`), as well as set explicitly for the `CODEOWNERS` file itself, the repository infra files (e.g. `/pyproject.toml`, `/justfile`), global docs, and the packages that we currently own. PRs adding a new library will add a `CODEOWNERS` entry for its package directory, which will require `@canonical/charm-tech` approval. After that, the owners of the package directory will be able to approve changes on their own.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,29 @@
+# default
+. @canonical/charm-tech
+/interfaces/ @canonical/charm-tech
+
+# codeowners
+/.github/CODEOWNERS @canonical/charm-tech
+
+# infra
+/.github/ @canonical/charm-tech
+/docs.just @canonical/charm-tech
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Nice! Will be more fun when there's something other than charm-tech here :-)

.github/CODEOWNERS
```diff
@@ -0,0 +1,29 @@
+# default
+. @canonical/charm-tech
```
dimaqq, line 2: There's a lot of redundancy here, but I guess that's the point.

dimaqq (approved):

PR #139, opened 2025-09-01
ci: install test deps into individual package venvs and respect uv.lock

james-garner-canonical:
Currently, we run all `just` recipes using the top-level virtual environment `.venv`. This isn't a great user experience, as the virtual environment is constantly changing under you. Additionally, we don't commit `uv.lock` files, because: (1) at the top-level, it would change constantly, (2) for the packages, we don't use it in CI, and (3) libraries will potentially be installed with different deps at runtime anyway, since `uv.lock` has no effect after packaging.

This PR modifies the `justfile` to:
1. use `<package>/.venv` when running recipes that affect a single package
2. use `uv export` to install package dependencies + top-level test dependencies, so the state of `<package>/uv.lock` matches the actual dependencies used -- `uv export` + `uv pip install` is necessary because there doesn't seem to be a way to include the top-level test dependencies when running `uv sync` or `uv run` currently
3. only install the `fast-lint` dependencies in the top-level `.venv` -- this will then only change when we upgrade those dependencies


benhoyt (approved):
Seems reasonable to me, thanks.

PR #147, opened 2025-09-02
ci: define ci workflow concurrency

james-garner-canonical:
This PR defines `concurrency` for the `ci` workflow, with `cancel-in-progress`:
- Each `pull_request` uses a single concurrency group, so pushing to a PR will cancel the in progress tests and start new ones.
- Each `push` to `main` uses a unique concurrency group (commit hash), so all commits to main are tested.
- Each `workflow_dispatch` uses a unique concurrency group (run ID), so all manually run workflows will only be cancelled manually.


james-garner-canonical:
Closing in favour of #153 -- we won't add concurrency for now.

PR #154, opened 2025-09-04
ci: simplify (and correct?) handling of dependency groups

james-garner-canonical:
It's painful trying to use dependency groups from both the repo-level `pyproject.toml` and the individual package `pyproject.toml`. We currently work around this by dropping down to the `uv venv` and `uv pip` commands instead of using `uv run`, which is a bit nasty, and may not actually be working properly.

If we instead specify the tool versions in the `justfile`, things are a lot simpler -- we can `uv run` using `--with` for our tool versions and `--group` for the package's dependency group(s). This should also make it clearer to library contributors where the `pyright` (etc) versions are coming from.

This requires packages to specify all dependency groups, empty if they don't need to put anything in them (which is how the template will start new projects). This PR updates the existing packages to do this.

### DIFF
--- a/.package/pyproject.toml
+++ b/.package/pyproject.toml
@@ -15,6 +15,12 @@ classifiers = [
 ]
 dynamic = ["version"]
 
+[dependency-groups]
+lint = []
+unit = []
+functional = []
+integration = []
+
 [project.urls]
 "Repository" = "https://github.com/canonical/charmtech-charmlibs"
 "Issues" = "https://github.com/canonical/charmtech-charmlibs/issues"
--- a/.package/uv.lock
+++ b/.package/uv.lock
@@ -1,0 +1,15 @@
+version = 1
+revision = 1
+requires-python = ">=3.8"
+
+[[package]]
+name = "charmlibs"
+source = { editable = "." }
+
+[package.metadata]
+
+[package.metadata.requires-dev]
+functional = []
+integration = []
+lint = []
+unit = []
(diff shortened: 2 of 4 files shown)


justfile
```diff
@@ -5,18 +5,17 @@ set ignore-comments  # don't print comment lines in recipes
...
 [doc('Describe usage and list the available recipes.')]
 _help:
     @echo 'All recipes require {{CYAN}}`uv`{{NORMAL}} to be available.'
     @just --list --unsorted --list-submodules
 
-[doc('Run `uv add` for package, respecting the global test dependency constraints.')]
-add package +args:
-    #!/usr/bin/env -S bash -xueo pipefail
-    cd '{{package}}'
-    uv add {{args}} --constraints=<( cd '{{justfile_directory()}}' && uv export --group=test )
```
james-garner-canonical, line 18: Constraints aren't respected for the package being added. For example, if the constraints file specifies `pyright==X`, `uv add pyright --constraints=...` will still install the latest version of `pyright`. On the other hand, if the constraints file specifies the version of a *dependency* of `pyright` (e.g. `nodeenv`), the installed version of `pyright` will be selected to respect this. Since the main purpose of this command would be to avoid adding a version of the test deps that clashes, this makes this command seem not-very-useful to me.

tonyandrewmeyer (approved):
Seems reasonable to me. Particularly since there's a template, it's not much of a burden to have the groups.

PR #155, opened 2025-09-08
docs: support Python domain links between package docs

james-garner-canonical:
WIP: not working correctly yet!


james-garner-canonical:
Closing until there's the opportunity to revisit this problem -- the approach tried in this PR is flawed, and it may be that we eventually take an entirely different approach to building the package docs.

PR #158, opened 2025-09-15
ci: add just add command (again)

james-garner-canonical:
In a previous PR I suggested that we didn't need `just add`, because `uv add --constraints` didn't respect the constraints for the package being added -- that is, if `pytest` is in our constraints with some version, `uv add --constraints=$FILE pytest` will still install the latest version of `pytest`. But it does seem important to provide a way for library authors to add their own dependencies while respecting our test package versions (this became clear while writing the tutorial in #159).

This PR achieves this by moving the test dependencies to a `requirements.txt` format file suitable for use with both `uv run --with-requirements` and `uv add --constraints`, which also feels like a more idiomatic way to specify dependency versions. `uv run` invocations are updated accordingly.

Additionally, this PR introduces more robust variadic argument handling via the `positional-arguments` directive, out of necessity for `just add` (see comments), and as a driveby fix for `just static`.

### DIFF
--- a/justfile
+++ b/justfile
@@ -32,21 +32,25 @@ format:
     uv run --only-group=fast-lint ruff check --preview --fix
 
 [doc("Run `uv add` for package, respecting repo-level version constraints, e.g. `just add pathops 'pydantic>=2'`.")]
+[positional-arguments]  # pass recipe args to recipe script positionally (so we can get correct quoting)
 add package +args:
     #!/usr/bin/env -S bash -xueo pipefail
+    shift 1  # drop $1 (package) from $@ it's just +args
     cd '{{package}}'
-    uv add --constraints {{quote(join(justfile_dir(), 'test-requirements.txt'))}} {{args}}
+    uv add --constraints {{quote(join(justfile_dir(), 'test-requirements.txt'))}} "${@}"
 
 [doc('Run global `fast-lint` and package specific `static` analysis, e.g. `just python=3.10 lint pathops`.')]
 lint package *pyright_args: fast-lint (static package pyright_args)
 
 [doc('Run package specific static analysis only, e.g. `just python=3.10 static pathops`.')]
+[positional-arguments]  # pass recipe args to recipe script positionally (so we can get correct quoting)
 static package *args:
     #!/usr/bin/env -S bash -xueo pipefail
+    shift 1  # drop $1 (package) from $@ it's just *args
     cd '{{package}}'
     {{_uv_run_with_test_requirements}} \
         --group lint --group unit --group functional --group integration \
-        pyright --pythonversion='{{python}}' {{args}}
+        pyright --pythonversion='{{python}}' "${@}"
 
 [doc("Run unit tests with `coverage`, e.g. `just python=3.10 unit pathops`.")]
 unit package +flags='-rA': (_coverage package 'unit' flags)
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -31,19 +28,29 @@ fast-lint:
...
+    uv run --only-group=fast-lint ruff format --preview
+    uv run --only-group=fast-lint ruff check --preview --fix
+
+[doc("Run `uv add` for package, respecting repo-level version constraints, e.g. `just add pathops 'pydantic>=2'`.")]
+[positional-arguments]  # pass recipe args to recipe script positionally (so we can get correct quoting)
+add package +args:
+    #!/usr/bin/env -S bash -xueo pipefail
+    shift 1  # drop $1 (package) from $@ it's just +args
+    cd '{{package}}'
+    uv add --constraints {{quote(join(justfile_dir(), 'test-requirements.txt'))}} "${@}"
```
james-garner-canonical, line 40: `just` writes the recipe script to a file after templating it (expanding `{{values}}`). The `positional-arguments` directive means that the recipe name and arguments are *also* passed to the script as positional arguments when it's executed (`$0` is the recipe name, `$1` is the first argument, and so on. `$@` is arguments `$1` onwards).

We need to use this here because at templating time, variadic arguments are resolved to a single string (while when passed positionally, argument separation is properly preserved). For example, `just add pathops 'pydantic>=2'` leaves us with package` = `pathops`, and `args` = `pydantic>=2`. If we write `{{args}}` into the script, we literally write `pydantic>=2` into the script -- that is `>` for file redirection, and `=2` for the target file. If we write `'{{args}}'` into the file (with enclosing quotes), then we can't handle multiple args, for example `--editable pydantic>=2` would be treated as a single argument instead of two.

You *can* resolve this to an extent by double quoting, e.g. `"'pydantic>=2'"`, but that's not a very good user experience, and the failure mode when you forget the extra quotes is unpredictable (e.g. silently ignoring a constraint and redirecting output to a file).

As far as I can tell, the best way to handle variadic arguments with correct separation is via `positional-arguments`. After calling `shift` appropriately, `"${@}"` correctly separates and passes the arguments at calling time.

justfile
```diff
@@ -31,19 +28,29 @@ fast-lint:
 
 [doc('Run `ruff check --fix` and `ruff --format`, modifying files in place.')]
 format:
-    uv run ruff format --preview
-    uv run ruff check --preview --fix
+    uv run --only-group=fast-lint ruff format --preview
+    uv run --only-group=fast-lint ruff check --preview --fix
```
james-garner-canonical, line 32: Driveby fix to specify the dependencies correctly for `just format`.

justfile
```diff
@@ -31,19 +28,29 @@ fast-lint:
...
+[positional-arguments]  # pass recipe args to recipe script positionally (so we can get correct quoting)
 static package *args:
     #!/usr/bin/env -S bash -xueo pipefail
+    shift 1  # drop $1 (package) from $@ it's just *args
     cd '{{package}}'
-    uv run {{_with_test_deps}} \
+    {{_uv_run_with_test_requirements}} \
         --group lint --group unit --group functional --group integration \
-        pyright --pythonversion='{{python}}' {{args}}
+        pyright --pythonversion='{{python}}' "${@}"
```
james-garner-canonical, line 53: Driveyby fix to apply the `positional-arguments` solution to `static` as well.

justfile
```diff
@@ -7,10 +7,7 @@ python := '3.10'
 # for integration tests, e.g. `just tag=24.04 pack-k8s` `just tag=foo integration-machine`
 tag := env('CHARMLIBS_TAG', '')
 
-_coverage := 'coverage==7.6.1'
-_pyright := 'pyright==1.1.397'
-_pytest := 'pytest==8.3.5'
-_with_test_deps := '--with ' + _coverage + ' --with ' + _pyright + ' --with ' + _pytest
+_uv_run_with_test_requirements := 'uv run --with-requirements ' + quote(join(justfile_dir(), 'test-requirements.txt'))
```
james-garner-canonical, line 10: Specifying these in a `requirements.txt` format file seems more idiomatic, and allows us to use the same file for the constraints for `uv add`.

tonyandrewmeyer: I imagine it's also more likely that scanning tools (like dependabot) will find them.

tonyandrewmeyer (approved):
This makes sense to me. It does seem like the just file is getting increasingly complex, which is a bit concerning, but perhaps it's not going to get worse than this.

justfile
james-garner-canonical: Great point, thanks!

PR #159, opened 2025-09-15
docs: add tutorial

james-garner-canonical:


dwilding:
Before I start reviewing - The link checker doesn't like `https://matrix.to/#/#charmhub-charmdev:ubuntu.com` because the URL fragment doesn't exist as the ID of a page element (it's read by JavaScript instead). You should be able to exclude this URL in `conf.py`. See the [starter pack docs](https://canonical-starter-pack.readthedocs-hosted.com/latest/reference/automatic_checks_links/). I haven't tried this yet, as in Ops we don't have the link checker enabled yet.

.docs/tutorial.md
```diff
@@ -0,0 +1,354 @@
...
+- `pebble` is a list of [Pebble version tags](https://github.com/canonical/pebble/tags), for example `["v1.24.0"]`.
+    - If specified, the Pebble version will be an additional row in the matrix your functional tests are run with, with the corresponding Pebble version installed.
+    - By default, your tests are run once without Pebble installed.
+    - For local testing, you'll need to install the version of Pebble that you want to test against yourself. You can then run `just functional-pebble <YOUR LIBRARY DIRECTORY>` instead of `just functional <YOUR LIBRARY DIRECTORY>` to have Pebble started and stopped before and after your tests.
+- `sudo` is a `true` or `false` option, defaulting to `false`.
+    - If it's `true`, your tests are run in CI with `sudo` permissions in their ephemeral runner.
+    - If your tests require `sudo` privileges, you'll need to manage that locally yourself, for example by testing in a virtual machine.
+
+If your library needs another piece of software like `pebble` installed for its functional tests, get in touch with us on [Matrix](https://matrix.to/#/#charmhub-charmdev:ubuntu.com) or open an issue about the option to install it in CI.
```
dwilding, line 240: I was a bit confused by this, as Pebble is already covered above. I think this would be fine:

```suggestion
If your library needs a specific piece of software for CI to run its functional tests, get in touch with us on [Matrix](https://matrix.to/#/#charmhub-charmdev:ubuntu.com) or open an issue in the `charmlibs` repository.
```

.docs/tutorial.md
```diff
@@ -0,0 +1,354 @@
...
+
+In our case, we don't need any extra software, we don't need `sudo`, and we don't expect any differences across Ubuntu versions, so we can just keep the default configuration.
```
dwilding, line 242: A couple of minor suggestions. Since we diverted away from talking about `uptime`, I think it would be good to mention it again

```suggestion
For our `uptime` library, we don't need any extra software, we don't need `sudo`, and we don't expect any differences across Ubuntu versions, so we can keep the default configuration.
```

.docs/tutorial.md
```diff
@@ -0,0 +1,354 @@
...
+
+Let's add an integration test for our `uptime` function.
+
+We'll start by taking a look at the files that will make up our packed charm, under `tests/integration/charms`.
+At the top level are directories for two test charms, with the directory name reflecting the substrate the charm is for: `k8s` and `machine`.
+You'll also see some common files which are symlinked into the structure for our two test charms -- these symlinks are resolved by `pack.sh` before `charmcraft pack` is executed.
+Taking a look inside one of the charm directories, you can see these symlinks, as well as a unique `charmcraft.yaml` file, and the usual `src/` directory.
+There's also a directory named `package/`, which contains symlinks to your library code and metadata -- this is how your library is made available to the charms.
```
dwilding, line 267: I can't see a directory named `package` - maybe I'm looking in the wrong place

.docs/tutorial.md
```diff
@@ -0,0 +1,354 @@
...
+Our `uptime` function should work just as well in a K8s charm as in a machine charm, so we'll test on both substrates, and we won't mark any of our tests as `k8s_only` or `machine_only`.
+
+We'll communicate with the library in our packed charm via a Juju action.
+Open `tests/integration/charms/actions.yaml` and add a new action:
+```yaml
+charm-uptime:
+```
+
+We'll also need an observer for this action, which can be the same for both charms.
+It will look a lot like the handler for `lib-version`, `_on_lib_version`, but we'll serialize the fancy version information as a JSON object so it can be passed over the wire to our test code.
```
dwilding, line 282: This part still mentions the fancy version from an earlier revision of the tutorial

.docs/tutorial.md
```diff
@@ -0,0 +1,354 @@
...
+Start by running `just lint uptime` from anywhere in the repository.
+(`just` commands execute from the `justfile` directory, regardless of where they're run.)
+This will run `ruff`, `codespell`, and `pyright` on your library, all of which should pass.
+This will also create a `uv.lock` file in your library's directory, which should be included in version control.
+
+Then run `just unit uptime` to verify that your unit tests pass, and `just functional uptime` to check the functional tests.
+We'll talk more about Juju integration tests later.
+
+You can also verify that everything looks right in an interactive session.
+Run `uvx --with=./uptime ipython --pdb` in the repository root to start an interactive Python shell with your library installed, and then import your library with:
```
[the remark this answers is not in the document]
james-garner-canonical: I'm wondering if a `just shell uptime` command like [this](https://github.com/canonical/charmlibs/pull/171/files) would be worthwhile.

.docs/tutorial.md
james-garner-canonical: I made this change in #172, I'll update the tutorial accordingly.

.docs/tutorial.md
```diff
@@ -0,0 +1,330 @@
...
+charm-uptime:
+```
+
+We'll also need an observer for this action, which can be the same for both charms, so it can go in the `Charm` base class in `common.py`.
+It will look a lot like the handler for `lib-version`, `_on_lib_version`, but we'll serialize the result as a JSON object to preserve its type for our test code.
+Open `tests/integraton/charms/common.py` and add this import statement:
+```python
+import json
+```
+Then add an observer for our new action to the `Charm` class body:
```
dwilding, line 281: We also need to add `framework.observe(self.on['charm-uptime'].action, self._on_charm_uptime)` in `__init__`.

.docs/tutorial.md
```diff
@@ -0,0 +1,330 @@
...
+    results = {'uptime': json.dumps(uptime.uptime().total_seconds())}
+    event.set_results(results)
+    logger.info('action [charm-uptime] set_results: %s', results)
+```
+
+Finally, we'll need a test to exercise this code. Copy `tests/integration/test_version.py`, to `tests/integration/test_uptime.py`, and add this import to `test_uptime.py`:
+```python
+import json
+```
+Then replace the `test_version` function with this:
```
dwilding, line 294: I needed to remove `from charmlibs import uptime` from `test_uptime.py`, otherwise linting failed due to the unused import.

.docs/explanation/charmlibs-tests.md
```diff
@@ -0,0 +1,65 @@
+(charmlibs-tests)=
+# Types of tests in the charmlibs monorepo
+
+In addition to static analysis, the `charmlibs` monorepo supports running three distinct kinds of tests: unit, functional, and integration tests.
+These are all optional from the perspective of the repository infrastructure, and will be skipped if the corresponding `tests/` directory does not exist.
+For example, if `<LIBRARY>/tests/functional` does not exist, then the functional testing job will be skipped for `<LIBRARY>`.
+
+`just <TESTS> <LIBRARY>` will run the tests under `<LIBRARY/tests/<TESTS>` with `pytest`.
```
dwilding, line 8:
```suggestion
`just <TESTS> <LIBRARY>` will run the tests under `<LIBRARY>/tests/<TESTS>` with `pytest`.
```

.docs/explanation/charmlibs-tests.md
```diff
@@ -0,0 +1,65 @@
...
+`<TESTS>` is one of unit, functional, or integration.
+`<LIBRARY>` is the path from the repository root to the library, for example `pathops` or `interfaces/.example`.
+
+Extra arguments are passed to `pytest`, so you can, for example, use `-k` to select specific tests, or pass `-x` to exit early on failure.
```
dwilding, line 12: Should the extra arguments come after `--` or directly after `<LIBRARY>`?

dwilding, line 12: Minor - I'd probably break the sentence:

```suggestion
Extra arguments are passed to `pytest`. For example, you can use `-k` to select specific tests, or `-x` to exit early on failure.
```

.docs/explanation/charmlibs-tests.md
```diff
@@ -0,0 +1,65 @@
...
+The specific version of `pytest` to run is specified in the repository's `test-requirements.txt`.
+
+```{tip}
+`just` command can be invoked from anywhere in the repository, as they always execute from the location of the `justfile`.
```
dwilding, line 16:
```suggestion
`just` commands can be invoked from anywhere in the repository, as they always execute from the location of the `justfile`.
```

.docs/explanation/charmlibs-tests.md
```diff
@@ -0,0 +1,65 @@
...
+```bash
+just unit <LIBRARY>
+```
+
+## Functional tests
+
+Functional tests are intended to be end-to-end tests of everything except the real Juju environment.
+Interacting with Juju itself is reserved for integration tests.
+Functional tests are most useful for libraries that interact with some significant external component that can be decoupled from the Juju context.
+For example, the `apt` and `snap` packages interact with the Ubuntu system's package management, the majority of which will be the same regardless of whether they're being called from a (machine) charm or not.
```
dwilding, line 33: I didn't get this at first. How about some slight changes - something like:

```suggestion
For example, the `apt` and `snap` packages interact with the Ubuntu system's package management tools, which will broadly act the same regardless of whether they're being called from a charm. (For machine charms, at least.)
```

.docs/tutorial.md
```diff
@@ -0,0 +1,330 @@
...
+```python
+def test_charm_uptime(juju: jubilant.Juju, charm: str):
+    result = juju.run(f'{charm}/0', 'charm-uptime')
+    uptime_seconds = json.loads(result.results['uptime'])
+    assert uptime_seconds > 0.0
+```
+
+You can test this locally by running `just pack-k8s uptime` and `just pack-machine uptime` to pack your K8s and machine charms, and then running `just integration-k8s uptime` and `just integration-machine uptime` to deploy the charms on your local Juju K8s and machine clouds and test them.
```
dwilding, line 302: I didn't think of this the first time, but is it obvious that you'll need Charmcraft installed? I don't think it's worth mentioning here, but perhaps in the setup section at the beginning? As an optional step. Maybe even expanded to say that if you want to run integration tests locally, you'll need Charmcraft and local Juju K8s and machine clouds.

Just an idle thought - I'm fine if you want to leave this out.

.docs/how-to/customize-integration-tests.md
```diff
@@ -0,0 +1,27 @@
...
+> Read more: {ref}`charmlibs-tests`
+
+Integration tests are only executed if the `<LIBRARY>/tests/integration` directory exists.
+If you don't need integration tests, remove the directory.
+However, it's recommended to have at least some Juju integration tests to ensure that your library can be used as expected in a real charm.
+
+Your integration tests will be run in CI with a matrix of Juju substrates, and a list of `tags` defined in your `pyproject.toml` under:
+```toml
+[tool.charmlibs.integration]
+```
```
dwilding, line 17: I think for this how-to, and the how-to about functional tests, it would be useful to have a section at the end called "Examples". You could link to the configurations of a couple of a packages in the monorepo. Perhaps better left for a follow-up PR.

In the meantime, would it be possible to add a link at this point to show how `tags` might be set up? That's the only aspect that I can't immediately visualise.

dwilding (approved):
Awesome work on this tutorial!

.docs/explanation/charmlibs-tests.md
james-garner-canonical: Directly after `<LIBRARY>`.
````suggestion
Extra arguments are passed to [pytest](https://docs.pytest.org/en/6.2.x/usage.html). For example:
```bash
just unit pathops -x  # exit on first failure
```
````

Comment (james-garner-canonical) [reply .docs/explanation/charmlibs-tests.md:12]: I'll use a full example instead.

File: .docs/tutorial.md (hunk 6 above)
Comment (james-garner-canonical) [reply .docs/tutorial.md:281]: Good catch, adding below.

File: .docs/tutorial.md [hunk 14]
```diff
@@ -0,0 +1,330 @@
...
+```python
+import json
+```
+Then replace the `test_version` function with this:
+```python
+def test_charm_uptime(juju: jubilant.Juju, charm: str):
+    result = juju.run(f'{charm}/0', 'charm-uptime')
+    uptime_seconds = json.loads(result.results['uptime'])
+    assert uptime_seconds > 0.0
+```
```
james-garner-canonical, line 300:
````suggestion
```
You'll also want to remove the `from charmlibs import uptime` line since it's now unused, and linting will complain about the unused import.
````

File: .docs/tutorial.md (hunk 12 above)
Comment (james-garner-canonical) [reply .docs/tutorial.md:302]:
```suggestion
You can test this locally by running `just pack-k8s uptime` and `just pack-machine uptime` to pack your K8s and machine charms, and then running `just integration-k8s uptime` and `just integration-machine uptime` to deploy the charms on your local Juju K8s and machine clouds and test them.
Note that this will require `charmcraft` installed locally for packing, and a Juju controller available for the K8s or machine clouds.
```

.docs/tutorial.md
```diff
@@ -0,0 +1,330 @@
...
+```bash
+sudo snap install --classic astral-uv
+```
+
+This repository uses [just](https://github.com/casey/just) as a command runner. With `uv` installed, we can install `just` with:
+```bash
+uv tool install rust-just
+```
+
+If you're not already using some alternative authentication for `git push`, you'll also want to [make sure you've set up your Github account with your SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account).
```
james-garner-canonical, line 35:
```suggestion
If you want to run the Juju integration tests locally, you'll also need `charmcraft`installed for packing, as well as a Juju controller for your local K8s or machine clouds.
In CI, these are installed and set up for you using [concierge](https://github.com/canonical/concierge?tab=readme-ov-file#presets), with the `microk8s` and `machine` presets.
The `dev` preset is suitable for local development and testing of both K8s and machine charms.

> See more:
> - {ref}`Charmcraft | Install charmcraft <charmcraft:manage-charmcraft>`
> - {ref}`Juju | Set up your Juju deployment <juju:set-up-your-deployment>`
> - {ref}`Juju | Set up an isolated environment <juju:set-things-up>`

If you're not already using some alternative authentication for `git push`, you'll also want to [make sure you've set up your Github account with your SSH key](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account).
```

PR #160, opened 2025-09-16
feat: add snap lib

james-garner-canonical:
This PR moves the `operator_libs_linux.v2.snap` library to be a `charmlibs` Python package.

The code has been moved as-is from [the original](https://github.com/canonical/operator/blob/main/docs/tutorial/write-your-first-machine-charm.md) to the `charmlibs.snap._snap` module. It has then been linted and formatted to comply with this repositories configuration, 99% automatically. Public names defined in `_snap` are then re-exported in `__init__.py`.

The tests were also ported as-is, formatted automatically, and then had their imports and mocking/patching manually adjusted to continue passing.

---

The following manual changes were made to the library code:
- Used `Mapping` for the type of some arguments instead of `dict` to satisfy the type checker (complaining in the tests), and as a backwards compatible type checking improvement.
- Made the `ansi_filter` regex private (super tox says it's not used as of my last charm snapshot).

The following manual changes were made to the test code:
- The type of the revision argument to several functions was switched to `str` or `int` to satisfy the type checker. Maybe it would be better to use a `cast` or `type: ignore` here instead though, comments appreciated. This is the tip of a bigger issue -- different functions inconsistently accept `str` or `int` for the revision argument. I'd like to make those consistent in a backwards compatible way by having them all accept `str | int`, but perhaps that should be a separate PR (and separate minor version release).

I'm also wondering whether we should drop the `JSONType` return type in favour of `dict[str Any]`, probably separately.

---

The reference docs build correctly. A couple of things stand out:
1. I guess we should exclude the `TypedDict` [magic fields from the docs](https://canonical-ubuntu-documentation-library--160.com.readthedocs.build/charmlibs/reference/charmlibs/snap/#snap.SnapServiceDict) (separate PR, changes repo-level config). Though I'm not 100% sure -- they're a bit ugly, but is the information about `total=True/False` and which keys are required made available elsewhere?
2. What to do about JSON argument type annotatons regarding docs and regarding exporting type aliases? [Example](https://canonical-ubuntu-documentation-library--160.com.readthedocs.build/charmlibs/reference/charmlibs/snap/#snap.Snap.set). In the docs, recursive type annotations just bottom out after a couple of levels. Looks kind of ugly, but I guess that's just the state of the art for now. Unclear whether we should export `JSONAble` for users -- feels kind of too general for this library, idk -- and should be unnecessary, as the argument type should be compatible with inferred dictionary typing -- e.g. {'foo': 'bar'} will be inferred as `dict[str, str]` which is compatible with `JSONAble`. Whether we export types like this also applies to `JSONType` -- I'd say don't export for now (technically breaking change like making `ansi_filter` private, but I don't think the types are used in charms right now).


IronCore864 (approved):
Since most of the code is moved from the other repo, I had a quick look and spent most of the time trying to do some tests with it to see if it works or not.

With the following piece of script, I can test it's working, so approving now.

```python
import logging
from snap.src.charmlibs import snap

logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
logger = logging.getLogger(__name__)


def snap_installation():
    logger.info('=== Snap Installation Examples ===')

    try:
        cache = snap.SnapCache()
        s = cache['hello-world']
        s.ensure(state=snap.SnapState.Present, classic=True)
        logger.info('✓ Successfully installed hello-world snap')
    except snap.SnapError as e:
        logger.error(f'✗ Failed to install hello-world: {e}')


def snap_removal():
    logger.info('=== Snap Removal Examples ===')

    try:
        cache = snap.SnapCache()
        s = cache['hello-world']
        s.ensure(state=snap.SnapState.Absent)
        logger.info('✓ Successfully removed hello-world snap')
    except snap.SnapError as e:
        logger.error(f'✗ Failed to remove hello-world: {e}')


def main():
    logger.info('Starting snap management test...')
    logger.info('This script tests various snap management operations')
    logger.info('using the charmlibs.snap library.')

    try:
        snap_installation()
        snap_removal()

        logger.info('=== Test Complete ===')
        logger.info('All snap management operations have been tested.')

    except Exception as e:
        logger.error(f'Unexpected error during test: {e}')
        raise


if __name__ == '__main__':
    main()

```

Regarding the last two questions in the description, I could not contribute.

james-garner-canonical:
I'm going to merge this as-is now, and potentially tweak the type annotations + docs before the initial release.

PR #164, opened 2025-09-22
chore: clean up JSON type aliases

james-garner-canonical:
This PR simplifies the JSON type aliases used in the `snap` library. The return type alias is dropped in favour of `dict[str, Any]`, as I'm not convinced there's much value in union types here, as the user still needs to narrow them with assertions, etc. The argument type alias has had its name shortened (which looks nicer in the docs) and made private -- neither type alias was used in charm code as of a grep through all our known charms today.

### DIFF
--- a/snap/src/charmlibs/snap/_snap.py
+++ b/snap/src/charmlibs/snap/_snap.py
@@ -38,6 +38,7 @@ from datetime import datetime, timedelta, timezone
 from enum import Enum
 from subprocess import CalledProcessError, CompletedProcess
 from typing import (
+    Any,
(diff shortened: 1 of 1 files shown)


benhoyt (commented):
I'm not totally opposed to this, however, unless it's causing actual issues, I'd be inclined to leave it, and save our "breaking changes" money till when we rewrite the lib completely next cycle.

Don't get me wrong, I think it's a good change, but there's a slim change somebody will be referencing `JSONType` / `JSONAble`, and we'd be breaking that here without much benefit.

PR #184, opened 2025-09-30
ci: allow pyproject.toml in non package dirs

james-garner-canonical:
This PR updates `.scripts/ls.py` to use `tomllib` for two purposes:
1. Only count a directory as a package if the `pyproject.toml` file has a `project` table -- this will allow us to use `pyproject.toml` files for tool configuration in non-package directories (like interface directories that only have the interface definition).
2. Remove the `aliases` table and read the package's distribution name from the `pyproject.toml` file directly.

### DIFF
--- a/.scripts/ls.py
+++ b/.scripts/ls.py
@@ -211,7 +211,8 @@ def _get_version(root: pathlib.Path, package: pathlib.Path | str) -> str | None:
     if not (root / package).exists():
         return None
     logger.debug('Computing version for %s', package)
-    dist_name = tomllib.loads((root / package / 'pyproject.toml').read_text())['project']['name']
+    with (root / package / 'pyproject.toml').open() as f:
+        dist_name = tomllib.load(f)['project']['name']
     script = f'import importlib.metadata; print(importlib.metadata.version("{dist_name}"))'
     cmd = ['uv', 'run', '--no-project', '--with', f'./{package}', 'python', '-c', script]
     return subprocess.check_output(cmd, cwd=root, text=True).strip()
--- a/.scripts/ls.py
+++ b/.scripts/ls.py
@@ -1,5 +1,11 @@
 #!/usr/bin/env -S uv run --script --no-project
 
+# /// script
+# requires-python = ">=3.11"
+# ///
+
+# ruff: noqa: I001  # tomllib is first-party in 3.11+
+
 # Copyright 2025 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


.scripts/ls.py
```diff
@@ -109,13 +116,18 @@ def _packages(include: list[str]) -> list[pathlib.Path]:
...
     """
     paths: set[pathlib.Path] = set()
     for root in _REPO_ROOT, _INTERFACES:
         paths.update(root.glob(r'[a-z]*'))
         paths.update(root / i for i in include)
-    return sorted(p.relative_to(_REPO_ROOT) for p in paths if (p / 'pyproject.toml').exists())
+    return sorted(
+        path.relative_to(_REPO_ROOT)
+        for path in paths
+        if (pyproject_toml := path / 'pyproject.toml').exists()
```
tonyandrewmeyer, line 128: Nit: I don't love things where you need to know the precedence of the operators (especially wacky ones, and here we have the crazy path division and the sad walrus). I assume from looking at this that `/` happens before `:=`, and testing that it's true, but this would be valid syntax if it was the other way around and you'd end up with a subtle bug. What about a couple of `()`?

.scripts/ls.py
```diff
@@ -199,17 +211,7 @@ def _get_version(root: pathlib.Path, package: pathlib.Path | str) -> str | None:
...
-        # placeholders
-        '.package': 'charmlibs',
-        'interfaces/.package': 'charmlibs-interfaces',
-        # examples
-        '.example': 'charmlibs-example',
-        'interfaces/.example': 'charmlibs-interfaces-example',
-        '.tutorial': 'charmlibs-uptime',
-    }
-    dist_name = aliases.get(str(package), f'charmlibs-{package}'.replace('/', '-'))
+    dist_name = tomllib.loads((root / package / 'pyproject.toml').read_text())['project']['name']
```
tonyandrewmeyer, line 214: Is there a reason this is `loads` with `read_text` rather than just a plain `load` with `open`? Is it to avoid using a context manager to get the close?

james-garner-canonical: I think that's fair, and it's quite a mouthful on one line, so I'll switch to the context manager.

.scripts/ls.py
james-garner-canonical: I guess the evidence is the mistake being made at all, rather than the precedence of `:=` being lower than both `/` and `is`?

I know it's not a massive deal either way, but I'm on the fence about whether adding parentheses improves or hurts readability. On the plus side, the precedence is unambiguous. On the minus side, it arguably hurts unfamiliar readers by implying they're necessary, and makes readability slightly worse by adding yet another pair of parentheses.

PR #195, opened 2025-10-02
ci: run interface tests

james-garner-canonical:
This PR adds the ability to run interface tests to our CI.

## Context

The `charmlibs` monorepo is the new home for interface definitions and tests. We need to be able to execute existing interface tests in our CI.

Interface test execution in `charm-relation-interfaces` is opaque and very abstract. I've attempted to improve this in a couple of PRs (see below), which have informed my approach here:
- https://github.com/canonical/charm-relation-interfaces/pull/292)
- https://github.com/canonical/charm-relation-interfaces/pull/297

The approach is:
1. Separate test setup, execution, and teardown from charm and test discovery.
2. Run interface tests with `pytest` as directly as possible.
3. Execute each `pytest` invocation in a separate job (one per `interface / version / role / charm / endpoint` combination). 

## Overview

This PR adds a new `interface-tests` workflow.

- It runs interface tests for a single interface, totally compatible with existing interface tests and charm specific interface test configuration.
- It's called from the `ci` workflow following the same approach as we use for the `tests` workflow.
- It runs a separate matrix job for each `(version / role (provide or require) / charm / endpoint)` combination known for the interface.

Two scripts have been added to support this workflow.
- `.scripts/get-interface-test-targets.py` calculates the `(version / role / charm / endpoint)` targets for the matrix
- `.scripts/run-interface-tests.py` runs the interface tests for a single target

These scripts replace [run_matrix.py](https://github.com/canonical/charm-relation-interfaces/blob/main/run_matrix.py) from `charm-relation-interfaces` and the interface test and endpoint discovery and execution parts of [pytest-interface-tester](https://github.com/canonical/pytest-interface-tester/tree/main/interface_tester).

## Implementation details

Interface tests are discovered in the local `charmlibs` repository, written to a cloned copy of the charm repo and then executed with `pytest`, using an `interface_tester` based fixture used to configure, setup, and teardown the interface tests (which are currently written using a pseudo-singleton [Tester](https://github.com/canonical/pytest-interface-tester/blob/main/interface_tester/interface_test.py#L175) class). In combination with execution in a separate Github job, this provides much better output in CI.

Currently, the `interface_tester` library only provides the `InterfaceTester.run` entrypoint, which includes test discovery and execution. Instead of calling this, we write a fixture to the test file that calls the (new) `InterfaceTester.context` entrypoint, which handles configuration, setup and teardown of the `interface_tester.Tester` context. Currently, we install `pytest-interface-tester` from my fork, but when [this PR](https://github.com/canonical/pytest-interface-tester/pull/32) is merged, we can switch to `pytest-interface-tester~=3.4`.

## Test failures

The tests for the `tracing` interface are currently failing. They're also [failing in charm-relation-interfaces](https://github.com/canonical/charm-relation-interfaces/actions/runs/18176547258/job/51743670143) with exactly the same (`scenario`) error -- so it's not related to this PR. This highlights the need to get interface tests running here alongside the interface libraries, and in charm repos.

## Future work

A workflow to run interface tests from charm repos will be a separate PR. This should perhaps be in its own repository so that it can be versioned easily.


.github/get-changed.py
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env -S uv run --script --no-project
```
james-garner-canonical, line 1: Changes to this file are a little noisy -- the only substantial change is a new mandatory argument, 'packages' or 'interfaces', which is passed on to `ls.py`.

.github/workflows/ci.yaml
```diff
@@ -1,7 +1,7 @@
 name: CI
```
james-garner-canonical, line 1: `ci.yaml` has been modified to call the `interface-tests.yaml` workflow following the same pattern used to call `tests.yaml`.

.scripts/run-interface-tests.py
```diff
@@ -0,0 +1,235 @@
...
+import os
+import pathlib
+import subprocess
+import sys
+import tempfile
+
+import yaml
+
+# TODO: release a new version of pytest-interface-tester with the features we need
+_INTERFACE_TESTER_DEPENDENCY = 'git+https://github.com/james-garner-canonical/pytest-interface-tester@25-09+feat+location-customization-for-charmlibs'
```
james-garner-canonical, line 37: Right now, all we need is the `context` method, which packages up the internals that we need in a public method. This should land in `pytest-interface-tester` version `3.4`.

.github/workflows/ci.yaml
```diff
@@ -46,7 +47,30 @@ jobs:
...
+    if: ${{ needs.init.outputs.interfaces != '[]' }}
+    strategy:
+      fail-fast: false
+      matrix:
+        interface: ${{ fromJson(needs.init.outputs.interfaces) }}
+    uses: ./.github/workflows/interface-tests.yaml
+    with:
+      interface: ${{ matrix.interface }}
+
+  interfaces-passed:
```
benhoyt, line 67: Maybe add a comment here saying why this job is needed.

.scripts/get-interface-test-targets.py
```diff
@@ -0,0 +1,145 @@
...
+import re
+import subprocess
+import tempfile
+
+import yaml
+
+_REPO_ROOT = pathlib.Path(__file__).parent.parent
+
+logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(str(pathlib.Path(__file__).relative_to(_REPO_ROOT)))
```
benhoyt, line 42: Is this just `.scripts`? Seems much clearer just to say `getLogger('.scripts')`.

.scripts/get-interface-test-targets.py
```diff
@@ -0,0 +1,145 @@
...
+import pathlib
+import re
+import subprocess
+import tempfile
+
+import yaml
+
+_REPO_ROOT = pathlib.Path(__file__).parent.parent
+
+logging.basicConfig(level=logging.DEBUG)
```
benhoyt, line 41: I'd probably move this line inside `_main`.

.scripts/get-interface-test-targets.py
```diff
@@ -0,0 +1,145 @@
...
+                }
+                for charm in charms
+                for endpoint in _get_endpoints(
+                    interface=interface.name,
+                    role=f'{role}s',
+                    charm_repo=charm['url'],
+                    charm_ref=charm.get('branch', 'main'),
+                    charm_root=charm.get('test_setup', {}).get('charm_root', ''),
+                )
+            )
```
benhoyt, line 103: Let's avoid the nested comprehension (see our [style guide](https://github.com/canonical/operator/blob/main/STYLE.md#avoid-nested-comprehensions-and-generator-expressions)). I also prefer avoiding line breaks in the `in` expression of a for loop (like how `_get_endpoints` is broken), for example:

```python
for charm in charms:
    endpoint = _get_endpoints(
        ...
    )
    for endpoint in endpoints:
        target = {
            ...
        }
        targets.append(target)
```

.scripts/get-interface-test-targets.py
```diff
@@ -0,0 +1,145 @@
...
+                    'interface': interface.name,
+                    'version': v.name,
+                    'role': role,
+                    'charm_name': charm['name'],
+                    'endpoint': endpoint,
+                }
+                for charm in charms
+                for endpoint in _get_endpoints(
+                    interface=interface.name,
+                    role=f'{role}s',
```
benhoyt, line 98: It was confusing to me that the `role` var here is "provide" but in `_get_endpoints` it's "provides". Perhaps rename the `_get_endpoints` one to `role_key`?

.scripts/get-interface-test-targets.py
```diff
@@ -0,0 +1,145 @@
...
+import json
+import logging
+import pathlib
+import re
+import subprocess
+import tempfile
+
+import yaml
+
+_REPO_ROOT = pathlib.Path(__file__).parent.parent
```
benhoyt, line 39: Given these are scripts not intended to be imported, could remove the `_` prefix on these to reduce noise slightly? Similar in `run-interface-tests.py`.

.scripts/run-interface-tests.py
```diff
@@ -0,0 +1,268 @@
...
+        - Writes test files and additional modules to cloned charm repo.
+        - Executes tests with pytest.
+    """
+    # write and execute interface tester file in cloned charm repo
+    with _clone_charm_repo(target.charm_config, keep_tempdir=keep_tempdir) as charm_root:
+        charm_test_file = _write_charm_test_file(target, charm_root=charm_root)
+        # create schema and plugin modules at charm root (included in PYTHONPATH)
+        (charm_root / f'{_SCHEMA_MODULE}.py').symlink_to(target.interface_dir / 'schema.py')
+        (charm_root / f'{_PLUGIN_MODULE}.py').write_text(_PLUGIN_CONTENT)
+        # execute interface tests
```
benhoyt, line 192: Can we change this comment to what you explained yesterday, that the intention here is to generate `requirements.txt` (or similar)? The actual execution of interface tests is below, so this comment could be moved down.

benhoyt (approved):
Looks good, thanks. A few very minor comments, but approving optimistically.

.scripts/get-interface-test-targets.py
james-garner-canonical: This is a reasonable idea, but if we were to do so I'd like to update all the scripts at once so they're following the same convention. That said, I kind of prefer keeping them private to make things cleaner if we were to end up sharing common CI code via imports in future.

PR #198, opened 2025-10-06
ci: fix lint recipe variadic argument passing

james-garner-canonical:
This PR fixes a bug in the interaction between the `lint` and `static` recipes, and improves the output of the `lint` recipe.

The `lint` and `static` recipes accept variadic arguments to pass to `pyright`. Currently, if we run `just lint $package` (with no variadiac args), the `static` is called as `just static $package ''`, resulting in a call like `pyright ''` -- that is, a single empty string as the only positional argument. This results in `pyright` type checking the current directory (desired) but ignoring the local config.

This PR fixes this issue like we've done with other cases where variadic arguments need to be passed to different recipes, by switching from using `just` dependencies to passing the arguments positionally to the script.

Additionally, I've taken the opportunity to improve the output of `lint`, so that a linting failure in `ruff` or `codespell` doesn't result in `pyright` being skipped.

### DIFF
--- a/justfile
+++ b/justfile
@@ -25,7 +25,7 @@ fast-lint:
     #!/usr/bin/env -S bash -xueo pipefail
     FAILURES=0
     uv run --only-group=fast-lint ruff check --preview || ((FAILURES+=1))
-    uv run --only-group=fast-lint ruff check --preview --diff || ((FAILURES+=1))
+    uv run --only-group=fast-lint ruff check --preview --diff || : 'Printed diff of changes to fix `ruff check` issues.'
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (approved):

tonyandrewmeyer:
> Additionally, I've taken the opportunity to improve the output of `lint`, so that a linting failure in `ruff` or `codespell` doesn't result in `pyright` being skipped.

I prefer this behaviour as well - I don't think we're particularly consistent on it (across packages), so we should probably look at that at some point (but not in this PR).

PR #205, opened 2025-10-07
ci: use up-to-date ref for git diff to detect changes vs PR target

james-garner-canonical:


.github/workflows/ci.yaml
```diff
@@ -24,19 +24,25 @@ jobs:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
+      - name: Calculate reference for git diff
+        id: ref
+        if: github.event_name == 'pull_request'
+        run: |
+          set -xueo pipefail
+          echo "result=$(git merge-base origin/${{ github.event.pull_request.base.ref }} HEAD)" >> "$GITHUB_OUTPUT"
```
james-garner-canonical: I've just merged `main`, and the idea of moving this logic to `get-changed.py` is no longer immediately viable, as there's now another location where we need to use this ref -- a direct call to `ls.py`.

We don't call `get-changed.py` in that case because we don't want the functionality it layers on top of `ls.py`, namely detecting changes to 'global' files (we don't care about them, we only care about which packages would be published).

If you're not happy with it in the current state, would you prefer this? It's a bit more verbose but perhaps makes the non `pull_request` event behaviour more explicit.
```suggestion
      - name: Calculate reference for git diff
        id: ref
        run: |
          set -xeo pipefail
          if [ ${{ github.event_name }} = 'pull_request' ]; then
            REF=$(git merge-base origin/${{ github.event.pull_request.base.ref }} HEAD)
          else
            REF=
          fi
          echo "result=$REF" >> "$GITHUB_OUTPUT"
```

james-garner-canonical: @dimaqq tagging you in case you don't see this message since it's not a direct reply.

.github/workflows/ci.yaml
```diff
@@ -23,13 +23,19 @@ jobs:
...
```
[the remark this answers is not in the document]
james-garner-canonical: Please see my suggestions in the other thread for some alternative way to spell this that might be easier on the eyes (not sure).

Having slept on this again, I'm thinking I'll factor out `get-changed.py` entirely in its current form as an `ls.py` wrapper in a separate PR, so in the workflow is really the best place for this right now -- I think it's too small to be good as a standalone Python script -- but I've suggested a Python shell step in case you prefer that.

.github/workflows/ci.yaml
james-garner-canonical: No it's ok, I know you didn't suggest these approaches, I just wondered if you'd prefer them. Thanks for the review.

PR #208, opened 2025-10-08
ci: nicer job names for interface test jobs

james-garner-canonical:
This PR , dropping the leading `interfaces/` from the output.

This requ


.scripts/get-interface-test-targets.py
```diff
@@ -94,7 +102,7 @@ def _target_from_interface(interface_str: str, has_tests_only: bool) -> list[dic
                 )
                 targets.extend(
                     {
-                        'interface': interface.name,
+                        **({'interface': interface.name} if include_interface else {}),
```
benhoyt, line 105: Ha, this is a good reason to prefer the append. I find the above quite clever/confusing, though maybe it's me just getting used to this idiom (is it an idiom?).

```python
for endpoint in endpoints:
    target = {
        'version': v.name,
        ...
    }
    if include_interface:
        target['interface'] = interface.name
    targets.append(target)
```

benhoyt (approved):
Seems reasonable to me. I'm a fan of clean/simple test names, and the additional complexity is not much.

PR #210, opened 2025-10-09
ci: push a tag when releasing a package to PyPI

james-garner-canonical:
This PR adds a step to the `Publish` job to push a tag to `main` after successfully publishing a library. The tag looks like `pathops-v1.2.3.post4`.

I've manually added tags in this format for the releases that went untagged after we switched from Github releases to publishing on merge.

## Testing

I tested this job by pushing these changes to `main` in my fork and commenting out the actual publishing step.
- [workflow run](https://github.com/james-garner-canonical/charmlibs/actions/runs/18370175043/job/52332088429)
- [tags](https://github.com/james-garner-canonical/charmlibs/tags)

## Implementation

The changes are a little more extensive than would be ideal, due to the need for version and distribution package name to be made available to the `.github/unify-publish-inputs.py` script. This is done by modifying the `.scripts/ls.py` script in a way that remains compatible with other users of the script, adding a `--output` flag to return a list of JSON objects instead of simple names or paths.

Figuring out the version requires building and installing each package, so to keep the script snappy we bubble up the version and distribution package name from where they are calculated when detecting with packages have had version bumps (and thus require publishing). When preparing to output the requested data, we then calculate the package version and name only if this wasn't already done.

This ends up being a bit verbose. It might be worth considering:
- just eating the performance cost and always (re)-calculating the package version and name (or only if they're requested)
- or using a global cache instead of bubbling up the version and name

### DIFF
--- a/.scripts/ls.py
+++ b/.scripts/ls.py
@@ -107,7 +107,7 @@ def _ls(
         category: What to iterate over -- 'packages' or 'interfaces'.
         old_ref: git reference to diff with.
             If `None`, no diff is performed.
-            Otherwiseonly changed items are returned.
+            Otherwise only changed items are returned.
         new_ref: if old_ref is not `None`, diff it with new_ref.
             If this new_ref is `None`, old_ref is diffed with the current state on disk.
         only_if_version_changed: Only output items that have had a version bump.


james-garner-canonical:
I've refactored the changes to the `ls.py` script with readability in mind.

Along the way, I made a correctness fix: we should calculate the initial list of packages in the repo snapshot of the new git reference.

It also turned out that making the script more complete helped readability -- instead of having various exclusions and errors for interface versions, it's cleaner to just support them by using the directory name (`v0`, etc).

The changes are still more extensive and the diff more verbose than I feel this feature really needs, but hopefully the script is now more readable, maintainable, and extensible.

I've validated the changes by running the script locally, and by pushing some changes to my fork's `main`:
- [unify inputs job](https://github.com/james-garner-canonical/charmlibs/actions/runs/18512598567/job/52756576061) showing successful run and output of `ls.py` and the surrounding wiring
- [publish job](https://github.com/james-garner-canonical/charmlibs/actions/runs/18512598567/job/52756649516) showing the tag push (with actual publishing skipped)
Additionally, the `ls.py` script is invoked in CI for this PR.

PR #220, opened 2025-10-14
docs: add changelog entries for previous releases

james-garner-canonical:
This PR adds changelog entries for the `apt`, `snap`, and `pathops` libraries, including past releases that did not get changelog entries.

It also updates their project URLs, and bumps their versions for a post-release to update this information on PyPI.


james-garner-canonical:
Will merge after #210 to avoid having to create the tags manually.

PR #221, opened 2025-10-15
test(interfaces-tls-certificates): fix timing based test failure

james-garner-canonical:
Well, at least give us another order of magnitude of wiggle room. Otherwise we may see failures like [this](https://github.com/canonical/charmlibs/actions/runs/18515483058/job/52764878185?pr=210):
```
AssertionError: assert datetime.timedelta(seconds=1, microseconds=31) <= datetime.timedelta(seconds=1)
```
Going from 1 -> 10 seconds might be overkill though, maybe 2 seconds is fine.

### DIFF
--- a/interfaces/tls-certificates/tests/unit/test_tls_certificates.py
+++ b/interfaces/tls-certificates/tests/unit/test_tls_certificates.py
@@ -266,7 +266,7 @@ def test_given_ca_certificate_attributes_when_generate_ca_then_ca_is_generated_c
     assert ca_certificate.common_name == "certifier.example.com"
     expected_expiry = datetime.now(timezone.utc) + timedelta(days=365)
     assert ca_certificate.expiry_time
-    assert abs(ca_certificate.expiry_time - expected_expiry) <= timedelta(seconds=1)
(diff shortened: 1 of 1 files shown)


interfaces/tls-certificates/tests/unit/test_tls_certificates.py
```diff
@@ -266,7 +266,7 @@ def test_given_ca_certificate_attributes_when_generate_ca_then_ca_is_generated_c
     assert ca_certificate.common_name == "certifier.example.com"
     expected_expiry = datetime.now(timezone.utc) + timedelta(days=365)
     assert ca_certificate.expiry_time
-    assert abs(ca_certificate.expiry_time - expected_expiry) <= timedelta(seconds=1)
+    assert abs(ca_certificate.expiry_time - expected_expiry) <= timedelta(seconds=10)
```
dimaqq: **minor**: personally I'd prefer an idiom like:

```py
expected - e < actual < expected + e
```

Or, if we know which comes first, something like:
```py
expected <= actual < expected + e
```

Pedants would call this `EPSILON`, but I'm fine with a local `e`.

PR #222, opened 2025-10-15
docs: make template prompt accurate for interface libraries

james-garner-canonical:
When running `just init` or `just interface init`, the user is prompted for their project name. The prompt previously said it should be the package name, but this is only the case for `just init` -- with `just interface init` it's important that the user instead specifies the canonical interface name.

This PR updates the prompt displayed to account for both cases, using a short generic prompt inside `cookiecutter`, but printing extra details beforehand via the `just` recipe.

**OLD**
image
Screenshot from 2025-10-16 12-51-51

**NEW**
Screenshot from 2025-10-16 12-50-29
Screenshot from 2025-10-16 12-50-58

### DIFF
--- a/.template/cookiecutter.json
+++ b/.template/cookiecutter.json
@@ -9,8 +9,8 @@
   "__import_pkg": "{{ cookiecutter.__ns }}.{{ cookiecutter.__pkg }}",
   "__dist_pkg": "{{ cookiecutter.__import_pkg.replace('.', '-').replace('_', '-') }}",
   "__prompts__": {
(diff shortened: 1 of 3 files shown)


dwilding:
I get you, thanks. So it sounds like the most important aspect of this change is the part that says "... as used in charmcraft.yaml"

dwilding (approved):

james-garner-canonical:
> So it sounds like the most important aspect of this change is the part that says "... as used in charmcraft.yaml"

Yes indeed, though I think the other clarifications in those messages are important too.

PR #226, opened 2025-10-16
fix: Log correct renewal_relative_time in error message

saltiyazan:


james-garner-canonical:
Glad that merging this went smoothly! Just a quick note that if you want the change to be released, you need to bump the library version -- but of course there's no obligation to release immediately.

PR #233, opened 2025-10-20
docs: normalize package docs urls, add labels, drop -index suffix

james-garner-canonical:
This poorly scoped PR adds labels to package and interface reference docs, drops the `-index` suffix from interface docs pages, and normalizes the URLs of package docs. A future PR will investigate adding redirects from the underscored variant of the package docs URLs, and perhaps adding redirects for the alternative (incorrect) spellings of interface names.

### DIFF
--- a/.github/workflows/conventional-pr-title.yaml
+++ b/.github/workflows/conventional-pr-title.yaml
@@ -1,0 +1,18 @@
+name: 'PR title follows conventional commits'
+
+on:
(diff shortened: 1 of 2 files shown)


dwilding:
If you add HACKING.md (or similar) in the future to document the documentation setup, I think it would be good to mention there. For now, if there's no obvious place, no worries

PR #234, opened 2025-10-20
ci: add workflow to complain if PR titles don't follow conventional commits

james-garner-canonical:
This PR adds a workflow that validates PR titles, using the same action as [we use in ops](https://github.com/canonical/operator/blob/main/.github/workflows/validate-pr-title.yaml).

I've used the same commit hash for the action as `ops` uses, but otherwise followed [the actions recommended configuration](https://github.com/marketplace/actions/semantic-pull-request#installation) which differs slightly from `ops` (default commit types, read permission at step scope instead of workflow scope, slightly different `pull_request_target` triggers).

Because the workflow runs on `pull_request_target`, it won't run on this PR -- it will only run on PRs after it's merged to `main`.

---

The interface test failures are unrelated.

### DIFF
--- a/.github/workflows/conventional-pr-title.yaml
+++ b/.github/workflows/conventional-pr-title.yaml
@@ -1,0 +1,18 @@
+name: 'PR title follows conventional commits'
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - reopened
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    permissions:
+      pull-requests: read
+    steps:
+      - uses: amannn/action-semantic-pull-request@48f256284bd46cdaab1048c3721360e808335d50  # v6.1.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/conventional-pr-title.yaml
```diff
@@ -0,0 +1,18 @@
+name: 'PR title follows conventional commits'
```
tonyandrewmeyer: This is quite a long name, particularly with the limited space GitHub provides for it. It's definitely informative, and just seeing "PR title ..." is probably enough. But maybe worth shortening?

.github/workflows/conventional-pr-title.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+on:
+  pull_request_target:
```
tonyandrewmeyer: I hope there's a Zizmor check at some point in this repo, so perhaps worth adding this in preparation?

```suggestion
  pull_request_target:  # zizmor: ignore[dangerous-triggers] Doesn't touch code - pull_request results in cancellation errors.
```

tonyandrewmeyer (approved):
A couple of small suggestions, but feel free to ignore. Otherwise looks good to me!

One thing I'm not sure about: the action's docs don't seem to say what the default values are for the config. If you don't provide any explicit `scopes` does that mean that you can use anything at all? If so, then this seems fine, but otherwise that might need to be configured.

.github/workflows/conventional-pr-title.yaml
tonyandrewmeyer: Sure, either's good with me.

PR #235, opened 2025-10-20
feat: add user facing interface tests workflow and docs

james-garner-canonical:
This PR adds the `charm-interface-tests.yaml` workflow from `main` to a separate `interface-tests-v0` branch, changing the name to `interface-tests.yaml` so that users can enjoy a shorter name.

The `interface-tests-v0` branch splits off from the initial commit to this repo, so it doesn't have any of the other packages, interfaces, docs, or CI. It just has:
- `.docs/.readthedocs.yaml`
    - Updated to just have minimal contents to pass the docs build, so we don't get spurious red Xs in CI.
    - It doesn't seem to be possible to disable PR builds for this branch only.
- `.github/workflows/interface-tests.yaml`
    - Copied from `charm-interface-tests.yaml` on `main`.
    - Updated to checkout `canonical/charmlibs@main` instead of my original PR fork (forgot to do this before merging #209).
- `.github/workflows/ci.yaml`
    - Different from `main`, just runs the user facing `interface-tests.yaml` workflow with a hardcoded list of charms.
    - I have some ideas for populating the charms list dynamically, but we can keep it simple for now.
- `.github/dependabot.yaml`
    - Copied from `main`, to keep dependencies up-to-date.
- `.gitignore` (unmodified from initial commit)
- `LICENSE` (unmodified)
- `README.md`
    - Provides information about the workflow.

This is the approach we decided on over #232

---

The interface test failures are unrelated -- the charms under test had an update to a shared library recently that somehow breaks their fixture setup. Assuming they use pinned dependencies, this would have been caught by their using this workflow in their PR CI.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -25,14 +25,14 @@ jobs:
       charm: tempo-coordinator-k8s
       charm-repo: canonical/tempo-operators
 
-  charm-with-charm-repo-and-charm-ref:
+  charm-with-charm-repo-and-charm-branch:
     uses: ./.github/workflows/interface-tests.yaml
     with:
       charm: tempo-coordinator-k8s
       charm-repo: canonical/tempo-operators
       charm-branch: main
 
-  charm-with-charm-ref-only:
+  charm-with-charm-branch-only:
     uses: ./.github/workflows/interface-tests.yaml
     with:
       charm: tempo-coordinator-k8s


README.md
```diff
@@ -0,0 +1,53 @@
...
+```yaml
+# .github/workflows/interface-tests.yaml
+
+on:
+  push:
+    - 'main'
+
+jobs:
+  interfaces:
+    uses: canonical/charmlibs/.github/workflows/charm-interface-tests.yaml@v0
```
benhoyt: Should this be (and similar below):
```suggestion
    uses: canonical/charmlibs/.github/workflows/interface-tests.yaml@interface-tests-v0
```

README.md
```diff
@@ -0,0 +1,53 @@
+# charm-interface-tests.yaml@v0
```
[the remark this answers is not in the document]
benhoyt: Yep, fine with me, thx

README.md
```diff
@@ -0,0 +1,53 @@
...
+```yaml
+on:
+  pull_request:
+
+jobs:
+  interfaces:
+    uses: canonical/charmlibs/.github/workflows/interface-tests.yaml@interface-tests-v0
+    with:
+      charm: <your-charm-name>
+      charm_repo: ${{ format('{0}/{1}.git', github.server_url, github.event.pull_request.head.repo.full_name) }}
```
james-garner-canonical: Sorry for a late follow-up / design question, but I wonder if it's worth moving this complexity into the workflow so that for users, it just looks like this:
```suggestion
      charm_repo: ${{ github.event.pull_request.head.repo.full_name }}
```
It would mean that the worfklow would need to do something like
```diff
-            --charm-repo '${{ inputs.charm_repo }}' \
+            --charm-repo '${{ inputs.charm_repo && format('{0}/{1}.git', github.server_url, inputs.charm_repo) || '' }}' \
```
Which I think you can see why I didn't write it that way originally!

An alternative to the `&&` `||` expression would be to add a preceding step to `init` that spells out the logic more explicitly.

Worth doing one of the above to make the UX nicer?

benhoyt: Yeah, I think you're right -- that's nicer for the user, and only a little bit less nice for us.

james-garner-canonical: Done

PR #236, opened 2025-10-21
ci: add zizmor

james-garner-canonical:
This PR adds `zizmor` to this repo, to provide security code scanning for our workflows. It will be enabled as a required security check for PRs after merging this PR.

The PR fixes the easy issues too -- narrowing permissions, avoiding cache poisoning, pinning some workflows, and manually ignoring a dangerous workflow trigger warning since it should be safe in practice.

I haven't fixed the `charm-interface-tests.yaml` workflow errors, as that whole file will shortly be removed in #238.

The big outstanding thing is the extensive use of template expansion. That's too big to fix in this PR. I'm not sure if we can merge as-is, or if we should defer enabling the security checks as required till this is addressed in a separate PR, or if we need to carve out an ignore in `zizmor.yml` for now.

[**Code scanning results**](https://github.com/canonical/charmlibs/security/code-scanning?query=pr%3A236+tool%3Azizmor+is%3Aopen)

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -65,6 +65,8 @@ jobs:
           persist-credentials: false
       - name: Install uv
         uses: astral-sh/setup-uv@v6
+        with:
+          enable-cache: false  # avoid cache poisoning when publishing
       - name: Build
         run: uv build
         working-directory: ./${{ matrix.package }}


tonyandrewmeyer (approved):
Thanks for doing this!

> The big outstanding thing is the extensive use of template expansion. That's too big to fix in this PR. I'm not sure if we can merge as-is, or if we should defer enabling the security checks as required till this is addressed in a separate PR, or if we need to carve out an ignore in zizmor.yml for now.

I think it's better to merge as-is, so that we have these improvements and more visibility on what else needs to be done.

I think an issue to handle the remaining ones would probably be best - that's more visible in terms of work to be done (perhaps a 26.04 label?) than only having the scanning alerts (but it can just link to those).

You might be right that the "fix" might be disabling the checks for some files and making sure we are as protected as possible (relying on human review, for example) because the system is somewhat *designed* to be templated with arbitrary inputs.

PR #239, opened 2025-10-21
fix: use locked dependencies to run interface tests

james-garner-canonical:
This PR fixes the failing `tracing` interface tests by updating the `tempo-coordinator-k8s` and `pyroscope-coordinator-k8s` interface test configuration to respect the charms' locked dependencies, by using `uv export` instead of `uv pip compile` in their `pre_run`.

Thanks @PietroPasotti for identifying the fix!


tonyandrewmeyer (approved):
LGTM!

PR #259, opened 2025-11-19
feat: add systemd library to charmlibs

NucciTheBoss:
Hello hello @james-garner-canonical :wave: 

I've opened this PR to migrate the `systemd` charm library from `operator-libs-linux` to here as I've started opening PRs to migrate our existing HPC charms off of `operator-libs-linux`. It's pretty much 1:1 copy of the `systemd` library, but I've made some modifications to the private `_systemctl` function such as using `capture_output` to capture both STDOUT and STDERR and condensed the `SystemdError` stack trace for if a `systemctl` command fails. It doesn't make much sense to include the captured `subprocess.CalledProcessError` exception in the same stacktrace since they present the same information. 

Some other things of note are:

1. I migrated the systemd library's unit tests from `unittest` to `pytest`.
2. I disabled static type checking on the `tests/` directory. `pyright` wasn't a fan of my `make_mock` fixture, and it kept "losing context" of what the return types for the tuple values were. Since test type annotations aren't the gravy the static typing enthusiasts are after, I decided it was best to disable static type check on `tests/` rather than have it be a blocker in CI.
3. I made the `charmlibs-maintainer` group the CODEOWNER for the `systemd` module. I could add the HPC team as the code owner since we use the library rather heavily, but systemd is a general enough library that I thought it would be better to have an interdisciplinary group responsible for reviewing changes.

I followed the tutorial on ReadTheDocs, so please let me know if I missed anything!


james-garner-canonical:
Hi @NucciTheBoss, this looks great, thanks for doing this. @dwilding is planning to migrate the remaining `operator-libs-linux` libs later this cycle, but we have no objections to you taking care of this one for us.

The only slightly tricky thing is that I need to make sure I review the changes from the `operator-libs-linux` version.

The commit history for this PR looks pretty minimal, so I'm wondering: would it be much trouble for you to add a first commit where the python files are all identical to the `operator-libs-linux` version -- just for ease of review? (Obviously everything will be broken in that specific commit, but that's fine.)

NucciTheBoss:
Sure thing! I can update the PR in a bit here. I'll just need to rework the main commit as I included the new charm library and everything else generated by `just init` in a single commit.

NucciTheBoss:
@james-garner-canonical I've updated the git history to include the original copy of the systemd library from `operator-libs-linux`. Let me know if there's anything else you need to help make the review easier!

systemd/pyproject.toml
```diff
@@ -0,0 +1,69 @@
+[project]
+name = "charmlibs-systemd"
+description = "The charmlibs.systemd package."
+readme = "README.md"
+requires-python = ">=3.10"
+authors = [
+    {name="The HPC team at Canonical"},
```
james-garner-canonical: I guess this should be the Charm Tech team since we're still owning it -- but like I said, we're certainly not opposed to having another team (like yours) take over maintenance of a lib like this.

systemd/pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+    # "E501",  # line too long
+]
+
+[tool.pyright]
+extends = "../pyproject.toml"
+include = ["src"]
+pythonVersion = "3.10"  # check no python > 3.10 features are used
+
+[tool.charmlibs.functional]
+ubuntu = []  # ubuntu versions to run functional tests with, e.g. "24.04" (defaults to just "latest")
```
james-garner-canonical: Do see any benefit to testing on any additional bases here?

systemd/src/charmlibs/systemd/_systemd.py
```diff
@@ -12,98 +12,47 @@
...
-            text=True,
-            bufsize=1,
-            encoding="utf-8",
-            check=check,
-        )
-        logger.debug(
-            f"Command {cmd} exit code: {proc.returncode}. systemctl output:\n{proc.stdout}"
-        )
-        return proc.returncode
+        result = subprocess.run(cmd, capture_output=True, text=True, check=check)
```
james-garner-canonical: I'm fine with moving the logging and return out of the try block, but let's minimise the diff otherwise.

systemd/src/charmlibs/systemd/_systemd.py
```diff
@@ -12,98 +12,47 @@
...
-
-    Returns:
-        Returncode of systemctl command execution.
+        args: Arguments to pass to the `systemctl` command.
+        check: If set to `True`, raise an error if the command exits with a non-zero exit code.
 
     Raises:
-        SystemdError: Raised if calling systemctl returns a non-zero returncode and check is True.
+        subprocess.CalledProcessError:
+            Raised if the called command fails and check is set to `True`.
```
james-garner-canonical: We still raise `SystemdError`, so the original docstring is more correct here.

systemd/src/charmlibs/systemd/_systemd.py
```diff
@@ -12,98 +12,47 @@
...
-            f"Command {cmd} exit code: {proc.returncode}. systemctl output:\n{proc.stdout}"
-        )
-        return proc.returncode
+        result = subprocess.run(cmd, capture_output=True, text=True, check=check)
     except subprocess.CalledProcessError as e:
         raise SystemdError(
-            f"Command {cmd} failed with returncode {e.returncode}. systemctl output:\n{e.stdout}"
-        )
+            f"systemctl command '{' '.join(e.cmd)}' failed with exit code {e.returncode}. "
+            + f'reason: {e.stderr}'
```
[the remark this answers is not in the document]
james-garner-canonical: I also think `Command ... failed` is better since all the commands will be prefixed with `systemd`.

systemd/tests/functional/test_systemd.py
```diff
@@ -0,0 +1,111 @@
+# Copyright 2025 Canonical Ltd.
```
james-garner-canonical: These are migrated as-is, with only the namespacing changed, nice.

systemd/pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+    # "typing_extensions",
+]
+unit = [  # installed for `just unit systemd`
+    "ops[testing]",
+]
+functional = [  # installed for `just functional systemd`
+]
+integration = []
+
+[project.urls]
```
james-garner-canonical: I guess I need to update the template to add these:
```suggestion
[project.urls]
"Documentation" = "https://documentation.ubuntu.com/charmlibs/reference/charmlibs/systemd/"
```

systemd/pyproject.toml
```diff
@@ -0,0 +1,69 @@
...
+"Repository" = "https://github.com/canonical/charmlibs"
+"Issues" = "https://github.com/canonical/charmlibs/issues"
```
james-garner-canonical:
```suggestion
"Issues" = "https://github.com/canonical/charmlibs/issues"
"Changelog" = "https://github.com/canonical/charmlibs/blob/main/systemd/CHANGELOG.md"
```

james-garner-canonical (changes requested):
Thanks again for this PR and for structuring the diff for easier review. I've gone through all the changes, and looked at the test suites side by side since their changes weren't covered by the diff. I've made a few suggestions that I think should be addressed, but happy to chat about any of them.

systemd/src/charmlibs/systemd/_systemd.py
```diff
@@ -12,98 +12,47 @@
...
+        ) from None
```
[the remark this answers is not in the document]
james-garner-canonical: After tinkering with this in a REPL, I'm pretty convinced that the `CalledProcessError` traceback doesn't add anything but noise, so your `from None` should be a good improvement to UX.

systemd/src/charmlibs/systemd/_systemd.py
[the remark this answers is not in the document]
NucciTheBoss: I updated the formatting for the error message in the reworked commits. I include both `stderr` and `stdout` in the output, and I'm reusing the original `Command ... failed` format.

james-garner-canonical:
Thanks, @NucciTheBoss, the changes are very easy to see now. Thanks for addressing my comments, everything's looking good. I'll merge and release the library tomorrow.

james-garner-canonical (approved):

PR #266, opened 2025-11-25
feat: add Temporal charm integration libraries

wyattrees:
This adds two new interface libraries, `temporal_worker_consumer` and `temporal_host_info` for upcoming changes to the Temporal charm suite to include these new relations.

### DIFF
--- a/interfaces/temporal-worker-consumer/tests/unit/dummy_requirer/src/charm.py
+++ b/interfaces/temporal-worker-consumer/tests/unit/dummy_requirer/src/charm.py
@@ -26,7 +26,8 @@ class DummyWorkerConsumerRequirerCharm(CharmBase):
         super().__init__(*args)
         self.worker_consumer = TemporalWorkerConsumerRequirer(self)
         self.framework.observe(
-            self.worker_consumer.on.worker_consumer_available, self._on_worker_consumer_available
+            self.worker_consumer.on.temporal_worker_consumer_available,
+            self._on_worker_consumer_available,
         )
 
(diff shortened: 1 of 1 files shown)


james-garner-canonical:
> @james-garner-canonical is there a way to skip the machine integration tests? These relations are for k8s charms so I don't think it makes sense to have machine integration tests

Yes, @wyattrees, CI uses `pytest` marks to collect substrate specific tests, so if you mark all your integration tests as `k8s_only` then the machine integration tests will be skipped entirely.

You can mark an entire module of tests by putting `pytestmark = pytest.mark.k8s_only` at the top. CI docs [here](https://documentation.ubuntu.com/charmlibs/how-to/customize-integration-tests/#how-to-customize-integration-tests).

james-garner-canonical:
> > @james-garner-canonical is there a way to skip the machine integration tests? These relations are for k8s charms so I don't think it makes sense to have machine integration tests
> 
> Yes, @wyattrees, CI uses `pytest` marks to collect substrate specific tests, so if you mark all your integration tests as `k8s_only` then the machine integration tests will be skipped entirely.
> 
> You can mark an entire module of tests by putting `pytestmark = pytest.mark.k8s_only` at the top. CI docs [here](https://documentation.ubuntu.com/charmlibs/how-to/customize-integration-tests/#how-to-customize-integration-tests).

Having thought about this more and talked it over with the team, it seems like the library doesn't have any k8s specific requirements or anything, so what do you think about just testing on both substrates? It doesn't look like the test charms would need to differ much, so it shouldn't lead to much duplication.

PR #270, opened 2025-11-26
docs: fix bad links in interface docs

james-garner-canonical:
We don't normally pay too much attention to the linkcheck job, as it's often a bit flaky due to Charmhub. This has resulted in a number of bad links being introduced with the bulk migration of interface docs.

For the most part, these are local links to JSON schemas which we don't host. These have all been replaced with links to the interface's Pydantic schema instead.

Additionally, the `openfga` docs included a deep link to the library's source code in the Github repository, the layout of which has now changed. I've replaced this with a hopefully more stable link to the library on Charmhub.

### DIFF
--- a/interfaces/openfga/interface/v0/README.md
+++ b/interfaces/openfga/interface/v0/README.md
@@ -4,7 +4,7 @@
 
 This relation interface describes the expected behaviour of any charm claiming to be able to interact with a OpenFGA.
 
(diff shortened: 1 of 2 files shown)


interfaces/auth_proxy/interface/v0/README.md
```diff
@@ -35,17 +35,14 @@ Both the requirer and the provider need to adhere to a certain set of criteria t
 
 ## Relation Data
 
-### Provider
+[\[Pydantic Schema\]](./schema.py)
```
dimaqq: Out of curiosity, why is the linked text in square brackets?

dimaqq (approved):

PR #272, opened 2025-11-27
feat: generate .interfaces.json file for Charmhub

james-garner-canonical:
This PR updates `.scripts/ls.py` to output additional interface metadata that we want to expose to Charmhub.

This includes adding support for new fields in `interface.yaml`: `lib`, `summary`, and `description`. These fields have been added to the `tracing` interface as an example and to exercise the new code.

The `just interfaces-json` recipe has been added to write out the `.interfaces.json` file, and a job has been added to CI to check that it is up-to-date on PRs.

---

I haven't added `providers` and `requirers` to `.interfaces.json` (nor added support to `.scripts/ls.py`). It seems like it could increase the file size by quite a bit, and I'm not sure if it makes things easier from Charmhub's side or not -- so that could be a future PR if it would.

### DIFF
--- a/.scripts/ls.py
+++ b/.scripts/ls.py
@@ -80,7 +80,7 @@ def _main() -> None:
     parser.add_argument('--exclude-examples', action='store_true')
     parser.add_argument('--exclude-placeholders', action='store_true')
     parser.add_argument('--only-if-version-changed', action='store_true')
-    parser.add_argument('--json-indent', type=int, default=None)
+    parser.add_argument('--indent-json', action='store_true')
     group = parser.add_mutually_exclusive_group()
     group.add_argument(
         '--output', action='append', choices=[f.name for f in dataclasses.fields(Info)]
@@ -104,7 +104,7 @@ def _main() -> None:
         )
     else:
         result = sorted(getattr(info, single_output) for info in infos)
-    print(json.dumps(result, indent=args.json_indent))
+    print(json.dumps(result, indent=2 if args.indent_json else None))
 
 
 def _ls(
--- a/justfile
+++ b/justfile
@@ -179,5 +179,5 @@ interfaces-json:
         --output docs_url \
         --output summary \
         --output description \
-        --json-indent 2 \
+        --indent-json \
         > .interfaces.json


.scripts/ls.py
```diff
@@ -69,7 +81,7 @@ def _main() -> None:
     parser.add_argument('--exclude-placeholders', action='store_true')
     parser.add_argument('--only-if-version-changed', action='store_true')
     group = parser.add_mutually_exclusive_group()
-    group.add_argument('--output', action='append', choices=['path', 'name', 'version'])
+    group.add_argument('--output', action='append', choices=tuple(dataclasses.asdict(Info(''))))
```
james-garner-canonical: Since there are a lot of fields now, it's simpler to pull them from the `Info` class.

.scripts/ls.py
```diff
@@ -214,9 +230,9 @@ def _changed_only(root: pathlib.Path, dirs: list[pathlib.Path], ref: str) -> lis
     return [p for p in dirs if p in changes]
 
 
-def _get_changed_version_info(
+def _get_changed_versions_only(
     category: str, root: pathlib.Path, dirs: list[pathlib.Path], ref: str
-) -> list[Info]:
+) -> list[pathlib.Path]:
```
james-garner-canonical: This function has been changed to narrow `dirs` instead of returning `Info` objects.

.scripts/ls.py
```diff
@@ -225,24 +241,27 @@ def _get_changed_version_info(
...
-        logger.info('%s (%s): %s -> %s', info.path, info.name, old_version, info.version)
-        if info.version == old_version:
+        name = _get_name(category, root, path)
+        old_version = old_versions.get(name)
+        new_version = _get_version(category, root, path)
+        logger.info('%s (%s): %s -> %s', path, name, old_version, new_version)
+        if new_version == old_version:
             logger.debug('Version unchanged')
-        elif 'dev' in info.version:
+        elif 'dev' in new_version:
```
james-garner-canonical: Again, use our new single field functions instead of pulling fields out of an `Info`.

.scripts/ls.py
```diff
@@ -263,37 +282,104 @@ def _snapshot_repo(ref: str | None):
...
+@functools.cache
+def _get_package_version(package: pathlib.Path, root: pathlib.Path = _REPO_ROOT) -> str:
+    name = _get_dist_name(package, root=root)
+    script = f'import importlib.metadata; print(importlib.metadata.version("{name}"))'
+    cmd = ['uv', 'run', '--no-project', '--with', root / package, 'python', '-c', script]
+    return subprocess.check_output(cmd, cwd=root, text=True).strip()
+
+
+@functools.cache
+def _get_interface_version(path: pathlib.Path, root: pathlib.Path = _REPO_ROOT) -> str:
```
james-garner-canonical: Cached since we do some filesystem globbing.

.scripts/ls.py
```diff
@@ -263,37 +282,104 @@ def _snapshot_repo(ref: str | None):
...
+    versions = [v.name for v in (root / path / 'interface').glob('v[0-9]*')]
+    return max((v.removeprefix('v') for v in versions), key=int)
+
+
+@functools.cache
+def _lib_urls() -> dict[str, str]:
```
james-garner-canonical: Cached since we read two files (converting them into a `dict`).

justfile
```diff
@@ -168,3 +168,16 @@ _integration package substrate label +flags:
...
+    .scripts/ls.py interfaces \
+        --output name \
+        --output version \
+        --output lib \
+        --output lib_url \
+        --output docs_url \
+        --output summary \
+        --output description \
+        --json-indent 2 \
+        > .interfaces.json
```
james-garner-canonical: The order of the fields on the CLI determines their order in the JSON output.

.interfaces.json
```diff
@@ -0,0 +1,596 @@
+[
+  {
+    "name": "auth_proxy",
+    "version": "0",
```
james-garner-canonical: String version for consistency with generalised versions like `0.0-a1` (not that interfaces have such a scheme yet), and so that all fields are just strings.

.interfaces.json
```diff
@@ -0,0 +1,596 @@
...
+    "lib": "",
+    "lib_url": "",
+    "docs_url": "https://documentation.ubuntu.com/charmlibs/reference/interfaces/auth_proxy/",
```
james-garner-canonical: Auto-generated URL.

.interfaces.json
```diff
@@ -0,0 +1,596 @@
...
+    "summary": "",
+    "description": ""
```
james-garner-canonical: More new `interface.yaml` fields.

.github/workflows/ci.yaml
```diff
@@ -96,6 +96,18 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          persist-credentials: false
+      - uses: astral-sh/setup-uv@v6
+      - name: Fail if .interfaces.json is not up-to-date.
+        run: |
+          uvx --from rust-just just interfaces-json
+          git diff --exit-code
```
james-garner-canonical: We had considered adding a check to `fast-lint`, but I'm not sure if it's worth it. We'd need to write the output to a temporary location and then assert the file contents are the same. I figure most PRs won't touch the relevant interface files anyway.

tonyandrewmeyer: It seems reasonable to me to wait and see how common it is for PRs to fail on this step. If we see it happening regularly, we can add a check at that time.

.scripts/ls.py
```diff
@@ -263,37 +282,104 @@ def _snapshot_repo(ref: str | None):
...
+
+
+@functools.cache
+def _pyproject_toml(package: pathlib.Path, root: pathlib.Path = _REPO_ROOT):
+    with (root / package / 'pyproject.toml').open('rb') as f:
+        return tomllib.load(f)
+
+
+@functools.cache
+def _interface_yaml(path: pathlib.Path, root: pathlib.Path = _REPO_ROOT):
```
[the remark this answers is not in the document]
tonyandrewmeyer: I wonder if `lru_cache` would be better here, since there can be an unlimited number of these.

.scripts/ls.py
```diff
@@ -263,37 +285,106 @@ def _snapshot_repo(ref: str | None):
...
+@functools.cache
+def _get_interface_version(path: pathlib.Path, root: pathlib.Path = _REPO_ROOT) -> str:
+    versions = [v.name for v in (root / path / 'interface').glob('v[0-9]*')]
+    return max((v.removeprefix('v') for v in versions), key=int)
+
+
+@functools.cache
+def _lib_urls() -> dict[str, str]:
+    result: dict[str, str] = {}
+    for p in 'general-libs.csv', 'interface-libs.csv':
```
tonyandrewmeyer: I wonder if we should be loading content from `.docs` here. It feels like `.docs` should be self-contained (or write-only, I guess).

If we need to get data from this CSV (which is starting to feel like it ought to be a sqlite3 db 😆) then it feels like it ought to be placed elsewhere and the docs could either have a symlink or access it from elsewhere.

.scripts/ls.py
tonyandrewmeyer: Also here I wonder if `lru_cache` would be a better choice.

.scripts/ls.py
```diff
@@ -263,37 +282,104 @@ def _snapshot_repo(ref: str | None):
...
-def _get_dist_name(package: pathlib.Path) -> str:
+def _get_dist_name(package: pathlib.Path, root: pathlib.Path = _REPO_ROOT) -> str:
     """Load distribution package name from pyproject.toml and normalize it."""
-    with (package / 'pyproject.toml').open('rb') as f:
-        return _normalize(tomllib.load(f)['project']['name'])
+    return _pyproject_toml(package, root=root)['project']['name']
+
+
+@functools.cache
+def _get_package_version(package: pathlib.Path, root: pathlib.Path = _REPO_ROOT) -> str:
```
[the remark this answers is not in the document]
tonyandrewmeyer: Maybe `lru_cache` here too.

tonyandrewmeyer (approved):
Looks good, thanks! A few small comments/suggestions.

With the various caching - it doesn't seem like all of that is used (although I didn't step through it in complete detail). Is it just there because if it's reused caching makes sense? If so, then I suggest `lru_cache` even more strongly, with low counts for the most part.

.github/workflows/ci.yaml
james-garner-canonical: Makes sense, thanks.

.scripts/ls.py
```diff
@@ -68,8 +80,9 @@ def _main() -> None:
     parser.add_argument('--exclude-examples', action='store_true')
     parser.add_argument('--exclude-placeholders', action='store_true')
     parser.add_argument('--only-if-version-changed', action='store_true')
+    parser.add_argument('--json-indent', type=int, default=None)
```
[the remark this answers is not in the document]
james-garner-canonical: The use case is that everywhere else in CI that we use this script, it makes sense for it just to be an un-indented json blob, since it either gets parsed immediately (e.g. in a Python script), or it's set as `GITHUB_OUTPUT`. It's only for making the `.interfaces.json` more maintainable (e.g. saner diffs) that we want to indent in this case.

So maybe it should be a boolean flag instead of a numerical one?

.scripts/ls.py
```diff
@@ -40,19 +45,26 @@
...
+
 _REPO_ROOT = pathlib.Path(__file__).parent.parent
 
 logging.basicConfig(level=logging.DEBUG)
 logger = logging.getLogger(str(pathlib.Path(__file__).relative_to(_REPO_ROOT)))
 
 
-@dataclasses.dataclass(frozen=True)
+@dataclasses.dataclass
 class Info:
```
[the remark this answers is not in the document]
james-garner-canonical: Good call, will do.

james-garner-canonical:
> With the various caching - it doesn't seem like all of that is used (although I didn't step through it in complete detail). Is it just there because if it's reused caching makes sense? If so, then I suggest `lru_cache` even more strongly, with low counts for the most part.

AFAICT it's all used. We cache loading `interface.yaml` and `pyproject.toml`, which will be called multiple times in the `Info` construction loop. That would be covered by a `lru_cache` with a single entry, but they'll also be called when we're narrowing the directories to return (we don't do this in `just interfaces-json`, but we do in other uses of this script).

Version info is only called once in the `Info` construction loop, but it's also used internally for loading `interface.yaml`, and when narrowing the directories to return.

The other thing we cache is loading the libraries CSVs. We actually only need a single call to this function per script execution, but I figured that calling it in the loop with a cache decorator was more readable than other solutions.

Ultimately none of what we load is too memory intensive, so I think the caching is a nice way to keep the high-level bit readable while abstracting away the need to think about whether you need to keep around information from subprocess calls etc.

.scripts/ls.py
tonyandrewmeyer: Yeah, not sure where would be best. Maybe the root does make sense, because they are basically summarising everything contained here (plus more)?

tonyandrewmeyer:
> AFAICT it's all used.

I looked more closely and I can see them being used, so I remove any objection to the caching... but:

> Ultimately none of what we load is too memory intensive, so I think the caching is a nice way to keep the high-level bit readable while abstracting away the need to think about whether you need to keep around information from subprocess calls etc.

A constrained cache does do all of this but without unbounded memory use, though 😄.

PR #289, opened 2025-12-10
feat: release charmlibs.apt 0.8.0 with Python 3.8 support

james-garner-canonical:
This PR adds a release of `charmlibs.apt` with Python 3.8 support. The version is 0.8.0, as we want a sub 1.0 version to convey the lack of long term support, and so that it doesn't show up as the latest release on PyPI. The differences from the 1.0 version are some `typing` changes an the required Python version.

We can't run the functional tests on Ubuntu 20.04 easily, as Github has discontinued that runner, but the unit tests pass as expected when run with Python 3.8.


james-garner-canonical:
The issue needs `charmlibs.snap`, not `charmlibs.apt`, so I'm closing this PR.

PR #290, opened 2025-12-10
ci: pull in changes to docs CI from main so linkcheck won't auto-fail

james-garner-canonical:
This PR pulls in changes to the docs CI from `main` as well, which was neglected in #287. We run `linkcheck` with slightly different arguments on `push` than we do on `pull_request`, so this wasn't caught in the PR. This PR updates the `docs.just` file to match the copy in `main`. The relevant change is the extra variadic argument for the `linkcheck` recipe.

Example failure: https://github.com/canonical/charmlibs/actions/runs/20085246192/job/57621101285

### DIFF
--- a/docs.just
+++ b/docs.just
@@ -21,13 +21,15 @@ to be combined into the final docs in a separate pass.
 """)]
 _packages *packages:
     #!/usr/bin/env -S uv run --script --no-project
-    import pathlib, subprocess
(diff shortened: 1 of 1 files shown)


dwilding (approved):
Ah yes, I remember, we wanted to avoid checking GitHub links that might not exist until a PR is merged.

PR #301, opened 2026-01-08
feat: add http_endpoint interface

chanchiwai-ray:
Add general `http_endpoint` interface to get the leader's http endpoint of a related application, and publish to application data bag. Typical use case is that when a charm simply need to get the ingress URL from another charm. (`juju_info` can do the job, but that's very unclear; `ingress` provided additional opinionated attributes, and it's not native to use without post processing of the returned address)

Example use case: 

https://github.com/canonical/falco-operators/pull/26

### DIFF
--- a/interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
+++ b/interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
@@ -77,6 +77,10 @@ class HttpEndpointProvider(Object):
         This method sets the HTTP endpoint information of the leader unit in the relation
         application data bag.
         """
+        if not self.charm.unit.is_leader():
+            logger.debug('Only leader unit can set http endpoint information')
+            return
+
         relations = self.charm.model.relations[self.relation_name]
         if not relations:
             logger.debug('No %s relations found', self.relation_name)
@@ -95,13 +99,12 @@ class HttpEndpointProvider(Object):
             )
             return
 
+        # Publish the HTTP endpoint to all relations" application data bags
         url = f'{self.scheme}://{ingress_address}:{self.listen_port}/{self.path.lstrip("/")}'
         try:
             http_endpoint = _HttpEndpointDataModel(url=HttpUrl(url))
             for relation in relations:
-                relation.save(http_endpoint, self.charm.unit)
-                if self.charm.unit.is_leader():
-                    relation.save(http_endpoint, self.charm.app)
+                relation.save(http_endpoint, self.charm.app)
                 logger.info(
                     'Published HTTP endpoint to relation %s: %s', relation.id, http_endpoint
                 )
@@ -149,34 +152,6 @@ class HttpEndpointRequirer(Object):
         self.charm = charm
         self.relation_name = relation_name
 
-    def get_unit_urls(self) -> dict[str, str]:
-        """Get the list of urls from HTTP endpoints from all related units.
-
-        This method retrieves the URLs from the HTTP endpoints provided by all related units.
-
-        Returns:
-            A dictionary of unit names to URLs from the HTTP endpoints of all related units if
-            available.
-        """
-        relations = self.charm.model.relations[self.relation_name]
-        if not relations:
-            logger.debug('No %s relations found', self.relation_name)
-            return {}
-
-        http_endpoints: dict[str, str] = {}
-        for relation in relations:
-            for unit in relation.units:
-                if unit not in relation.data and not relation.data.get(unit):
-                    logger.warning('Relation data (%s) is not ready', self.relation_name)
-                    continue
-                try:
-                    data = relation.load(_HttpEndpointDataModel, unit)
-                    http_endpoints[unit.name] = str(data.url)
-                    logger.info('Retrieved URL from relation %s: %s', relation.id, data)
-                except ValidationError as e:
-                    logger.error('Invalid URL endpoint data in relation %s: %s', relation.id, e)
-        return http_endpoints
-
     def get_app_urls(self) -> dict[str, str]:
         """Get the list of urls from HTTP endpoints from all related applications.
 
@@ -184,7 +159,7 @@ class HttpEndpointRequirer(Object):
         related applications.
 
         Returns:
-            A dictionary of unit names to URLs from the HTTP endpoints of all leader units if
+            A dictionary of app names to URLs from the HTTP endpoints of all leader units if
             available.
         """
         relations = self.charm.model.relations[self.relation_name]
--- a/interfaces/http_endpoint/tests/unit/conftest.py
+++ b/interfaces/http_endpoint/tests/unit/conftest.py
@@ -68,14 +68,9 @@ def requirer_charm_relation_1() -> ops.testing.Relation:
     return ops.testing.Relation(
         endpoint='http-endpoint',
         interface='http_endpoint',
-        remote_app_name='remote_0',
+        remote_app_name='remote_1',
         remote_app_data={
             'url': '"http://10.0.0.1:8080/"',
-        },
-        remote_units_data={
-            0: {'url': '"http://10.0.0.1:8080/"'},
-            1: {'url': '"http://10.0.0.2:8080/"'},
-            2: {'url': '"http://10.0.0.3:8080/"'},
         },
     )
 
@@ -86,14 +81,9 @@ def requirer_charm_relation_2() -> ops.testing.Relation:
     return ops.testing.Relation(
         endpoint='http-endpoint',
         interface='http_endpoint',
-        remote_app_name='remote_1',
+        remote_app_name='remote_2',
         remote_app_data={
             'url': '"https://10.0.1.1:8443/"',
-        },
-        remote_units_data={
-            0: {'url': '"https://10.0.1.1:8443/"'},
-            1: {'url': '"https://10.0.1.2:8443/"'},
-            2: {'url': '"https://10.0.1.3:8443/"'},
         },
     )
 
@@ -104,9 +94,6 @@ def provider_charm_relation_1() -> ops.testing.Relation:
     return ops.testing.Relation(
         endpoint='http-endpoint',
         interface='http_endpoint',
-        local_unit_data={
-            'ingress-address': '10.0.0.5',
-        },
     )
 
 
--- a/interfaces/http_endpoint/tests/unit/test_http_endpoint.py
+++ b/interfaces/http_endpoint/tests/unit/test_http_endpoint.py
@@ -149,6 +149,33 @@ class TestHttpEndpointProvider:
                     path=path, scheme=scheme, listen_port=listen_port
                 )
 
+    def test_non_leader_does_not_publish(
+        self,
+        provider_charm_meta: dict[str, Any],
+        provider_charm_relation_1: ops.testing.Relation,
+        provider_charm_relation_2: ops.testing.Relation,
+    ):
+        """Test that non-leader units do not publish endpoint data."""
+        ctx = ops.testing.Context(
+            ProviderCharm,
+            meta=provider_charm_meta,
+        )
+
+        relation_1 = provider_charm_relation_1
+        relation_2 = provider_charm_relation_2
+
+        state_in = ops.testing.State(
+            leader=False,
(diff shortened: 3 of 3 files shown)


interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
```diff
@@ -0,0 +1,237 @@
...
+import logging
+from collections.abc import MutableMapping
+
+from ops import CharmBase, CharmEvents, EventBase, EventSource, Object
+from pydantic import BaseModel, HttpUrl, ValidationError
+
+logger = logging.getLogger(__name__)
+
+
+class DataValidationError(Exception):
```
james-garner-canonical: WDYT about being extra clear and defensive about accidentally exposing this internal-only exception type and giving it an underscore prefix?

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
```diff
@@ -0,0 +1,237 @@
...
+    """Exception raised for invalid data."""
+
+
+class InvalidHttpEndpointDataError(Exception):
+    """Exception raised for invalid http_endpoint data."""
+
+
+class HttpEndpointDataModel(BaseModel):
```
james-garner-canonical: I'm wondering whether it would be better not to expose this class publicly. It seems that the provider doesn't use it, as the ingress address is pulled from the information that Juju provides, while the requirer is only really getting the URL string.

The user-facing change I'm suggesting is:
1. Make this class private.
2. Have `HttpEndpointRequirer` only expose a `get_urls -> list[str]` method instead of exposing the (now private) databag model class.

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
```diff
@@ -0,0 +1,237 @@
...
+            )
+            return
+
+        # Publish the HTTP endpoint to all relations" application data bags
+        url = f'{self.scheme}://{ingress_address}:{self.listen_port}/{self.path.lstrip("/")}'
+        try:
+            http_endpoint = HttpEndpointDataModel(url=HttpUrl(url))
+            for relation in relations:
+                relation_data = relation.data[self.charm.app]
+                relation_data.update(http_endpoint.dump())
```
james-garner-canonical: Here's how this would look if we used [ops.Relation.save](https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.Relation.save):
```suggestion
                relation.save(http_endpoint)
```

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
```diff
@@ -0,0 +1,237 @@
...
+            raise DataValidationError(msg) from e
+        except json.JSONDecodeError as e:
+            msg = f'invalid databag contents: expecting json. {databag}'
+            logger.error(msg)
+            raise DataValidationError(msg) from e
+
+        return data_model
+
+
+class HttpEndpointConfigChangedEvent(EventBase):
```
james-garner-canonical: I'm a bit unclear on the value of custom events for the provider side. It looks like the event is only emitted via the public `update_config` method, which is presumably intended to be called explicitly by charms. The provider object observes the custom event and uses it to trigger a `_configure` call, but this could be replaced by calling `_configure` directly in `update_config`.

Would it be better to drop the custom events for the provider side?

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
```diff
@@ -0,0 +1,237 @@
...
+        self.relation_name = relation_name
+
+        self.path = path
+        self.scheme = scheme
+        self.listen_port = listen_port
+
+        self.framework.observe(charm.on[relation_name].relation_broken, self._configure)
+        self.framework.observe(charm.on[relation_name].relation_changed, self._configure)
+        self.framework.observe(self.on.http_endpoint_config_changed, self._configure)
+        self.framework.observe(charm.on.config_changed, self._configure)
```
james-garner-canonical: If I understand this correctly, we're assuming that we (the provider) only need to update the relation data if:
1. The relation changes (presumably when a new requirer joins) -- I'm not sure why we update it on relation broken though?
2. Our config changes (possibly due to the charm being upgraded, per your other thread -- worth adding a comment about this here?). But is this the only time the ingress URL might change?

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
```diff
@@ -0,0 +1,237 @@
...
+        self.listen_port = listen_port
+        self.on.http_endpoint_config_changed.emit()
+
+
+class HttpEndpointAvailableEvent(EventBase):
+    """Event emitted when an HTTP endpoint becomes available."""
+
+
+class HttpEndpointUnavailableEvent(EventBase):
+    """Event emitted when an HTTP endpoint becomes unavailable."""
```
james-garner-canonical: It looks like charms are intended to observe these events. I wonder how intuitive their use is though. The names kind of imply a change to me, like an endpoint has just become (un)available. But unavailable is emitted if there are no endpoints, while available is emitted otherwise.

So for example, if we're currently related to 3 apps, and we get a relation joined when a new app joins, we'll get an endpoint available event. But likewise if we got a relation broken event, and now only 2 apps were related, we'd get an endpoint available event, which feels a little unintuitive to me. Likewise if we have an (unrelated) config changed event, we'll get an available/unavailable event emitted based on the (unchanged) current state of our relations.

This makes me wonder if it would be better to drop the events and requirer side observing entirely and just have `get_http_endpoints` (or `get_urls`, see earlier comment) be the entire public interface for requirers. Charms can then either call that method unconditionally (for reconciler style charms), or call it in their observers for the relevant (relation) events (for delta charms).

PS: Why does the requirer observe `config_changed`?

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
chanchiwai-ray: Updated

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
chanchiwai-ray: Thanks for pointing out! Updated

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
chanchiwai-ray: I removed these events now

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
chanchiwai-ray: Yes, relation_broken might not be necessary, unless we want to clear the app data. let me check again.  charm upgrade / configure changed might cause an ingress URL, most likely due to charm author uses `update_config` method in the charm code, so it's good to observe this here, rather than requiring each charm author to explicitly observe that

chanchiwai-ray: >  I'm not sure why we update it on relation broken though? 

Not necessary, removed

interfaces/http_endpoint/src/charmlibs/interfaces/http_endpoint/_http_endpoint.py
```diff
@@ -0,0 +1,177 @@
...
+from ops import CharmBase, EventBase, Object
+from pydantic import BaseModel, HttpUrl, ValidationError
+
+logger = logging.getLogger(__name__)
+
+
+class _HttpEndpointDataModel(BaseModel):
+    """Data model for http_endpoint interface."""
+
+    url: HttpUrl
```
arturo-seijas: I can see a scenario in which we'll want multiple endpoints, for instance, when the backend has multiple nodes and LB is done on the client side. WDYT about having a list of urls instead?

dimaqq: I've asked at https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$SGb0-faFXL26Ta2_9X9GaEUa6PdIuEX2MC456YjsEgI?via=ubuntu.com&via=matrix.org&via=tchncs.de wrt. rel data vs hook command.

dimaqq: Juju response:

> When you call network-get, it is from the calling unit's perspective, effectively saying, this is my ingress and egress in the context of the relation.
It is set in relation data so that counterparts can answer the question, where are you and where will traffic from you appear to come from?
So use relation data to ask about others and network-get to ask about yourself.

Thus, I think the correct behaviour would be along the lines of:
* when the relation exists
* this side calls `network-get` with the relation id (`ops.hookcmds.network_get(rel_id)` or `ops.Model.get_binding(rel_object)`)
* this side then interpolates the address and port and forms a URL (or base URL)
* this side places the URL in own unit|app databag on the relation
* the other side accesses this app using the URL from the relation

Note that this setup (or in fact any other) doesn't capture the possibility of the machine ip address getting changed. Juju promises to recompute the addresses on config-changed, but that's not guaranteed to happen in a timely manner. See something like `mysql-operator/src/ip_address_observer.py` for existing solution.

interfaces/http_endpoint/pyproject.toml
```diff
@@ -0,0 +1,72 @@
+[project]
+name = "charmlibs-interfaces-http-endpoint"
```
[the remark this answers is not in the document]
james-garner-canonical: To be honest, given that the two charms are in a monorepo already, I'd be inclined to skip PyPI and split the library code between the two charms. I think the only thing that would actually need to be duplicated the pydantic model, which is very small, and so seems totally acceptable to duplicate (to me).

I actually don't know which is the provider and which is the requirer, but how about something like this?
```
falco/src/http_endpoint_requirer.py
falco_sidekick/src/http_endpoint_provider.py
```
Though maybe a more falco specific interface name could be a good idea too, depending how you think the relation between the two might evolve.

Sorry for the back and forth about this, it took me a little bit to get my head around what the library was doing, the context for the PR, and the Juju networking side. I do like the simplicity of this interface, and I think it could be generally useful -- for example, if there already existed an interface like this, I assume you'd happily use it for `falco`. But I think a bit more design (e.g. a spec) around semantics and backwards compatibility, along with some additional use cases would probably be in order before adding a new 'official' interface, given the struggles that a lot of interfaces have had with compatibility across upgrades etc.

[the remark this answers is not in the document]
chanchiwai-ray: > Given that this interface is between falco and falco-sidekick, how do you feel about keeping this interface private? Then the package name could be <your team>-interface-http-endpoint, you'd host it in your github project, push to PYPI and consume from PYPI in the two charms?

Oops, thanks. It's not my intention to make it owned by `charmlibs`. I will update that.

> I'd be inclined to skip PyPI and split the library code between the two charms.

That's my initial approach... cc @arturo-seijas what do you think? However, if the above issue was fixed, is that still a concern for you?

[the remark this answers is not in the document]
arturo-seijas: HI!

I don't consider this to be a Falco specific use case. I would expect more charm to rely on this interface, for instance expect this interface to replace the milter interface for the unpublished DKIM signing charm.

With regards to duplicating the schema in both charms, it seems to me like a cohesion problem that we can avoid by having a library, so I don't really see the point of this approach.

If you prefer, we can put it in our Falco charm, but I strongly believe this would be a better place since is not charm specific but we'd still like to know what are the requirements to land an interface in this repository

PR #312, opened 2026-01-26
feat: codeowners for Identity interfaces

james-garner-canonical:
This PR updates `CODEOWNERS` for all the interfaces with `owner: identity`.

It also adds the new `lib`, `summary`, and `description` fields to the interfaces that aren't marked with `internal: true`. These fields will be consumed by Charmhub in future. They may also be used in the `charmlibs` docs. They're intended as high-level descriptions for charmers browsing interfaces and interface libraries.


james-garner-canonical:
Thanks for the reviews, merging this shortly.

PR #317, opened 2026-01-28
feat: codeowners for Data Platform and Analytics interfaces

james-garner-canonical:
This PR adds the Data Platform teams as `CODEOWNERS` for their interfaces.

This PR also uses Copilot, backed by Haiku, to add the new `lib`, `summary`, and `description` metadata fields to the interfaces owned by the `data-platform` teams, and update `CODEOWNERS` accordingly. This is a bit of an experiment, and it would be great to get feedback from the team if the descriptions are accurate, as I'm not familiar with the details of each interface.

These fields will be consumed by Charmhub in future. And will also be displayed in the `charmlibs` docs. They're intended as high-level descriptions for charmers browsing interfaces and interface libraries.

With the `CODEOWNERS` entries added, the teams will be able to review future updates to the interfaces themselves.


taurus-forever (approved):
LGTM, tnx!

PR #318, opened 2026-01-29
feat: add codeowners and update metadata for TLS interfaces

james-garner-canonical:
This PR adds the TLS team as `CODEOWNERS` for the Vault interfaces, and adds the new metadata fields to the public interfaces owned by the team. I would appreciate it if the team could suggest `summary` and `description` fields for the `vault-kv` interface.


interfaces/index.json
```diff
@@ -32,11 +32,11 @@
   {
```
james-garner-canonical: This file is autogenerated from the `interface.yaml` files and doesn't need to be reviewed manually.

PR #321, opened 2026-01-29
feat: `sloth` charm interface

mmkay:
SLO (Service Level Objective) management library for Juju charms, providing integration with the Sloth operator for generating Prometheus recording and alerting rules.

It is developed together with [sloth-k8s-operator](https://github.com/canonical/sloth-k8s-operator), a charm to generate and manage SLOs. 

Tandem PRs for the requirer: https://github.com/canonical/sloth-k8s-operator/pull/9

An example provider: https://github.com/canonical/prometheus-k8s-operator/pull/767


mmkay:
Hi @james-garner-canonical - it is a library that did not eventually get released as a charm library. Before we had any actual production usage (or any PRs in the charms merged for its support), we decided to pivot and start with including this here in charmlibs from the very beginning. The design of the databag contents is part of this PR and there indeed is a spec that I have sent over.

interfaces/slo/src/charmlibs/interfaces/slo/slo.py
```diff
@@ -0,0 +1,466 @@
...
+
+        return slo_spec
+
+    def provide_slos(self, slo_config: str) -> None:
+        """Provide SLO specifications to Sloth as a raw YAML string.
+
+        This method accepts a raw YAML string containing one or more SLO specifications.
+        Multiple specs should be separated by YAML document separators (three dashes).
+        The YAML is validated for parseability but full validation happens on the
+        requirer side.
```
james-garner-canonical: Since the validator code lives in the library, and the round trip to the related application would add delay, why not run the validation logic on the requirer side too? I guess validating on only one side allows us to avoid false positives if validation tightens (and the provider is newer), and false negatives if validation loosens (and again the provider is newer).

james-garner-canonical (commented):
I will continue this review tomorrow, but feel free to address any comments in the meantime -- don't wait on my account. Thanks again for the PR and for your patience.

interfaces/slo/src/charmlibs/interfaces/slo/slo.py
```diff
@@ -0,0 +1,466 @@
...
+        if self._inject_topology:
+            slo_specs = [self._inject_topology_into_slo(spec) for spec in slo_specs]
+
+        # Merge multiple specs back into a single YAML with document separators
+        slo_yaml_docs = [yaml.safe_dump(spec, default_flow_style=False) for spec in slo_specs]
+        merged_yaml = '---\n'.join(slo_yaml_docs)
+
+        for relation in relations:
+            # Write SLO spec to app databag so it's shared across all units
+            relation.data[self._charm.app]['slo_spec'] = merged_yaml
```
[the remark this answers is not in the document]
dimaqq: A couple of comments on the above analysis:
- there's also `inject_topology_something` step
- there's no validation for `objective` or `alerting`, there's partial validation and `slo_spec` dict is passed on verbatim

interfaces/slo/src/charmlibs/interfaces/slo/_slo.py
```diff
@@ -0,0 +1,355 @@
...
+        query: The Prometheus query string
+        topology: Dictionary of label names to values (e.g., {"juju_application": "my-app"})
+
+    Returns:
+        Query with topology labels injected
+
+    Examples:
+        >>> inject_topology_labels(
+        ...     'sum(rate(metric[5m]))',
+        ...     {"juju_application": "my-app"}
```
dimaqq: Suppose that the charm is a database charm that creates a new logical database for each connected "API" application.

How would the charm add the additional/own topology bits?

interfaces/slo/src/charmlibs/interfaces/slo/_slo.py
```diff
@@ -0,0 +1,355 @@
...
+
+    def _get_topology_labels(self) -> Dict[str, str]:
+        """Get Juju topology labels for this charm.
+
+        Returns:
+            Dictionary of topology labels (juju_application, juju_model, etc.)
+        """
+        if JujuTopology is None:
+            # Fallback to basic topology if cosl is not available
+            return {'juju_application': self._charm.app.name}
```
dimaqq: What about the Juju model name?
Or is that inferred on the other side of the relation?

interfaces/slo/src/charmlibs/interfaces/slo/_slo.py
```diff
@@ -0,0 +1,355 @@
...
+
+class SLOError(Exception):
+    """Base exception for SLO library errors."""
+
+
+class SLOValidationError(SLOError):
+    """Validation error for SLO specifications."""
+
+
+def inject_topology_labels(query: str, topology: Dict[str, str]) -> str:
```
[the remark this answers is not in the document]
james-garner-canonical: My recommendation would be to make this private, since it seems like the intended UX is for this to be handled automatically by the library (controlled via an init parameter).

interfaces/slo/src/charmlibs/interfaces/slo/_slo.py
```diff
@@ -0,0 +1,355 @@
...
+"""Internal implementation of the SLO Provider and Requirer library.
+
+This module contains the core implementation classes for the SLO interface.
+For user-facing documentation, see the package-level docstring in __init__.py.
+"""
+
+import copy
+import logging
+import re
+from typing import Any, Dict, List, Match
```
james-garner-canonical: Since this library is Python 3.10+, we can use the builtin `dict` and `list` in annotations, and [re.Match instead of typing.Match](https://docs.python.org/3/library/typing.html#typing.Match).

interfaces/slo/src/charmlibs/interfaces/slo/_slo.py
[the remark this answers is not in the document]
mmkay: Extracted it to a helper module. For making it private, I think @PietroPasotti's comment is valid here: https://github.com/canonical/charmlibs/pull/321#discussion_r2764106440

interfaces/sloth/src/charmlibs/interfaces/sloth/_sloth.py
```diff
@@ -0,0 +1,355 @@
...
+
+        return match.group(0)
+
+    # Match metric_name[time]
+    query = _METRIC_WITH_TIME.sub(replace_time, query)
+
+    return query
+
+
+class SLOSpec(BaseModel):
```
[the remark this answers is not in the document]
james-garner-canonical: If pre-validation is a good idea, then wouldn't it be worth having the lib do it automatically? On the other hand, if requirer-side-only validation is preferred for greater flexibility with updates, then giving charms a convenient tool to do pre-validation seems like it could be a bit of a footgun.

interfaces/slo/src/charmlibs/interfaces/slo/_slo.py
james-garner-canonical: Our policy for `charmlibs` is that releases from `main` support no older than Python 3.10+ -- the tools we use in `main` aren't guaranteed to be compatible with Python 3.8, and we don't test with Python 3.8 in CI.

As this is a brand new interface, we'd encourage you to release it for Python 3.10+ only. If you have charms that are stuck on Python 3.8, I'd imagine they should be in maintenance mode rather than adding new features -- but happy to chat more about this.

---

A couple more thoughts: If the charms support multiple bases, perhaps they could drop Ubuntu 20.04 from their main releases, and make maintenance releases only for older bases? I know things are different for subordinate charms, but I imagine `sloth` is more for workload charms. And it's good to decrease the charm maintenance burden and gradually create incentives for end users to upgrade to supported versions.

That said, we do have the `20.04-maintenance` branch that still has support for Python 3.8. If you do need Python 3.8 support after releasing your 1.0 version from `main`, would making an (e.g.) 0.8.0 release from `20.04-maintenance` with Python 3.8 support work?

dstathis: The libraries we write are 3.8+ not 3.8 specifically. They will work just fine with tools written for 3.10.

While I agree with you that we probably can and should move to 3.10+, the current lib will work just fine.

interfaces/sloth/src/charmlibs/interfaces/sloth/_sloth.py
```diff
@@ -0,0 +1,279 @@
...
+import yaml
+from pydantic import BaseModel, Field, ValidationError, field_validator
+
+from ._topology import inject_topology_labels
+
+try:
+    from cosl import JujuTopology
+except ImportError:
+    # Fallback if cosl is not available
+    JujuTopology = None
```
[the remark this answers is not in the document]
mmkay: I've reconsidered this and IMO we can require `cosl` as a direct dependency. That way it's also much simpler and we don't need any quirky fallbacks.

interfaces/sloth/README.md
```diff
@@ -0,0 +1,67 @@
+# charmlibs.interfaces.sloth
+
+Sloth integration library for Juju charms, providing SLO (Service Level Objective) management with the Sloth operator for generating Prometheus recording and alerting rules.
+
+To install, add `charmlibs-interfaces-sloth` to your Python dependencies. Then in your Python code, import as:
+
+```py
+from charmlibs.interfaces.sloth import SLOProvider, SLORequirer
+```
```
james-garner-canonical: How about merging this with the 'Usage' section, and maybe rename to 'Getting started' or similar?

interfaces/sloth/tests/functional/conftest.py
```diff
@@ -0,0 +1,15 @@
+# Copyright 2026 Canonical Ltd.
```
james-garner-canonical: You can remove this entire directory since functional tests don't make much sense for this library.

interfaces/sloth/tests/unit/conftest.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2026 Canonical Ltd.
```
james-garner-canonical: You can drop this file if you want.

james-garner-canonical:
PS: Please merge `main` to pull in the latest CI changes. We've bumped the target Python version for `ruff` formatting in `main`.

interfaces/sloth/tests/unit/test_version_in_charm.py
```diff
@@ -0,0 +1,26 @@
+# Copyright 2026 Canonical Ltd.
```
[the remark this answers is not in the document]
mmkay: Done.

interfaces/sloth/tests/functional/conftest.py
mmkay: Done, good idea.

mmkay:
hi @james-garner-canonical - the provider & requirer have been updated to be `SlothProvider` and `SlothRequirer`. 

> I also want to mention again the question of whether SLOSpec and inject_juju_topology_labels should be part of the library's public API. My recommendation would be to start with them private unless you're already using them in your team's charms. It's not much work to make a release that makes them public if you find you want them to be later. That said, if you're sure you want them to be public from the get go, this isn't a blocker for me :)

IMO we'd like to keep them public. I think we'll be using them soon in some scenarios in the charms too. I understand that deprecating the method is more complicated than introducing it, however it looks like they will be useful as discussed above.

interfaces/sloth/README.md
mmkay: Moved it a bit - hope this looks better now.

interfaces/sloth/pyproject.toml
```diff
@@ -0,0 +1,82 @@
...
+    "ops",
+    "pydantic>=2.0",
+    "pyyaml",
+    "cosl>=0.0.10",
+]
+
+[dependency-groups]
+lint = [  # installed for `just lint interfaces/sloth` (unit, functional, and integration are also installed)
+    "typing_extensions",
+    "cosl>=0.0.10",
```
james-garner-canonical:
```suggestion
```

james-garner-canonical (approved):
LGTM, thanks.

I've noted some redundant dependencies that can be dropped from the dependency groups.

I think you should be able to merge this PR yourself when ready as `tracing-and-profiling` are already contributors. LMK if you can't and I'll push the button.

I've set up trusted publishing so the library should release on merge.

PR #322, opened 2026-01-29
fix: Make  hashable to prevent unnecessary recreation of certificate requests in relation data.

saltiyazan:


dimaqq (approved):

interfaces/tls-certificates/src/charmlibs/interfaces/tls_certificates/_version.py
```diff
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "1.6.0"
+__version__ = "1.7.0"
```
dimaqq: **minor**: given that it's a bug fix, I would expect 1.6.1

PR #325, opened 2026-02-03
ci: lint should exit with failure when any linting job fails

james-garner-canonical:
This PR fixes a bug in CI, where `just lint <package>` always exits with success, even if individual checks fail. This masked type checking failures in the TLS Certificates library (resolved in #326), which is why those jobs are failing in this PR.


interfaces/tls-certificates/src/charmlibs/interfaces/tls_certificates/_tls_certificates.py
```diff
@@ -56,7 +56,7 @@ class _OWASPLogEvent:
     level: str
     description: str
     type: str = "security"
-    labels: dict[str, str] = field(default_factory=dict)
+    labels: dict[str, str] = field(default_factory=dict[str, str])
```
james-garner-canonical: The new `pyright` version narrows the type of `labels` based on the assigned type (which is derived from `default_factory`) so it will complain without this change.

The default value will not be any different at runtime.

This change doesn't require a release, as this field is in a private class.

pathops/src/charmlibs/pathops/_types.py
```diff
@@ -22,7 +22,7 @@
 
 if typing.TYPE_CHECKING:
     import os
-    from collections.abc import Generator, Sequence
+    from collections.abc import Iterator, Sequence
```
james-garner-canonical: See the changelog for discussion of this change.

pathops/tests/typecheck.py
```diff
@@ -35,3 +35,13 @@ def typecheck_local_path_implemnents_protocol(path: LocalPath) -> None:
...
+
+
+def typecheck_container_path_is_compatible_with_protocol() -> None:
+    class SubClass(ContainerPath, PathProtocol):  # pyright: ignore[reportUnusedClass]
+        pass
+
+
+def typecheck_local_path_is_compatible_with_protocol() -> None:
+    class SubClass(LocalPath, PathProtocol):  # pyright: ignore[reportUnusedClass]
+        pass
```
james-garner-canonical: We already had tests covering compatibility with the protocol (which is why there was a `lint` failure prompting the changes made in this PR), but the error messages aren't very informative. Errors here will give a more informative error message, showing exactly how the classes are incompatible.

PR #327, opened 2026-02-03
ci: pass `package` arg from `lint` to `fast-lint` to scope linting

james-garner-canonical:
`just fast-lint` runs fast linting checks for all files in the repo. `just lint <package>` runs `fast-lint`, as well as the slower type checking jobs for a `<package>`.

This means that `just lint <package>` can show fast linting errors from other packages or global files. This isn't a problem in CI since we keep it green (bugs aside!), but it can be annoying locally.

This PR updates the `fast-lint` to take a path argument, defaulting to `.`, the `justfile` directory (the repository root). Since `ruff` runs recursively on subdirectories, this preserves the existing behaviour. The `lint` recipe has been updated to pass the `<package>` argument to `fast-lint`, scoping the fast linting checks to the package actually being targeted.

### DIFF
--- a/justfile
+++ b/justfile
@@ -43,12 +43,12 @@ init *args:
     @env CHARMLIBS_TEMPLATE=$(realpath .template) uvx cookiecutter .template {{args}}
 
 [doc('Run `ruff`, failing afterwards if any errors are found.')]
-fast-lint:
+fast-lint path='.':
     #!/usr/bin/env -S bash -xueo pipefail
     FAILURES=0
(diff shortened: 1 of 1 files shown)


james-garner-canonical:
> I have no objection to this change, and it seems to work in my limited testing (and limited just skills).
> 
> I'm curious why it's needed: given that, as you say, main should be green, why are errors appearing outside what I'm working on?

For me this comes up when tinkering with multiple packages at once, working on CI and a package at the same time, or having throwaway scripts in my worktree. Everything will need to be green to merge, but it's nice when working locally to be able to easily verify if specific packages are clean or not.

PR #337, opened 2026-02-10
feat: add grub lib

dwilding:
This PR migrates `operator_libs_linux.v0.grub` to `charmlibs`.

**[Preview reference docs](https://canonical-ubuntu-documentation-library--337.com.readthedocs.build/charmlibs/reference/charmlibs/grub/)**

I've added comments to highlight changes that I made during the migration.

To get the unit and functional tests working, I needed to reference the private `_grub` module in several places. See https://github.com/canonical/charmlibs/pull/337/changes/b49cf0388c1e6c9ad83964b77422b130f5e186fe and https://github.com/canonical/charmlibs/pull/337/changes/fc61f123965b709eef7eacdec6c7f1cf503106ad.

### DIFF
--- a/grub/src/charmlibs/grub/__init__.py
+++ b/grub/src/charmlibs/grub/__init__.py
@@ -52,6 +52,12 @@ Example of use::
 """
 
 from ._grub import (
+    CHARM_CONFIG_PREFIX,
+    CONFIG_DESCRIPTION,
+    CONFIG_HEADER,
+    FILE_LINE_IN_DESCRIPTION,
+    GRUB_CONFIG,
+    GRUB_DIRECTORY,
     ApplyError,
     Config,
     IsContainerError,
@@ -62,6 +68,12 @@ from ._grub import (
 from ._version import __version__ as __version__
 
 __all__ = [
+    'CHARM_CONFIG_PREFIX',
+    'CONFIG_DESCRIPTION',
+    'CONFIG_HEADER',
+    'FILE_LINE_IN_DESCRIPTION',
+    'GRUB_CONFIG',
+    'GRUB_DIRECTORY',
     'ApplyError',
     'Config',
     'IsContainerError',
--- a/grub/tests/functional/test_grub.py
+++ b/grub/tests/functional/test_grub.py
@@ -26,8 +26,8 @@ logger = logging.getLogger(__name__)
 def clean_configs():
     """Clean main and charms configs after each test."""
     yield  # run test
-    _grub.GRUB_CONFIG.unlink(missing_ok=True)
-    for charm_config in _grub.GRUB_DIRECTORY.glob(f'{_grub.CHARM_CONFIG_PREFIX}-*'):
+    grub.GRUB_CONFIG.unlink(missing_ok=True)
+    for charm_config in grub.GRUB_DIRECTORY.glob(f'{grub.CHARM_CONFIG_PREFIX}-*'):
         charm_config.unlink(missing_ok=True)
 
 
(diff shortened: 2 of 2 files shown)


grub/src/charmlibs/grub/_grub.py
```diff
@@ -0,0 +1,384 @@
...
+from ._version import __version__
+
+logger = logging.getLogger(__name__)
+
+GRUB_DIRECTORY = Path('/etc/default/grub.d/')
+CHARM_CONFIG_PREFIX = '90-juju'
+GRUB_CONFIG = GRUB_DIRECTORY / '95-juju-charm.cfg'
+CONFIG_HEADER = f"""# This config file was produced by GRUB lib v{__version__}.
+# https://documentation.ubuntu.com/charmlibs/reference/charmlibs/grub/
+"""
```
dwilding: Since `CONFIG_HEADER` contains the lib version and URL, this introduces a behaviour change. The original version was:

```python
CONFIG_HEADER = f"""# This config file was produced by GRUB lib v{LIBAPI}.{LIBPATCH}.
# https://charmhub.io/operator-libs-linux/libraries/grub
"""
```

See https://github.com/canonical/charmlibs/pull/337/changes/0ddcbf3330b2d56f2145797238e1179cdfbb86fe

james-garner-canonical: Change seems perfectly acceptable to me. Could even say: `produced by charmlibs.grub ({__version__}).`

grub/src/charmlibs/grub/_grub.py
```diff
@@ -0,0 +1,384 @@
...
+    to update it with simple validation, remove charm option and apply those changes.
+    """
+
+    _lazy_data: dict[str, str] | None = None
+
+    def __init__(self, charm_name: str) -> None:
+        """Initialize the GRUB config."""
+        self._charm_name = charm_name
+
+    def __contains__(self, key: object) -> bool:
```
dwilding: I corrected the type of `key`, from `str` to `object`. See https://github.com/canonical/charmlibs/pull/337/changes/a1c14d422fd897137beb9c87b0c2d75641033849.

grub/src/charmlibs/grub/_grub.py
james-garner-canonical: Though maybe it's wiser to keep the format as close to the previous one as possible.

grub/tests/functional/test_grub.py
```diff
@@ -0,0 +1,215 @@
...
+    [
+        {'GRUB_CMDLINE_LINUX_DEFAULT': '$GRUB_CMDLINE_LINUX_DEFAULT hugepagesz=1G'},
+        {
+            'GRUB_CMDLINE_LINUX_DEFAULT': '$GRUB_CMDLINE_LINUX_DEFAULT hugepages=64 hugepagesz=1G',
+            'GRUB_DEFAULT': '0',
+        },
+        {'GRUB_TIMEOUT': '0'},
+    ],
+)
+def test_single_charm_valid_update(config: dict[str, str]):
```
dwilding: I added a type annotation for `config`, and some other type annotations elsewhere. See https://github.com/canonical/charmlibs/pull/337/changes/67220eb12982c997bee4a3c6c9b0b90877d64b40

grub/src/charmlibs/grub/__init__.py
```diff
@@ -0,0 +1,71 @@
...
+            self.framework.observe(self.on.install, self._on_install)
+            self.framework.observe(self.on.update_status, self._on_update_status)
+            self.framework.observe(self.on.remove, self._on_remove)
+            self.grub = grub.GrubConfig(self.meta.name)
+            log.debug("found keys %s in GRUB config file", self.grub.keys())
+
+        def _on_install(self, _):
+            new_value = "$GRUB_CMDLINE_LINUX_DEFAULT hugepagesz=1G"
+            try:
+                self.grub.update({"GRUB_CMDLINE_LINUX_DEFAULT": new_value})
```
dwilding: I pulled `new_value` out of the dictionary because the line would be too long otherwise. The example could be generally improved, but that's out of scope of this PR.

james-garner-canonical (approved):
Thanks, @dwilding, this looks great! And thanks for structuring the commits to make it easy to review, that was very helpful. One thing that would make it a tinier bit easier in future would be if the test files were also added in the initial few commits.

I've set up publishing on the PyPI side, so this should get published when we merge.

james-garner-canonical (changes requested):
One thing just occurred to me though, for complete backwards compatibility the various constants in the `_grub.py` module should also be exported in `__init__.py`.

They don't really look like they're intended for end users to use or modify directly, but I think we should be consistent with what we did for the other `operator-libs-linux` libraries and just export the same public symbols for the initial release as the Charmhub-hosted version did, even if it's a bit messy.

PR #343, opened 2026-02-19
docs: add a tips for real libraries to tutorial

james-garner-canonical:
I've noticed that a couple of new interface libraries have started out with `just init` instead of `just interface init`. I'm guessing this is because they're following the tutorial, which assumes the a non-interface library.

This PR adds a tip box the instruction to run `just init` to say you should use `just interface init` for interface libraries, and adds some advice for integration tests for interface libraries. It also adds a couple of tips around unit and functional tests to contextualise the tutorial's steps for real libraries.

**[[Preview build]](https://canonical-ubuntu-documentation-library--343.com.readthedocs.build/charmlibs/tutorial/)**

### DIFF
--- a/.docs/tutorial.md
+++ b/.docs/tutorial.md
@@ -235,6 +235,11 @@ def _on_start(self, event: ops.StartEvent):
 
 Now run `just unit uptime` to verify that both tests pass.
 
+````{tip}
+If you're working on a real library you intend to open a PR for, you should remove `test_version.py` and `test_version_in_charm.py` once you have some working unit tests of your own.
+You should also feel free to split your tests across as many files as needed to keep them organised nicely.
+````
+
 For more on `ops.testing`, see:
 - [ops.testing reference docs for custom events](ops.testing.CharmEvents.custom)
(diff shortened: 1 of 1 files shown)


.docs/tutorial.md
```diff
@@ -286,6 +301,14 @@ Under `src/`, you'll see a unique `charm.py` file, and a symlink to `common.py`.
 
 Our `uptime` function should work just as well in a K8s charm as in a machine charm, so we'll test on both substrates, meaning that we don't need to change anything so far.
 
+````{tip}
+If you're working on an integration library, the K8s / machine distinction may or may not be important, depending on what information your library needs to populate the databags with.
+As long as your library is intended to work on both substrates, it's a good idea to test on both.
+
+However, whether you test on both substrates or just one, you'll definitely want to be packing both a requirer and provider charm.
+Consider adding `provider` and `requirer` directories under `tests/integration/charms`, duplicating the existing charm structure in each, updating `pack.sh` to pack both `requirer` and `provider` charms, and deploying both in `conftest.py`.
```
james-garner-canonical: I should probably update `just interface init` to do this for interface libraries, but I don't have the time right now.  I've created #344 to track this.

PR #349, opened 2026-02-26
docs: fix build failures + minor cleanup

james-garner-canonical:
This PR fixes the docs build failure that was recently introduced on `main` via #333.

As a driveby, I've also fixed the incorrectly formatted link in the `tls-certificates` package docstring (added by me originally), and added new Python fields to exclude from the docs (we don't exclude all magic methods because `pathops` classes document some of them, but perhaps we should revisit this decision in future).

**[[Preview build]](https://canonical-ubuntu-documentation-library--349.com.readthedocs.build/charmlibs/reference/charmlibs/interfaces/tls-certificates/)**

### DIFF
--- a/.docs/conf.py
+++ b/.docs/conf.py
@@ -200,7 +200,9 @@ autodoc_default_options = {
     'members': None,  # None here means "yes"
     'special-members': None,  # meaning all
     'exclude-members': (
+        '__annotate_func__,'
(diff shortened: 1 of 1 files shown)


james-garner-canonical:
@canonical/tls please review and feel free to merge afterwards

PR #352, opened 2026-03-04
feat: `otlp` interface 

sinapah:
This PR introduces the `otlp` relation interface to `charmlibs`. `otlp` is the standard data format and transport protocol used by Opentelemetry Collector to send observability data and it will play a huge role in how the Opentelemetry Collector charm will forward telemetry to COS going forward.

The first implementation of OTLP rolled out as [a module file within the Opentelemetry Collector K8s repo](https://github.com/canonical/opentelemetry-collector-k8s-operator/blob/main/src/otlp.py) very recently (through work in PR [196](https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/196) by @MichaelThamm). It received other iterations in [193](https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/193) for example.

We have tested sourcing `otlp` from this branch as examples in the following PRs:
- Otelcol K8s: https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/193
- Otelcol machine:  https://github.com/canonical/opentelemetry-collector-operator/pull/175

Other refs:
What is `otlp`: https://opentelemetry.io/docs/specs/otlp/

### DIFF
--- a/interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
+++ b/interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
@@ -190,9 +190,10 @@ class OtlpRequirer:
         return valid_endpoints
 
     def _favor_modern_endpoints(self, endpoints: list[_OtlpEndpoint]) -> _OtlpEndpoint:
-        """Return the endpoint with the most modern protocol.
+        """Return the preferred endpoint using protocol priority.
 
-        If an unknown protocol is encountered, it is given the lowest priority.
+        Modern protocols receive higher priority.
+        Protocol ranking is `grpc` > `http`; unknown protocols rank lowest.
         """
         modern_score: Final = {'grpc': 2, 'http': 1}
         return max(endpoints, key=lambda e: modern_score.get(e.protocol, 0))
@@ -246,13 +247,11 @@ class OtlpRequirer:
         """Return a mapping of relation ID to OTLP endpoint.
 
         For each remote's list of OtlpEndpoints, the requirer filters out
-        unsupported endpoints and telemetries. If there are multiple supported
-        endpoints, the requirer chooses the first available endpoint in the
-        list. This allows providers to specify multiple endpoints with
-        different protocols and/or telemetry types and the requirer can choose
-        one based on its own capabilities. For example, a provider may specify
-        both an HTTP and gRPC endpoint, and a requirer that only supports HTTP
-        will choose the HTTP endpoint.
+        unsupported endpoints and telemetries. If multiple compatible
+        endpoints remain, the requirer prefers newer protocols (`grpc` over
+        `http`). Unknown protocols are treated as the lowest priority. This
+        allows providers to expose multiple endpoints with different protocol
+        and telemetry combinations while the requirer selects the best match.
         """
         endpoint_map: dict[int, OtlpEndpoint] = {}
         for relation in self._charm.model.relations[self._relation_name]:
--- a/interfaces/otlp/src/charmlibs/interfaces/otlp/__init__.py
+++ b/interfaces/otlp/src/charmlibs/interfaces/otlp/__init__.py
@@ -91,6 +91,9 @@ subset of protocols and telemetries, which can be configured at instantiation::
 Given the defined, supported protocols and telemetries, the OtlpRequirer will filter out
 unsupported endpoints and prune unsupported telemetries. After filtering, requirer selection
 condenses the list to a single endpoint per relation.
+Endpoints with modern protocols are favoured over legacy ones.
+That means an endpoint supporting the `gRPC` protocol will be selected over one supporting `HTTP`.
+Unknown protocols will receive the lowest priority.
 
 The OtlpRequirer also publishes rules to related OtlpProvider charms with the ``publish()``
 method::


interfaces/otlp/src/charmlibs/otlp/_otlp.py
```diff
@@ -0,0 +1,374 @@
...
+DEFAULT_PROVIDER_RELATION_NAME = 'receive-otlp'
+DEFAULT_LOKI_RULES_RELATIVE_PATH = './src/loki_alert_rules'
+DEFAULT_PROM_RULES_RELATIVE_PATH = './src/prometheus_alert_rules'
+
+
+logger = logging.getLogger(__name__)
+
+
+class RulesModel(BaseModel):
+    """A pydantic model for all rule formats."""
```
dimaqq: The class makes it pretty clear that it's a Pydantic model.
Instead, please consider describing what an object of this model represents.

interfaces/otlp/src/charmlibs/otlp/_otlp.py
```diff
@@ -0,0 +1,374 @@
...
+    following command:
+    ```bash
+    <rules-from-show-unit> | base64 -d | xz -d | jq
+    ```
+    """
+
+    model_config = ConfigDict(extra='forbid')
+
+    rules: RulesModel | str
+    metadata: OrderedDict[str, str]
```
dimaqq: Please document what metadata can and cannot be.

dimaqq (commented):
A partial code review from me.

interfaces/otlp/pyproject.toml
```diff
@@ -0,0 +1,74 @@
...
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/charmlibs"]
+
+[tool.hatch.version]
+path = "src/charmlibs/otlp/_version.py"
```
james-garner-canonical:
```suggestion
path = "src/charmlibs/interfaces/otlp/_version.py"
```

james-garner-canonical:
Hi @sed-i, I think we're in agreement here -- I don't think splitting this interface up is a good idea due to the UX that would create.

My concern was that the current name feels at first both too broad (OTLP but the relation is only sharing endpoints) and too narrow (the relation also shares non-OTLP rules for specific services in the observability stack). I think this is a good interface design for the observability solution being crafted, I'm just worried that the name doesn't quite fit.

I am open to the name if we're saying "this is the canonical way that charms will set up to use OTLP". I still wonder if there's a name that makes the scope and purpose clearer, but perhaps the current one is the right tradeoff.

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -0,0 +1,364 @@
...
+        for relation in self.model.relations[self._relation_name]:
+            provider = relation.load(OtlpProviderAppData, relation.app)
+            provider.endpoints = self._filter_endpoints(provider.endpoints)
+
+            # Choose the first valid endpoint in list
+            endpoint_choice = next(
+                (e for e in provider.endpoints if e.protocol in self._protocols), None
+            )
+            if endpoint_choice is not None:
+                endpoint_map[relation.id] = endpoint_choice
```
james-garner-canonical: On a second look, it seems that `_filter_endpoints` takes care of the `e.protocol in self._protocols` logic already.

It's probably a bit silly, but it also seems somehow cleaner to me not to mutate the provider object -- I know it doesn't really make a difference since it's ephemeral anyway.

 So how about:
```suggestion
            endpoints = self._filter_endpoints(provider.endpoints)
            if endpoints:
                endpoint_map[relation.id] = endpoints[0]
```

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -0,0 +1,374 @@
...
+                return ''
+
+    @staticmethod
+    def encode_value(obj: Any) -> str:
+        """Encode relation data values using BaseModel serialization."""
+        try:
+            RulesModel.model_validate(obj)
+            return LZMABase64.compress(json.dumps(obj, sort_keys=True))
+        except ValidationError:
+            return json.dumps(obj, sort_keys=True)
```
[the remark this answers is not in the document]
james-garner-canonical: Probably the way to go here is [Pydantic's field_serializer](https://docs.pydantic.dev/latest/api/functional_serializers/#pydantic.functional_serializers.field_serializer)

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -0,0 +1,369 @@
...
+    ```bash
+    <rules-from-show-unit> | base64 -d | xz -d | jq
+    ```
+    """
+
+    rules: RulesModel | str = Field(
+        description='Rules to be forwarded to the provider.'
+        ' Stored as an LZMA-compressed, base64-encoded JSON string to reduce payload size.'
+    )
+    metadata: OrderedDict[str, str] = Field(
```
james-garner-canonical: I think it could make sense for this to stay required in practice, as it's core to the design of how rules are processed, and can act as a sentinel for whether the requirer side databag has been populated or not. Alternatively, it could be `None`-able with the semantics that if it's missing, rules won't be processed as we assume the other side isn't ready.

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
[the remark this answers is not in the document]
tonyandrewmeyer: > Doesn't this make the usage of the `encoder` and `decoder` from the relation `load` and `save` methods redundant? We should use either Pydantic field serializer and deserializers, or the `encoder`/`decoder` arg.

The main use-cases for the encoder/decoder args are (a) you're not using Pydantic, or (b) you want to disable the default JSON encoding/decoding, probably for backwards compatibility reasons.

james-garner-canonical (commented):
Per discussion, the OTLP name is reasonable as this is the canonical way that observability products will use OTLP via Juju in future, moving away from custom formats per application as upstream standardises on OTLP.

The library, documentation, and tests are in good shape, modulo the remaining concerns raised about the databag format:
- The different kind of rules fields should be optional to set the pattern for adding more in future.
- Same concern for top-level `rules` field itself (future proofing for perhaps eventually adding (e.g.) dashboards, as discussed).
- Consider whether the requirer side library API should treat the ordering of endpoints in the databag as meaningful -- see [ordering concerns in spec](https://docs.google.com/document/d/1Zub28RVLE8NFpdITtTdVS9sLwbeiUvmf2u40_rjzX_E/edit?tab=t.ef3zs55gt10f#heading=h.clt2f2vrp55s) and [my specific comment](https://github.com/canonical/charmlibs/pull/352/changes#r2921294584).
    - An interim solution might be to just document this as an implementation detail of the library API currently, not to be relied on.
    - Alternatively, if it's important for the priority to be controlled by the provider side, consider adding priority keys to the databag format.

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -0,0 +1,369 @@
...
+        Other data is serialized into a JSON formatted str.
+        """
+        try:
+            RulesModel.model_validate(obj)
+            return LZMABase64.compress(json.dumps(obj, sort_keys=True))
+        except ValidationError:
+            return json.dumps(obj, sort_keys=True)
+
+
+class OtlpRequirer(Object):
```
james-garner-canonical: Minor: there's no need to inherit from `Object` since this class doesn't observe events or emit custom events. Holding onto a reference to the charm is sufficient for what it's doing.

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -0,0 +1,369 @@
...
+                # the databags haven't initialized yet, continue
+                continue
+            endpoints = self._filter_endpoints(provider.endpoints)
+            if endpoints:
+                endpoint_map[relation.id] = endpoints[0]
+
+        return endpoint_map
+
+
+class OtlpProvider(Object):
```
james-garner-canonical: Likewise, this class shouldn't need to inherit from `ops.Object`.

james-garner-canonical (approved):
I'm happy to approve this PR :)

There's one docstring update that should be taken care of before merging this PR and automatically releasing the library (document prioritising GRPC endpoint over HTTP endpoint).

james-garner-canonical:
Just for clarity: I believe that `observability-core` should be able to merge this PR now that it has been approved by `charmlibs-maintainers`. I'll leave merging to you so you can take care of the docstring update before merging.

It would also be a good idea to add a `CHANGELOG.md` entry like:
```
# 0.1.0 - 13 March 2026

Initial library release.
```

If there are any problem with merging, just let me know.

PR #353, opened 2026-03-05
feat(interface-k8s-backup-target): Add k8s_backup_target interface library

jneo8:
Add the k8s_backup_target interface library for declaring Kubernetes backup specifications.                                                                                                                      
                                                                                                                                                                                                                   
This interface allows client charms to specify what Kubernetes resources should be backed up, and backup integrator charms to consume these specifications and forward them to backup operators.

## How to test

```
just format interfaces/k8s_backup_target
just interfaces-json
just lint interfaces/k8s_backup_target
just unit interfaces/k8s_backup_target
```


interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,254 @@
...
+
+SPEC_FIELD = "spec"
+APP_FIELD = "app"
+RELATION_FIELD = "relation_name"
+MODEL_FIELD = "model"
+
+logger = logging.getLogger(__name__)
+
+
+class BackupTargetSpec(BaseModel):
```
james-garner-canonical: Given that the interface and library name are being hedged with the `k8s` prefix, maybe the library objects should be too?

interfaces/k8s_backup_target/interface/v0/README.md
```diff
@@ -0,0 +1,60 @@
+# `k8s_backup_target/v0`
+
+## Overview
+
+The `k8s_backup_target` interface enables a **client charm** to specify what Kubernetes data should be backed up by relating to a **backup integrator charm**. In practice, a client charm declares the namespaces and resource types to include or exclude in backups, and the backup integrator uses this specification to forward backup requests to a backup operator. This design lets the client define backup requirements **without direct cluster elevated permissions**, delegating the backup execution to the backup operator charm.
```
james-garner-canonical: I suggest we make it clear that the client charm could relate to a backup charm or to a backup integrator charm -- really from the client charm's perspective there's no difference, as they both provide (well, require) the interface.

interfaces/k8s_backup_target/interface/v0/schema.py
```diff
@@ -0,0 +1,100 @@
...
+    )
+    ttl: str | None = Field(
+        None,
+        description="Optional TTL (time-to-live) for the backup (e.g. '72h' or '30d').",
+        title="Backup TTL",
+        examples=["24h"],
+    )
+
+
+class ProviderAppData(BaseModel):
```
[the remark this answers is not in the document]
james-garner-canonical: @dimaqq What I'm mainly wondering is if you think it would be more robust to nest the current model under some top-level key; and if you think it would be better to make it a list of these objects?

@jneo8 I assume there wouldn't be a problem with nesting this object under a top-level key in the on-the-wire databag format; the current library design doesn't benefit from making it a list, but is it at all plausible that a future extension of the interface might want this? For instance, maybe someday the library API is extended so that a coordinator charm can request backups for its worker charms, so they don't all have to relate to the backup manager.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,254 @@
...
+
+class BackupTargetProvider(Object):
+    """Provider class for the backup target configuration relation."""
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        relation_name: str,
+        spec: BackupTargetSpec,
+        refresh_event: BoundEvent | list[BoundEvent] | None = None,
```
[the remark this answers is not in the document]
james-garner-canonical: On the other hand, I see that this class doesn't provide any public way to explicitly send the data.

It does have a nice set of default events to publish on though. When would a charm need to publish on another event?

interfaces/k8s_backup_target/interface/v0/schema.py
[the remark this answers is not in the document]
james-garner-canonical: > That said, the current flat structure is already in use by both velero-operator and velero-integrator-operator. If we adopt the nested format, we'd want to provide clear migration guidance and documentation for teams transitioning from the existing velero backup config interface, as it would require some very little additional effort on their side to adapt.

Do you mean that this interface is in use, or are you talking about compatibility with the valero specific backup interface?

If you do mean this interface, then we probably shouldn't change things, depending on the exact value of 'already in use', could you please elaborate if so.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,186 @@
...
+        relations = self.model.relations[self._relation_name]
+
+        for relation in relations:
+            data = relation.data.get(relation.app, {})
+            if (
+                data.get(APP_FIELD) == app_name
+                and data.get(MODEL_FIELD) == model
+                and data.get(RELATION_FIELD) == endpoint
+            ):
+                json_data = data.get(SPEC_FIELD, "{}")
```
dimaqq: Looks fragile to me, because spec contains the optional field, which when missing means "back up everything".

interfaces/k8s_backup_target/interface/v0/schema.py
```diff
@@ -0,0 +1,94 @@
...
+        description="Name of the relation on the client providing this spec.",
+        title="Client Relation Name",
+        examples=["backup"],
+    )
+    model: str = Field(
+        ...,
+        description="Model name of the client application.",
+        title="Client Model Name",
+        examples=["my-model"],
+    )
```
dimaqq: I feel that these three fields are redundant:

- the requirer can get the remote app name from the relation
- and the remote model name
- what's the use for endpoint, can there be more than one?

dimaqq (commented):
First pass review.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
jneo8: This is intentional and mirrors Velero's own API behavior — a Backup spec with no namespace/resource filters means "back up everything." The client charm (provider) is the one that controls what goes into the spec, so an empty spec is a valid choice.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,186 @@
...
+                return K8sBackupTargetSpec.model_validate_json(json_data)
+
+        logger.warning("No backup spec found for app '%s' and endpoint '%s'", app_name, endpoint)
+        return None
+
+    def get_all_backup_specs(self) -> list[K8sBackupTargetSpec]:
```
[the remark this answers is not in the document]
james-garner-canonical: My guess: all the information needed to perform scheduled(?) backups/etc is in the spec itself, and the app/endpoint/model key is only intended for users to interact with backup functionality via actions. Is that right @jneo8?

But I notice that valero-operator doesn't seem to use `get_all_backup_specs` currently.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,186 @@
...
+            charm: The charm instance that requires backup configuration.
+            relation_name: The name of the relation (from metadata.yaml).
+        """
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+
+    def get_backup_spec(
+        self, app_name: str, endpoint: str, model: str
+    ) -> K8sBackupTargetSpec | None:
```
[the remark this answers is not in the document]
james-garner-canonical: It looks like the charm on this side of the relation relies on Juju user's to supply that information via an action. I'm not clear on whether backups are performed in other ways too.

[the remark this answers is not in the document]
jneo8: It's pass from action

https://github.com/canonical/velero-operator/blob/aa1afee8dcbfa4c3447abd3b908f0fcd219bb6fa/src/charm.py#L241-L266

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
dimaqq: Yeah I figured as much.

What I'm pointing out here is that the decision the code effectively makes to back up or not is not-trivial:
- databag empty: no backup
- databag contains app/model/endpoint but no spec key at all: backup everything
- databag contains app/mod/end and an empty spec object: backup everything
- databag contains app/mode/end and filled-out spec: backup specific things

I agree that it's kinda similar to how k8s often behaves, but it's also a bit counter-intuitive from the perspective of someone trying to integrate over this interface a few years hence.

interfaces/k8s_backup_target/interface/v0/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+[\[Pydantic Schema\]](./schema.py)
+
+On the provider side, the application data bag must contain structure with the following fields:
+
+- `app` (string, required): Name of the client application that requires backups.
+- `relation_name` (string, required): Name of the relation on the client side through which this spec is sent (from metadata.yaml).
+- `model` (string, required): Name of the model where the client application is deployed.
+- `spec` (dict, required): A dictionary defining what to back up. This includes the following keys:
+  - `include_namespaces` (list of str, optional): Specific Kubernetes namespaces to include in the backup. If not provided, all namespaces are included.
```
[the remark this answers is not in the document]
dimaqq: Note that Velero uses `["*"]` notation to indicate all namespaces.
Would that be a good practice in the interface too?
A single element, asterisk, feels very much like K8s.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,252 @@
...
+        Args:
+            charm: The charm instance that requires backup configuration.
+            relation_name: The name of the relation (from metadata.yaml).
+        """
+        super().__init__(charm, relation_name)
+        self._charm = charm
+        self._relation_name = relation_name
+
+    @property
+    def is_ready(self) -> bool:
```
james-garner-canonical: Nice to see this :)

cc: @dimaqq

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,252 @@
...
+            True if at least one relation has parseable backup_targets data.
+        """
+        relations = self.model.relations[self._relation_name]
+        for relation in relations:
+            data = relation.data.get(relation.app, {})
+            raw = data.get(BACKUP_TARGETS_FIELD)
+            if not raw:
+                continue
+            if self._parse_backup_targets(raw):
+                return True
```
james-garner-canonical: If we define `ProviderAppData` in `_schema.py` and import it here as discussed out of band, then WDYT about using `relation.load(ProviderAppData, relation.app)`?

I guess this should look like this:
```suggestion
            try:
                data = relation.load(ProviderAppData, relation.app)
            except (JSONDecodeError, ValidationError):
                continue
            if data.backup_targets:
                return True
```

Which doesn't really save any lines of code here, but I think replaces `_parse_backup_targets` and `_validate_entry` entirely.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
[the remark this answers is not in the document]
james-garner-canonical: Thanks for the context. Let's leave this as-is for now. It's a library-API-only concern, so it can always be safely changed in future anyway.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,252 @@
...
+            for event in refresh_event:
+                self.framework.observe(event, self._send_data)
+
+    def _send_data(self, event: EventBase):
+        """Handle any event where we should send data to the relation."""
+        if not self._charm.model.unit.is_leader():
+            logger.warning(
+                "K8sBackupTargetProvider handled send_data event when it is not a leader. "
+                "Skipping event - no data sent"
+            )
```
james-garner-canonical: Is the expectation that charms should conditionally set up `K8sBackupTargetProvider` in their `__init__` only if they're the leader? If so, this should be documented.

Otherwise, do you think that logging this is necessary? That is, if it's expected that every non-leader charm unit will set up `K8sBackupTargetProvider`, then it's probably fine not to log a warning.

This is a minor thing -- if you consider logging this to be standard practice, then by all means leave it as-is.

interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_backup_target.py
```diff
@@ -0,0 +1,252 @@
...
+class _BackupTargetEntry(BaseModel):
+    """Internal model for a single backup target entry on the wire."""
+
+    app: str
+    relation_name: str
+    model: str
+    spec: K8sBackupTargetSpec
+
+
+class K8sBackupTargetRequirer(Object):
```
james-garner-canonical: This class shouldn't need to inherit from `Object`, since it doesn't observe or emit any events.

james-garner-canonical (commented):
Thanks for all your work on this, @jneo8. In this review I've covered the databag related concerns we discussed in our meeting, as well as a final code review pass.

PS: I think CI is failing because `uv.lock` needs an update with `uv lock`.

interfaces/k8s_backup_target/interface/v0/schema.py
jneo8: In juju cross model relation, the app.name returns munged proxy name, and the remote model isn't accessible from the relation object. These fields carry the real values for action-based lookups.

Check kubeflow backup example also: [doc link](https://documentation.ubuntu.com/charmed-kubeflow/latest/how-to/manage/back-up-control-plane/#:~:text=juju%20switch%20velero%0A%0Ajuju%20run%20velero%2Doperator,workloads%2Dbackup%2Dconfig%20%5C%0A%20%20%20%20model%3Dkubeflow)

PR #366, opened 2026-03-17
ci: support libs with requires-python higher than 3.10

james-garner-canonical:
#364 has revealed that CI breaks with packages with a `requires-python` lower bound that's higher than 3.10 (it has `requires-python>=3.12`).

This is because we would run `just integration-k8s` and `just integration-machine` in the `init` step to check if they have any tests, and that uses our default `justfile` Python version of 3.10.

This PR fixes this by recording the minimum supported Python version when we calculate the Python versions to test the library with, and using that version in the failing step.


tonyandrewmeyer (approved):
LGTM!

PR #372, opened 2026-03-18
docs: how to design relation interfaces

dimaqq:
Make a how-to out of the relation interface design specification.

Preview: https://canonical-ubuntu-documentation-library--372.com.readthedocs.build/charmlibs/how-to/design-relation-interfaces/


.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,391 @@
...
+
+To allow interface evolution, the charm-facing API should be decoupled from the interface parsing code. The Python code in the charm library typically deals with:
+
+- logic: combining fields, filtering values
+- stable Python API: both new and legacy interface fields are processed
+- run-time: wrapping and suppressing errors in further dependencies and secrets
+- arguments: charm context, for example the arguments that charm passes to the library
+- third-party dependencies, for example loading PEM content in `cryptography.x509` primitives
+
+Adopt the following conventions in charm libraries that wrap interfaces.
```
dimaqq: Should be small, because it should live in a separate page in the long term.

.docs/how-to/design-relation-interfaces.md
dimaqq: Open question, if we should somehow reference "OP083 - Relation Interface Design":
- argument from authority: there's an approved spec, follow it!
- for someone years down the line, to look at how it came to be and the extended rationale

dimaqq: https://mystmd.org/spec/footnotes

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+The databag format is long-lived contract, while the charm-facing API is easier to change.
+Please keep the two separate from the start.
+
+Additionally, the ergonomics of using multiple libraries in a charm require specific API patterns.
+When interface evolves, some version of a library has to support both old and new schema, which should not leak to the charms.
+
+### Library API for delta and holistic charms
+
+The ideal charm library API shape is different for delta and holistic charms. Specifically, the comparison between current relation data takes place in the library for a delta charm (against old relation data), and in the charm (againts the workload) in a holistic charm.
+
```
james-garner-canonical: Seems out of place in this document. We hope to add a doc next cycle to cover library API, but this doc is explicitly not about the library API itself.

```suggestion
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+### Process
+
+First, decide what data bits should appear in the databag in the first place.
+
+Then, design the JSON representaion for these bits with provisions for backwards and forwards compatibility.
+
```
james-garner-canonical: I  don't really understand this process, I think it's the term "data bits". Do you mean that first we decide what data needs to be transmitted over the relation (e.g. endpoints, rules), and then choose how to represent that data (e.g. integer, collection of objects)? If so, I think that could be expanded on here.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+Using newer Pydantic, prefer the `MISSING` sentinel value over the more traditional `None`.
+
+```py
+# missing field is read as <MISSING>; deleted when written out
+foo: str | MISSING = MISSING
+
+# missing field is read as None; written out as JSON null
+foo: str | None = None
+```
+
```
james-garner-canonical: Does this recommendation actually come from OP083? I know that OP083 used `MISSING` as its chosen way of representing data that was left out of the databag, and I think it was a good choice for that. What I'm not convinced of is that we should be recommending that libraries adopt this experimental Pydantic feature right now -- and I think that's what we're doing here since schemas are very likely going to end up embedded in libraries at runtime.

We are saying "using newer Pydantic", so the fact that this requires at Pydantic >= 2.12 (October 2025 release), as well as Ops >= 3.6.0 (February 2026) if using `Relation.save`, isn't a big problem.

What does worry me is that `MISSING` requires import from the `pydantic.experimental` namespace, which means that the import location could change (or `MISSING` might go away entirely) in a minor Pydantic version bump. This means that libraries would need to use tight Pydantic version constraints to avoid breakage, but also libraries *shouldn't* use tight library constraints otherwise using multiple libraries can become painful due to dependency resolution issues.

Additionally, `MISSING` won't type check cleanly unless using Pyright >= 1.1.402 with `enableExperimentalFeatures`, and I'm not sure that we should be pushing our libraries to use `enableExperimentalFeatures`.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+## Databag schema
+
+The only changes allowed on a published interface are:
+
+- adding a new field (top level or nested)
+- removing a field (ideally on major version bump)
```
james-garner-canonical:
```suggestion
- removing a field: This is a backwards incompatible change, and must be clearly communicated by a major version bump of the interface library.
```
Also, is there still an obligation for the library not to choke on databags that contain the dropped field, even if they don't understand it anymore?

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+- with extra caution: tweaking field validators; or extending or narrowing an enum range
```
james-garner-canonical: We should be more explicit about what extra caution entails.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+
+class Databag(pydantic.BaseModel):
+    endpoints: frozenset[Endpoint] | MISSING = MISSING
+```
+
+Collections of primitive types are strongly discouraged.
+
+Data maps are strongly discouraged. An exception to this rule if when the data map key is a Juju entity with a well-known string representation, such as unit name or machine id.
+
+### URLs and URIs
```
james-garner-canonical: This section feels a little out of place with the other rules. I wonder about moving it to the end. While the other rules are more general, this rule is for one specific type of value.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+data = {"host": '"fe80::1"'}
+rel = testing.Relation("foo", remote_app_data=data)
+state_in = testing.State(relations={rel})
+state_out = ctx.run(ctx.on.relation_changed(rel), state_in)
+assert state_out.unit_status == ops.testing.BlockedStatus(
+    "foo not ready: host must be a domain name"
+)
+```
+
+[^op083]: OP083 - Relation Interface Design
```
james-garner-canonical: See earlier comment -- let's drop the footnote.
```suggestion
```

dwilding (commented):
I've suggested some restructuring and James has left several comments. I'll wait for you to have a look at these before doing a final pass.

.docs/how-to/design-relation-interfaces.md
dimaqq: Well, I would prefer our charmers to use `pydantic ~= 2.12, ops ~= 3.6, pyright ~= 1.1.402`. And I would prefer **new interfaces** to use `| MISSING` vs `| None`.

The doc is specifically for new interfaces.

So far, I couldn't find any experimental-namespaced feature removed from Pydantic, though an actual experimental feature (in the larger sense), "plugins" did get yanked IIRC.

One could argue that we ought to wait for feature becoming stable, or perhaps until we're sure that the new interface will never be used by a 20.04-based charm, therefore Python 3.8, therefore Ops 2.x and Pydantic 2.11.x. (in which case the library can't use `Relation.load/save`, and needs manual (de)serialisaion anyway).

I think, as the maintainer of this repo, you can totally make the call here @james-garner-canonical .

.docs/how-to/design-relation-interfaces.md
dimaqq: Indeed, library must not choke on the missing field.
However, `.is_foo_ready` is allowed to return `False` for foos that use the field, making the interface unusable to express foo.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+Unexpected enum values should be parsed as `MISSING` or a pre-defined catch-all `UNKNOWN` value:
+
+```py
+foo: Enum(A, B) | MISSING = MISSING
+bar: Enum(UNKNOWN, A, B) = UNKNOWN
+```
+
+### No mandatory fields
+
+Top-level fields must not required or optional.
```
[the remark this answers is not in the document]
dimaqq: not required means that field can be missing
optional means that field can be `None`

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+
+Likewise most sub-fields must be not required or optional.
+
+```py
+role: Role | MISSING = MISSING
+  subject: str | MISSING = MISSING
+  session: str | MISSING = MISSING
+```
+
+A default value may be used instead in some cases.
```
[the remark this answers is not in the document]
dimaqq: re: reader or writer, in the Ops API, it's both for:
- optional fields
- fields with real default values

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+
+### No mandatory fields
+
+Top-level fields must not required or optional.
+
+```py
+foo: str | MISSING = MISSING
+```
+
+Likewise most sub-fields must be not required or optional.
```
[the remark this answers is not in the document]
dimaqq: I didn't want to expand too much in the how-to...

the cases where a sub-field may be marked mandatory are:
- a sub-field in a object in a collection, where the validator for the collection field filters out the misfits
- a sub-field with a real default value

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+
+- adding a new field (top level or nested)
+- removing a field (ideally on major version bump)
+- with extra caution: tweaking field validators; or extending or narrowing an enum range
+
+### Fixed field types
+
+Once a field has been declared, the type of that field must not be changed.
+
+Field types cannot be narrowed, widened or changed entirely.
```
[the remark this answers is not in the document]
dimaqq: I'm adding short explanations. However, I want to keep this scoped to a how-to, not explanation or a spec.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+])
+def test_secret_content(secret_content: dict[str, Any], status):
+    ...
+    state_out = ctx.run(
+        ctx.on.relation_changed(relation=rel, remote_unit=1), state_in)
+
+    assert state_out.unit_status == status
+```
+
+[Full test code](https://github.com/dimaqq/op083-samples/blob/main/test_secret_content.py)
```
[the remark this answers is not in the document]
dimaqq: Nice, I've made https://github.com/canonical/charmlibs/issues/414 for that.

.docs/how-to/design-relation-interfaces.md
dimaqq: I'm not sure where to stash it. At the same time, I wouldn't want to remove this.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+The databag content should be structured to reflect the meaning of data, for example:
+
+```py
+{
+    "direct": {"host": ..., "port": ...},
+    "upstream": {"base_url": ..., "path": ...}
+}
+```
+
+### Secret content schema
```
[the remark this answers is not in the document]
dimaqq: added.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+
+If a field has been removed from the interface, another field with the very same name must not be added.
+
+The exception is reverting removal of a field, where the field is brought back with the exact same type and semantics.
+
+### Collections
+
+Collections must be represented as arrays of objects on the wire, with few exceptions.
+
+Collections must be emitted in some stable order, and the order must be ignored on reception. In other words, collections are sets.
```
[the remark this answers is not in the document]
dimaqq: updated.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,420 @@
...
+
+### No field reuse
+
+If a field has been removed from the interface, another field with the very same name must not be added.
+
+The exception is reverting removal of a field, where the field is brought back with the exact same type and semantics.
+
+### Collections
+
+Collections must be represented as arrays of objects on the wire, with few exceptions.
```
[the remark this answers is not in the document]
dimaqq: I'm rewriting this.

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
dimaqq: I'm not sure I should go into too much detail. I want to keep the how-to prescriptive.

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
dimaqq: expanded.

.docs/how-to/design-relation-interfaces.md
james-garner-canonical: Now that the library related content is moved to a separate page, I think moving this section last would make sense.

.docs/how-to/design-relation-interfaces.md
james-garner-canonical: I've added a daily discussion item for this. In short though, I don't think we should use or recommend `pyright`'s `enableExperimentalFeatures` flag for production targeted content like charm libraries. This means that we probably shouldn't recommend `MISSING`, since it won't type check cleanly without experimental features enabled.

Unfortunately, right now `MISSING` is the only way for `ops.Relation.save` to cleanly omit values (and erase them from the databag if they were present). If this is a key part of the strategy that we want to recommend for managing relation data, then we may need to extend `ops.Relation.save` to support omitting values in other ways (e.g. add an `exclude_none` argument), or recommend an alternative pattern.

.docs/how-to/design-relation-interfaces.md
james-garner-canonical: Are you thinking that the library-behaviour-relateed info is out of scope for this doc and should be covered in the other one?

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
james-garner-canonical: Thanks for explaining. Reading the collections section, I see we don't cover anything about this. Do you think it would be worth explaining the intended behaviour for objects in collections? Reading the doc right now, my impression is that the same rules for fields/sub-fields should apply, and an empty object should always be a valid member of any collection. But is the intention the opposite -- that objects may have required fields, and collections may drop malformed objects while still reporting ready?

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
james-garner-canonical: I think my confusion stems from the order the information in this section is presented. WDYT about framing it like "Top-level fields must not be mandatory -- that is, any and all top-level fields may be absent in the databag, and the databag should still parse cleanly. This implies that the empty databag must have a valid representation in the databag model."

Then we cover the two-to-three different ways of handling missing values: parsing them to a special `MISSING` sentinel that indicates they should be omitted (and cleared) if writing back out, vs parsing them to a default value which may indicate that they're missing (e.g. `None`), or a default value from the expected domain (special enum-like cases only).

The question of reader vs writer setting the default is somewhat overlapping and somewhat orthogonal -- `MISSING` and `None` both imply that the reader determines the default value/behaviour, while writing some other default value to the databag is the opposite. Because we're advocating for `MISSING` elsewhere, I assume that we're recommending reader determined defaults, but I think it would be good to be explicit about this.

I think it's good to be prescriptive here (recommend specifics rather than describe the general possibilities), but I'm hoping we can avoid having gaps (Should the empty databag parse? Should the reader or writer generally determine defaults?).

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
james-garner-canonical: In the meantime, I'm not sure if we should link out to examples like this. WDYT @dwilding? Perhaps OK if it's explicitly temporary. But it would be nice to review the example as part of reviewing this doc.

---

For example, I wonder if this bit is potentially confusing:
```py
    def get_secret_thing(self) -> str:
        if not self.is_ready:
            raise SomeCharmLibError("missing secret uri")
```
Because we're papering over some real pain where `is_ready` and `get_secret_thing` both need to know if the secret URI is available. Or rather, in `get_secret_thing`, we want to know not just if `is_ready`, but why.

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,305 @@
...
+When the interface is modified, running unit tests against both new and old test vectors shows the charm library developer what has been extended and what has been broken.
+The developer then updates the unit tests to encode the deliberate choice for how the old data is meant to be handled.
+
+## Design the relation data format
+
+The only changes allowed on a published interface are:
+
+- adding a new field, at the top level or nested: this is a new feature that must be communicated by a minor version bump of the library.
+- removing a field: this is a backward-incompatible change, and must be clearly communicated by a major version bump of the library.
+- tweaking field validators or extending or narrowing an enumeration value range: must be done with extra care, including compatibility testing between the old and new versions of the library.
```
james-garner-canonical: I'm liking the content in this doc. But this initial section feels a bit scattered right now. As a final step before merging, let's massage this a bit.

- I think it would be a good idea to lead with an overview what's in the document.
- Have an explicit "why take special care with relation data" section (using the existing content).
- Somehow group the miscellaneous feeling bits like unit testing, permitted changes, version bumps. (While some of these are a bit lib related, I think they're fine here, as there is always going to be some overlap between the two.)

WDYT @dwilding?

.docs/how-to/design-relation-interfaces.md
dimaqq: Thank you, let's see what the team says.

dimaqq: To be clear if we wanted to use `MISSING` without `enableExperimentalFeatures`:
- `# type: ignore` needs to be slapped on field declaration
- no changes are needed in comparisons or assignments, thus
- no changes are needed for the runtime code
- no changes are needed in tests

Demo:
https://github.com/dimaqq/op083-samples/pull/1

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
dimaqq: Thanks, this nice and actionable, let me update the doc.

.docs/how-to/design-relation-interfaces.md
dwilding: How about this approach? https://github.com/dimaqq/charmlibs/pull/1

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
dwilding: We should avoid linking to isolated example code in personal forks, so I prefer if we find a way to incorporate this into the Charmlibs repo. Even better if it can be part of a more complete example. But I don't think it needs to block this PR, as long as we resolve it soon after.

.docs/how-to/design-relation-interfaces.md
dimaqq: Stand-up decision: we'll be recommending `| None` at least until `MISSING` is stabilised in both Pydantic and Pyright.

I'll update the PR.

dimaqq: Updated.

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
dimaqq: Done.

.docs/how-to/design-relation-interfaces.md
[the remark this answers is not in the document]
dimaqq: Done.

.docs/how-to/design-relation-interfaces.md
dimaqq: I've rearranged this a bit, and added an example.

.docs/how-to/design-relation-interfaces.md
dimaqq: Looks nice!

dimaqq: Done

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,342 @@
...
+
+Relation data outlives a single charm revision: either side of the relation may be upgraded first, and the upgrade itself is not atomic. The same applies to secret content when a Juju secret is shared over a relation.
+
+When the interface evolves, some version of the library has to support both the old and new schema, and that complexity should not leak into charm code.
+
+## General requirements
+
+### Charm-facing API
+
+The relation data format is a long-lived contract, while the charm-facing API is easier to change. Design the charm-facing API and relation data format separately from the start.
```
james-garner-canonical: Perhaps worth making clear that the charm-facing API is typically/always a charm library.

```suggestion
The relation data format is a long-lived contract, while the charm-facing library API is easier to change. Design the charm-facing API and relation data format separately from the start.
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,342 @@
...
+]}
+
+def test_foos():
+    accepted_foos = charm_lib.parse(DATABAG).foos
+    assert accepted_foos == {"a", "b"}
+```
+
+Collections of primitive types are strongly discouraged.
+
+Data maps are strongly discouraged. An exception to this rule is when the data map key is a Juju entity with a well-known string representation, such as a unit name or machine id:
```
james-garner-canonical: I wonder if it would be helpful to have a do/don't example, followed by the exception + example.

Something I think might be a little confusing is that a JSON object is essentially a data map, and an example showing that we're concerned about sequences with seemingly natural keys would help clear that up.


````suggestion
Data maps are strongly discouraged.

```py
# Do this:
{
    "top-level-key": [{"name": "foo", "n": 1}, {"name": "bar", "n": 2}]
}

# Avoid this:
{
    "top-level-key": {"foo": 1, "bar": 2},
}
```

An exception to this rule is when the data map key is a Juju entity with a well-known string representation, such as a unit name or machine id:
````

**Review by james-garner-canonical — COMMENTED** [review]
This is looking good, thanks! I have two relatively minor suggestions for your consideration.

File: .docs/how-to/design-relation-interfaces.md (hunk 19 above)
Comment (dimaqq) [reply .docs/how-to/design-relation-interfaces.md]: I've updated the examples for collections, ptal.

File: .docs/how-to/design-relation-interfaces.md [hunk 20]
```diff
@@ -0,0 +1,350 @@
...
+When the interface is modified, running unit tests against both new and old test vectors shows the charm library developer what has been extended and what has been broken.
+The developer then updates the unit tests to encode the deliberate choice for how the old data is meant to be handled.
+
+### Allowed interface changes
+
+The only changes allowed on a published interface are:
+
+- adding a new field, at the top level or nested: this is a new feature that must be communicated by a minor version bump of the library.
+- removing a field: this is a backward-incompatible change, and must be clearly communicated by a major version bump of the library.
+- tweaking field validators or extending or narrowing an enumeration value range: must be done with extra care, including compatibility testing between the old and new versions of the library.
```
dwilding: I'd adjust these, as they are full sentences.

```suggestion
- Adding a new field, at the top level or nested: this is a new feature that must be communicated by a minor version bump of the library.
- Removing a field: this is a backward-incompatible change, and must be clearly communicated by a major version bump of the library.
- Tweaking field validators or extending or narrowing an enumeration value range: must be done with extra care, including compatibility testing between the old and new versions of the library.
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+
+Once a field has been declared, the type of that field must not be changed.
+
+Field types cannot be narrowed, widened, or changed entirely, because such a field would fail to validate in either the newer or older application.
+
+The same applies to significant changes to the range of values that a field validator accepts. For example:
+
+- narrowing the set of allowed protocols from any value to HTTP and HTTPS is probably a bug fix, if other protocols such as FTP could not be used by the workload
+- extending an IP address field to accept IPv6 addresses represents a breaking change, because the older remote application is likely to reject the value, potentially making the interface unusable
+- narrowing an IP address field by removing IPv4 addresses represents a breaking change, because if the older remote application sends them, this side of the relation is likely to reject the value, potentially making the interface unusable
```
dwilding: I'd make these full sentences. Plus a few clarifications for easier reading.

```suggestion
- Narrowing the set of allowed protocols from any value to HTTP and HTTPS is probably a bug fix, if other protocols such as FTP could not be used by the workload.
- Extending an IP address field to accept IPv6 addresses represents a breaking change. The older remote application is likely to reject an IPv6 address, potentially making the interface unusable.
- Narrowing an IP address field by removing IPv4 addresses represents a breaking change. If the older remote application sends an IPv4 address, this side of the relation is likely to reject the value, potentially making the interface unusable.
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+
+- is this a base URL, an endpoint, a full URL, an opaque identifier, or an application-specific URI or string
+- limits for the URL as a whole, such as max length or allowed alphabet
+- is the scheme required, optional or not allowed; what schemes are allowed
+- is the userinfo required, optional or not allowed; what elements of userinfo are allowed
+- is the host required, optional or not allowed; what kind of values: domain names, IPv4 addresses, and/or IPv6 addresses
+- is the port required, optional or not allowed; what range of values
+- is the path required, optional or not allowed; any restrictions on the path, such as max length
+- is the query required, optional or not allowed; any restrictions on keys, expected treatment for duplicate keys
+- is the fragment required, optional or not allowed; any restrictions, such as max length
```
dwilding: I'd make these into full-sentence questions.

```suggestion
- Is this a base URL, an endpoint, a full URL, an opaque identifier, or an application-specific URI or string?
- Are there limits for the URL as a whole, such as max length or allowed alphabet?
- Is the scheme required, optional or not allowed; what schemes are allowed?
- Is the userinfo required, optional or not allowed; what elements of userinfo are allowed?
- Is the host required, optional or not allowed; what kind of values: domain names, IPv4 addresses, and/or IPv6 addresses?
- Is the port required, optional or not allowed; what range of values?
- Is the path required, optional or not allowed; any restrictions on the path, such as max length?
- Is the query required, optional or not allowed; any restrictions on keys, expected treatment for duplicate keys?
- Is the fragment required, optional or not allowed; any restrictions, such as max length?
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+        Data.model_validate({"number": bad_value})
+```
+
+### No mandatory fields
+
+Top-level fields must not be mandatory. Any and all top-level fields may be absent in the relation data, and it must still parse cleanly.
+
+This implies that the empty relation data must have a valid representation in the interface model.
+In practice, use a single model for both reading and writing.
+That means the writer will naturally materialise the model's default values when writing the relation data.
```
dwilding: Perhaps more easily understood?

```suggestion
That means the writer will use the model's default values when writing the relation data.
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+@pytest.mark.parametrize("field_to_remove", ["name", "surname"])
+def test_missing_fields(field_to_remove):
+    data = {**V1_DATABAG}
+    del data[field_to_remove]
+    assert DataV2.model_validate(data)
+```
+
+### No field reuse
+
+If a field has been removed from the interface, another field with the very same name must not be added. This rule exists to make field removal possible without the risk of misinterpretation when two applications from different eras are integrated.
```
dwilding: This would be fine I think

```suggestion
If a field has been removed from the interface, another field with the same name must not be added. This rule exists to make field removal possible without the risk of misinterpretation when two applications from different eras are integrated.
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+
+The exception is reverting removal of a field, where the field is brought back with the exact same type and semantics.
```
dwilding:
```suggestion
The exception is reverting removal of a field, where the field is brought back with identical type and semantics.
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+rel = testing.Relation('endpoint', remote_app_data=data)
+state_in = testing.State(leader=True, relations={rel})
+ctx.run(ctx.on.relation_changed(rel), state_in)
+```
+
+### Collections
+
+Collections must be represented as arrays of objects on the wire when using the default JSON serialisation.
+
+Collections must be emitted in a stable order, so that setting the same data does not trigger interface bounce. The order must be ignored on reception, and the recipient is expected to discard duplicates. In other words, collections are sets.
```
dwilding: I'm not sure what "interface bounce" means? Is there a brief way to say that this means? Or adjust the wording?

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+
+
+# This is preferred
+SAMPLE_DATABAG = {"endpoints": [
+    {"id": "foo", "some_url": "//foo-path"},
+    {"id": "bar", "some_url": "//bar-path"},
+]}
+```
+
+The definition must be accompanied by a unit test, which may look as follows. Note that including a custom validator requires a comprehensive set of unit tests.
```
dwilding: The second sentence feels like a more general point. I'd move it, remove it, or work it into the first sentence:


```suggestion
The definition must be accompanied by comprehensive unit tests. A unit test may look like:
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
...
+]}
+
+def test_foos():
+    accepted_foos = charm_lib.parse(DATABAG).foos
+    assert accepted_foos == {"a", "b"}
+```
+
+Collections of primitive types are strongly discouraged, because they are impossible to extend.
+
+Data maps are strongly discouraged. An exception to this rule is when the data map key is a Juju entity with a well-known string representation, such as a unit name or machine id:
```
dwilding:
```suggestion
Data maps are strongly discouraged. An exception to this rule is when the data map key is a Juju entity with a well-known string representation, such as a unit name or machine ID:
```

.docs/how-to/design-relation-interfaces.md
```diff
@@ -0,0 +1,350 @@
+(design-relation-interfaces)=
```
dwilding, line 7: Let's include a description. Something like:

```suggestion
---
myst:
  html_meta:
    description: Follow the design rules for relation interfaces between charms, so that interfaces can evolve without causing breaking changes or downtime. 
---

(design-relation-interfaces)=
```

dwilding (approved):
Good to go after a few minor tweaks 🙂

.docs/how-to/design-relation-interfaces.md
dimaqq: updated to "doesn't trigger spurious relation-changed events", thanks.

james-garner-canonical (approved):
Looks good, thanks!

PR #411, opened 2026-04-10
feat: add modelerror check when loading relation data

### LINKED ISSUE #298: Error on loading databag when relation available but databag isn't
When trying to load relation data, it is possible to have the relation be available but the databag unreadable. The lib is running into an error when loading the databag because it doesn't handle the case where relation is available but the databag contents aren't.

``` 
   def _load_provider_certificates(self, relation: Relation) -> List[_Certificate]:
        try:
            provider_relation_data = _ProviderApplicationData.load(relation.data[self.charm.app])
        except DataValidationError:
            logger.debug("Invalid provider relation data")
            return []
        return copy.deepcopy(provider_relation_data.certificates)
```

`relation.data[self.charm.app]` is breaking in the above.

kayra1:
This was causing an issue with self-signed-certificates [here](https://github.com/canonical/self-signed-certificates-operator/issues/530).

When we check for relations, it's possible that a relation is available but the databag is not accessible. This is possible sometimes during relation-broken or relation-departed events. In those cases we receive a model error, which we now capture to avoid crashing charms unnecessarily. 

fixes #298


james-garner-canonical:
Please make the `pyproject.toml` changes in a separate PR -- I think they're a good idea. In the meantime, `just format` and `just lint` use `--preview` (or you can pass `--preview` yourself if you're running `ruff` separately).

EDIT: And we can probably drop `--preview` in the `justfile` if we add this to `pyproject.toml`, which is a nice bonus.

EDIT2: I see the rationale for the change is for LSPs, which makes sense.

kayra1:
> Please make the `pyproject.toml` changes in a separate PR -- I think they're a good idea. In the meantime, `just format` and `just lint` use `--preview` (or you can pass `--preview` yourself if you're running `ruff` separately).
> 
> EDIT: And we can probably drop `--preview` in the `justfile` if we add this to `pyproject.toml`, which is a nice bonus.
> 
> EDIT2: I see the rationale for the change is for LSPs, which makes sense.

Hello, yes I was going to reach out. adding preview has also surfaced another problem, where ruff's preview behavior seems to have changed between 0.11 which is specified in the project and 0.15 which my editor uses. Should I manually pin my editor's ruff version to 0.11 or do you think it's worth updating to the most recent version or dropping the preview all together? We could also always just leave it.

james-garner-canonical:
> > Please make the `pyproject.toml` changes in a separate PR -- I think they're a good idea. In the meantime, `just format` and `just lint` use `--preview` (or you can pass `--preview` yourself if you're running `ruff` separately).
> > 
> > EDIT: And we can probably drop `--preview` in the `justfile` if we add this to `pyproject.toml`, which is a nice bonus.
> > 
> > EDIT2: I see the rationale for the change is for LSPs, which makes sense.
> 
> Hello, yes I was going to reach out. adding preview has also surfaced another problem, where ruff's preview behavior seems to have changed between 0.11 which is specified in the project and 0.15 which my editor uses. Should I manually pin my editor's ruff version to 0.11 or do you think it's worth updating to the most recent version or dropping the preview all together? We could also always just leave it.

I suggest pinning your editor's version for now, but I'm open to bumping the project version or dropped preview depending what formatting changes either entails. If you want to try either in a separate PR, please do, and/or I'll experiment next week.

kayra1:
@james-garner-canonical just opened the other PR

saltiyazan (approved):

PR #412, opened 2026-04-10
feat: add --preview to pyproject.toml

kayra1:
Following our conversation with @james-garner-canonical in #411 

This change adds the preview = true flag to the pyproject.toml file, removing the need to pass it as a cli variable in the just file. This lets LSP's also automatically pick up that configuration and run it while formatting on save.

One issue with that remains is some LSP's don't respect the ruff version declaration in the configuration, which may produce a different output depending on which one is used. The project uses 0.11, and mine coincidentally was using 0.15, which produced inconsistent outputs.

The solutions to that are:
* pinning the editor's ruff version to 0.11 
* updating to the most recent version 
* dropping the preview all together

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,7 @@ exclude_also = [
 # Linting tools configuration
 [tool.ruff]
 line-length = 99
+required-version = "==0.11.0"
 target-version = "py310"
 exclude = [
     # These are the default values of ruff's exclude
@@ -64,9 +65,11 @@ exclude = [
 
 # Ruff formatter configuration
 [tool.ruff.format]
+preview = true
 quote-style = "single"
 
 [tool.ruff.lint]
+preview = true
 select = [
     # flake8-builtins
     "A",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,6 @@ exclude_also = [
 # Linting tools configuration
 [tool.ruff]
 line-length = 99
-required-version = "==0.11.0"
 target-version = "py310"
 exclude = [
     # These are the default values of ruff's exclude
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -28,6 +28,7 @@ exclude_also = [
 # Linting tools configuration
 [tool.ruff]
 line-length = 99
+required-version = "==0.11.0"
```
james-garner-canonical: I didn't know this was an option, neat. It seems odd to specify the version twice though, and in the same file (it's specified in the `fast-lint` dependency group). Since not all tools will actually respect this, as you note, let's drop this line for now.

```suggestion
```

james-garner-canonical (commented):
Thanks for taking care of this. Since this should make the `--preview` flags in the `justfile` redundant, could you remove them in this PR too?

james-garner-canonical:
> One issue with that remains is some LSP's don't respect the ruff version declaration in the configuration, which may produce a different output depending on which one is used. The project uses 0.11, and mine coincidentally was using 0.15, which produced inconsistent outputs.
> 
> The solutions to that are:
> 
> * pinning the editor's ruff version to 0.11
> * updating to the most recent version
> * dropping the preview all together

I think the right approach is to pin the editor's version, which hopefully isn't too painful to do on a per-project basis. I'm happy to consider updating the repo-wide `ruff` version separately though (need to see what changes would be introduced by doing so).

kayra1:
@copilot please remove the --preview flags from ruff commands in the justfiles in the repo

james-garner-canonical (approved):
Thanks for taking care of this.

pyproject.toml
kayra1: @james-garner-canonical this is probably to hint at the LSP to use the rules in this specific version, because otherwise LSP's would have to make exceptions for which programs' versions they'd have to passthrough. Maybe it's "=0.11.0" that would have worked but I haven't investigated.

PR #417, opened 2026-04-13
ci: drop markdownlint, following our other projects

james-garner-canonical:
This PR drops the markdown linting action, as it doesn't add much value -- it never fails, and seemingly only lints `README.md` etc, not actual docs source files.


dwilding (approved):
I agree. We decided not to use this in our other projects too.

james-garner-canonical:
The CI failure is unrelated to this PR, but needs investigating separately.

PR #419, opened 2026-04-13
feat: add etcd lock and single API orchestration

patriciareinoso:
## Summary
 
This PR introduces the etcd sync and async locking mechanism for rolling ops together with a single API that fallbacks to the peer-relation solution in case of etcd failure or unavailability.

### etcd async lock
- An background process is spawn on lock request
- It performs the lock acquisition and triggers a lock_granted juju custom hook
- The charm observes this event and executes the corresponding operation and indicates when it finished using etcd keys.

### etcd sync lock
- Implemented a distributed lock using etcd (lease + transactional lock key).
-  We use the <unit-name>:sync as the owner of the lock to differentiate from the lock trying to be acquired in the background process. Meaning that the sync lock may take priority over operations on the async lock.
- Lock acquisition uses retries (via tenacity) on etcdctl commands.
- The lock is tied to a lease and automatically released if the lease expires.

### common API
- The `RollingOpsManager` is the public API for advanced rolling ops
- It decides on which backend the operation will run: etcd or the peer-relation
- If etcd operations fail (e.g. etcdctl errors, lease issues), the system falls back to the peer-relation-based implementation.
- This ensures operations can continue even when etcd is not usable.
- If there is any failure on the etcd background process a juju custom hook is trigger, this way the RollingOpsManager knows it needs to fallback to peer-relations
- Every operation request is:
  - written to etcd
  - also duplicated into the peer relation databag
- After execution, the operation state is updated in both places.
- This ensures:
  - no operations are lost during fallback
  - scheduling can continue seamlessly regardless of backend
- The system only switches back to etcd when the operation queue is empty
- Before resuming etcd usage any remaining etcd queue state is cleaned up. This guarantees a clean slate and avoids inconsistencies

## Peer sync lock 
The `SyncLockBackend` defines the interface that any charm author would need to implement in order to use sync locking in the context of the peer-relation solution given that using the peer-solutions we are not able to guarantee mutual exclusion when tearing down.


rollingops/src/charmlibs/rollingops/common/_base_worker.py
```diff
@@ -0,0 +1,273 @@
...
+        raise NotImplementedError
+
+    def _on_existing_worker(self, pid: int) -> bool:
+        """Handle case where a worker is already running.
+
+        Returns:
+            True if a new worker should be started,
+            False if start() should return early.
+        """
+        raise NotImplementedError
```
Gu1nness: Looks like this could be a property ?

rollingops/src/charmlibs/rollingops/common/_base_worker.py
```diff
@@ -0,0 +1,273 @@
...
+                *self._worker_args(),
+            ],
+            cwd=str(self._charm_dir),
+            stdout=log_out,
+            stderr=log_out,
+            env=env,
+        ).pid
+
+        self._set_pid_str(str(pid))
+        logger.info('Started %s process with PID %s', self._handle_name, pid)
```
Gu1nness: Why no context manager here for the log_out file ?

rollingops/src/charmlibs/rollingops/common/_base_worker.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def stop(self) -> None:
```
Gu1nness, line 231: Comment on this function  and what I think: I'd rather have the PID always to be an int, set to a default value if it doesn't exist: 0 or -1 are never valid PIDs so you could use them, reducing the number of checks, and reducing the risk of `None vs ''` in comparisons

rollingops/src/charmlibs/rollingops/common/_exceptions.py
```diff
@@ -0,0 +1,75 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Exceptions used in rollingops."""
+
+
+class RollingOpsError(Exception):
```
Gu1nness, line 18: Note for the future if we want to improve all of our error frameworks: Each error should have a unique error code and a message so that we can log it and it would improve tracing and monitoring.

No need to do it now, just food for thoughts for the future.

Idea of implementation:
```python3
from pydantic import BaseModel, ConfigDict
from dataclasses import dataclass, field
class CommonExceptionModel(BaseModel):
	# Model config
	model_config = ConfigDict(from_attributes=True)

	# Attributes
    code: int
    message: str
    name: str


@dataclass
class CommonException(Exception):
    message: str
    code: int = 0
    name: str = field(init=False)

    def __post_init__(self):
        try:
            self.name = self.__class__.__name__
        except AttributeError:
			# For inheritance scenarios.
            pass
        super().__init__(self.message)

    def __str__(self):
        return str(self.message)

    def serialize(self):
            return CommonExceptionModel.model_validate(self).model_dump_json()
```

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+    RollingOpsDecodingError,
+    RollingOpsNoRelationError,
+)
+from charmlibs.rollingops.common._utils import datetime_to_str, now_timestamp, parse_timestamp
+
+logger = logging.getLogger(__name__)
+
+
+class OperationResult(StrEnum):
+    """Callback return values."""
```
Gu1nness: Please extend this documentation to explain each value (according to spec)
1. `release`: means that we have succeeded and we release the lock I guess
2. `retry-release`: means that execution failed or timed out, but wants to release the lock so other units can proceed, but we want to retry later, right ? Or we don't retry ?
3. `retry-hold`: means that execution failed or timed out, and wants to run the operation again, but without releasing the lock so that no other operation can run in between ? (for more critical paths I guess ?)

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+
+
+class RollingOpsStatus(StrEnum):
+    """High-level rolling-ops state for a unit.
+
+    This status reflects whether the unit is currently executing, waiting,
+    or idle with respect to rolling operations.
+    """
+
+    INVALID = 'invalid'
```
Gu1nness: What does this status mean ? Is it a fallback if we fail to get a valid status for a unit ?

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+    WAITING = 'waiting'
+    GRANTED = 'granted'
+    IDLE = 'idle'
+
+
+@dataclass
```
Gu1nness:
```suggestion
@dataclass(frozen=True)
```

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+    op_id: str | None = None
+    result: OperationResult | None = None
+
+
+@dataclass
+class BackendState:
+    """Unit-scoped backend ownership and recovery state."""
+
+    processing_backend: str = ProcessingBackend.PEER
+    etcd_cleanup_needed: str = 'false'
```
Gu1nness, line 140: Could this be a boolean or not ?

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+    status: RunWithLockStatus
+    op_id: str | None = None
+    result: OperationResult | None = None
+
+
+@dataclass
+class BackendState:
+    """Unit-scoped backend ownership and recovery state."""
+
+    processing_backend: str = ProcessingBackend.PEER
```
Gu1nness, line 139: Could this be of type `ProcessingBackend` or not ?

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+    def __init__(self, model: Model, relation_name: str, unit: Unit):
+        relation = model.get_relation(relation_name)
+        if relation is None:
+            raise RollingOpsNoRelationError()
+
+        self._relation = relation
+        self.unit = unit
+
+    def _load(self) -> BackendState:
+        return self._relation.load(BackendState, self.unit, decoder=lambda s: s)
```
Gu1nness: Seeing this the way it is, it looks like BackendState would benefit from being a pydantic BaseModel so that it could have better typing and you would not need to provide a decoder.
This would probably give more trust in the values that it can hold.

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+
+    def __init__(self, model: Model, relation_name: str, unit: Unit):
+        relation = model.get_relation(relation_name)
+        if relation is None:
+            raise RollingOpsNoRelationError()
+
+        self._relation = relation
+        self.unit = unit
+
+    def _load(self) -> BackendState:
```
Gu1nness: To prevent having to load on every call in the method, maybe it makes sens that it is loaded once when initializing the class ?
Then your methods could still have the same logic, but without the `self._load()` everytime your do it?
eg:
```python3
class UnitBackendState:
    """Manage backend ownership and fallback state for one unit queue."""

    def __init__(self, model: Model, relation_name: str, unit: Unit):
        relation = model.get_relation(relation_name)
        if relation is None:
            raise RollingOpsNoRelationError()

        self._relation = relation
        self.unit = unit
		self._backend_state = self._relation.load(BackendState, self.unit, decoder=lambda s: s)

	@property
    def backend(self) -> ProcessingBackend:
        """Return which backend owns execution for this unit's queue."""
        return self._backend_state.backend

    def fallback_to_peer(self) -> None:
        """Switch this unit's queue to peer processing and mark etcd cleanup needed."""
        self._backend_state.backend = ProcessingBackend.PEER
        self._backend_state.cleanup_needed = True
        self._save(data)
```

One step further would be to save only once at the end of your hook to remove useless writes (even if they are quite cost effective for now).

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+            self.result = OperationResult.RETRY_RELEASE
+
+    def retry_hold(self) -> None:
+        """Mark the operation for retry if it has not reached the max retry."""
+        self.increase_attempt()
+        if self.is_max_retry_reached():
+            self.result = OperationResult.RELEASE
+            logger.warning('Operation max retry reached. Dropping.')
+        else:
+            self.result = OperationResult.RETRY_HOLD
```
Gu1nness, line 341: Docs are the same, maybe they should be explicit about the difference between both scenarios.

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+                'Failed to deserialize data to create an OperationQueue.'
+            ) from e
+        if not isinstance(items, list) or not all(isinstance(s, str) for s in items):  # type: ignore[reportUnknownVariableType]
+            raise RollingOpsDecodingError(
+                'OperationQueue string must decode to a JSON list of strings.'
+            )
+
+        operations = [Operation.from_string(s) for s in items]  # type: ignore[reportUnknownVariableType]
+        return cls(operations)
+
```
Gu1nness, line 439: I don't understand why you're working around json and your classes here.
You could have `to_string` that does `return json.dumps([item.to_dict() for item in self.operations], …)`
And then you only need to load your json once and `return cls(Operation.from_dict(s) for s in items` in the `from_string` method.
It also looks like all this would benefit from a better data modelling so that all those quite complex serialization/deserialization methods would not need existing.

It's not a strong blocker so you can keep it that way as we're under time constraints, but I fear that it will just make it hard to maintain and improve in the future, increasing the risk of failures as we have kind of DIYed the serialization/deserialization process.

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+        """Acquire the lock, blocking until it is granted or timeout expires.
+
+        Args:
+            timeout: Maximum time in seconds to wait for the lock.
+                None means wait indefinitely.
+
+        Raises:
+            TimeoutError: If the lock could not be acquired within the timeout.
+        """
+        pass
```
Gu1nness:
```suggestion
        raise NotImplementedError
```

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -0,0 +1,456 @@
...
+
+    @abstractmethod
+    def release(self) -> None:
+        """Release the lock.
+
+        Implementations must ensure that only the lock owner can release
+        the lock and that any associated resources are cleaned up.
+        """
+        pass
```
Gu1nness:
```suggestion
        raise NotImplementedError
```

rollingops/src/charmlibs/rollingops/common/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+def parse_timestamp(timestamp: str) -> datetime | None:
+    """Parse epoch timestamp string. Return None on errors."""
+    try:
+        return datetime.fromtimestamp(float(timestamp), tz=UTC)
+    except Exception:
+        return None
+
+
+def datetime_to_str(datetime: datetime) -> str:
+    return str(datetime.timestamp())
```
Gu1nness: Nit: reduce confusion with:
```suggestion
def datetime_to_str(dt: datetime) -> str:
    return str(dt.timestamp())
```

rollingops/src/charmlibs/rollingops/common/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+    This functions is used in the context of the background process.
+
+    Args:
+        log_file: Path to the log file where logs should be written.
+    """
+    handler = RotatingFileHandler(
+        log_file,
+        maxBytes=10 * 1024 * 1024,  # 10 MB
+        backupCount=3,
```
Gu1nness:
```suggestion
        backupCount=10,
```

rollingops/src/charmlibs/rollingops/common/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+def datetime_to_str(datetime: datetime) -> str:
+    return str(datetime.timestamp())
+
+
+def setup_logging(log_file: str) -> None:
+    """Configure logging with file rotation.
+
+    This sets up the root logger to write INFO-level (and above) logs
+    to a rotating file handler. Log files are capped at 10 MB each,
+    with up to 3 backup files retained.
```
Gu1nness: I would keep a bit more than 3 backup files just to be sure.
Costs almost nothing could help a lot in the future if this gets very noisy.
Do we have an idea of the volume of log emitted in size for an hour of operation where everything goes bad ? (meaning maximal generation of logs due to many locks all over the place).
```suggestion
    with up to 10 backup files retained.
```

rollingops/src/charmlibs/rollingops/common/_utils.py
```diff
@@ -0,0 +1,120 @@
...
+    """Dispatch the 'rollingops_lock_granted' hook on a unit.
+
+    Args:
+        unit_name: The Juju unit name (e.g., "app/0").
+        charm_dir: Filesystem path to the charm directory.
+
+    Raises:
+        subprocess.CalledProcessError: If the hook execution fails.
+    """
+    hook_name = 'rollingops_lock_granted'
```
Gu1nness: Make that a constant at top level maybe?

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -0,0 +1,389 @@
...
+        )
+
+        self.etcd = EtcdRequiresV1(
+            charm,
+            relation_name=etcd_relation_name,
+            cluster_id=self.keys.cluster_prefix,
+            shared_certificates=self.shared_certificates,
+        )
+
+        self.keys = RollingOpsKeys.for_owner(cluster_id=cluster_id, owner=owner)
```
Gu1nness: Duplicated, already on line 89.
```suggestion
```

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -0,0 +1,389 @@
...
+        self._async_lock = EtcdLock(lock_key=self.keys.lock_key, owner=owner)
+        self._sync_lock = EtcdLock(lock_key=self.keys.lock_key, owner=f'{owner}:sync')
+        self.operations = ManagerOperationStore(self.keys, owner)
```
Gu1nness: Nit: rename to `self.operations_store` for consistency with the object it holds.

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -0,0 +1,389 @@
...
+    Each unit manages its own etcd worker process and operation queues.
+    Operations are scoped using a cluster identifier and a per-unit owner.
+    """
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        peer_relation_name: str,
+        etcd_relation_name: str,
+        cluster_id: str,
```
Gu1nness, line 64: What happens if cluster_id is an empty string ? Does all behave correctly still ?

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -0,0 +1,389 @@
...
+        self.worker.start()
+
+    def _on_etcd_relation_created(self, event: RelationCreatedEvent) -> None:
+        """Validate that the etcdctl command is available when etcd is related.
+
+        Args:
+            event: The relation-created event for the etcd relation.
+        """
+        if not etcdctl.is_etcdctl_installed():
+            logger.error('%s is not installed.', etcdctl.ETCDCTL_CMD)
```
Gu1nness, line 190: Shouldn't we raise in that case ? Or it is because we'll allow ourselves to fallback on the peer relation ?

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -0,0 +1,389 @@
...
+            relation_name=etcd_relation_name,
+            cluster_id=self.keys.cluster_prefix,
+            shared_certificates=self.shared_certificates,
+        )
+
+        self.keys = RollingOpsKeys.for_owner(cluster_id=cluster_id, owner=owner)
+        self._async_lock = EtcdLock(lock_key=self.keys.lock_key, owner=owner)
+        self._sync_lock = EtcdLock(lock_key=self.keys.lock_key, owner=f'{owner}:sync')
+        self.operations = ManagerOperationStore(self.keys, owner)
+        self._lease = None
```
Gu1nness: Nit: type hints for this attribute so we now what it means from the start of the function.

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -0,0 +1,389 @@
...
+        """
+        self._lease = EtcdLease()
+
+        deadline = None if timeout is None else time.monotonic() + timeout
+
+        try:
+            self._lease.grant()
+            while True:
+                try:
+                    if self._sync_lock.try_acquire(self._lease.id):  # type: ignore[reportArgumentType]
```
Gu1nness: This warning makes me wonder if the checks are done correctly here;
Maybe we should first check that `self._lease.id is not None` before trying to acquire.
This would reduce the obscure path where we try to acquire, we raise an error in the `try_acquire` and catch it here.
I'd rather reduce amount of type exceptions if possible.

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -0,0 +1,389 @@
...
+        unit's queued operation state.
+
+        Returned values:
+            - INVALID: the peer or etcd relation is missing
+            - GRANTED: the async lock is currently held by this unit
+            - WAITING: this unit has queued work but does not hold the lock
+            - IDLE: this unit has no pending work
+
+        Returns:
+            The current rolling-ops status for this unit.
```
Gu1nness, line 389: This method can also raise if `etcdctl.ensure_initialized()` raises.

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -0,0 +1,488 @@
...
+class EtcdLease:
+    """Manage the lifecycle of an etcd lease and its keep-alive process."""
+
+    def __init__(self):
+        self.id: str | None = None
+        self.keepalive_proc: subprocess.Popen[str] | None = None
+
+    def grant(self) -> None:
+        """Create a new lease and start the keep-alive process."""
+        res = etcdctl.run('lease', 'grant', str(LOCK_LEASE_TTL))
```
Gu1nness: Given that we use it only once in this method, it looks like `LOCK_LEASE_TTL` could be a string from the start.

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -0,0 +1,488 @@
...
+        # parse: "lease 694d9c9aeca3422a granted with TTL(1800s)"
```
Gu1nness: Nit: Respect the `LOCK_LEASE_TTL` value here:
```suggestion
        # parse: "lease 694d9c9aeca3422a granted with TTL(60s)"
```

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -0,0 +1,488 @@
...
+        parts = res.split()
+        self.id = parts[1]
```
Gu1nness, line 55: What happens if res is not in that format ? Would we pick a random string a lease id ?
If we know the format of the string, maybe it's worth using a basic regexp and raise if we're unable to parse it to get the lease id.

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -0,0 +1,488 @@
...
+        if lease_id is None:
+            logger.info('Lease ID is None. Keepalive for this lease cannot be started.')
+            return
+        etcdctl.ensure_initialized()
+        self.keepalive_proc = subprocess.Popen(
+            [etcdctl.ETCDCTL_CMD, 'lease', 'keep-alive', lease_id],
+            env=etcdctl.load_env(),
+            stdin=subprocess.PIPE,
+            stdout=subprocess.DEVNULL,
+            stderr=subprocess.DEVNULL,
```
Gu1nness: I think we should not dump the logs here, rather write them in the logging file so we have some tracking of what happens. We could even use that as audit logging.

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -0,0 +1,488 @@
...
+
+    The lock is attached to an etcd lease so that it is
+    automatically released if the owner stops refreshing the lease.
+    """
+
+    def __init__(self, lock_key: str, owner: str):
+        self.lock_key = lock_key
+        self.owner = owner
+
+    def try_acquire(self, lease_id: str) -> bool:
```
Gu1nness, line 161: For this function and the two following: Validation of the three values to ensure they are not empty ? Otherwise how would that call behave ?

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -0,0 +1,488 @@
...
+            True if the operation was moved successfully, otherwise False.
+        """
+        kv = etcdctl.get_first_key_value_pair(self.prefix)
+        if kv is None:
+            return False
+
+        op_id = kv.key.split('/')[-1]
+        new_key = f'{to_queue_prefix}{op_id}'
+        data = json.dumps(kv.value)
+        value_escaped = data.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
```
Gu1nness: NGL: this looks like a recipe for disaster. It looks like we're hacking around the escapes for the subprocess call that will happen in `etcdctl.txn`.
Probably there's a better way to do this?

rollingops/src/charmlibs/rollingops/etcd/_etcdctl.py
```diff
@@ -195,21 +288,106 @@ def run(*args: str) -> str | None:
...
+    Returns:
+        True if the transaction succeeded, otherwise False.
+
+    Raises:
+        RollingOpsEtcdNotConfiguredError: if etcdctl is not configured.
+        PebbleConnectionError: if the remote container cannot be reached.
+        RollingOpsEtcdctlError: etcdctl command error.
+    """
+    res = _run_checked('txn', cmd_input=txn_input)
+    return 'SUCCESS' in res.stdout
```
Gu1nness: QQ: what if the string `SUCCESS` ends up in the `txn_input` string ? Would this check this mean something ?
Shouldn't we check the returncode of res ? 
Also: reading `_run_checked`, I feel like we should be using a try/catch:
```python3
try:
	res = _run_checked(…) # This does not raise only on success from what I understand ?
	return True
except …:
	return False

File: rollingops/src/charmlibs/rollingops/etcd/_rollingops.py [hunk 34]
```diff
@@ -0,0 +1,164 @@
...
+    The worker always attempts to revoke its lease and release the lock
+    before exiting.
+    """
+    setup_logging('/var/log/etcd_rollingops_worker.log')
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--unit-name', required=True)
+    parser.add_argument('--charm-dir', required=True)
+    parser.add_argument('--owner', required=True)
+    parser.add_argument('--cluster-id', required=True)
```
Gu1nness: Nit: Add types and action to all arguments to ensure they behave properly.

rollingops/src/charmlibs/rollingops/etcd/_rollingops.py
```diff
@@ -0,0 +1,164 @@
...
+
+            if not operations.has_pending():
+                time.sleep(NEXT_OP_SLEEP)
+                continue
+
+            if not lock.is_held():
+                if lease.id is None:
+                    lease.grant()
+
+                logger.info('Try to get lock.')
```
Gu1nness: Nit: For all following logs, improve logging with context (eg: lease id, cluster_id, owner, etc).
Or better, have them added automatically in the setup logging function by providing some extra args/kwargs and a formatter that uses those values.

rollingops/src/charmlibs/rollingops/etcd/_rollingops.py
```diff
@@ -0,0 +1,164 @@
...
+
+    This notifies the charm that the etcd worker encountered an
+    unrecoverable error so that higher-level logic can fall back to the
+    peer backend.
+
+    Args:
+        unit_name: Name of the unit dispatching the hook.
+        charm_dir: Path to the charm root directory.
+    """
+    hook_name = 'rollingops_etcd_failed'
```
Gu1nness: Nit: bring to a constant at top level.
Maybe at some point we could even build a collection / enum of all hook names so they are centralised somewhere.

rollingops/src/charmlibs/rollingops/peer/_worker.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Args:
+            pid: The PID of the currently running worker.
+
+        Returns:
+            True to indicate that the existing worker was handled and a new
+            worker can be started.
+        """
+        logger.info('Stopping existing RollingOps worker PID %s before restart.', pid)
+        self.stop()
```
Gu1nness, line 110: Maybe we should catch the possible errors on stop ? Or there are none ?

rollingops/tests/integration/charms/actions.yaml
```diff
@@ -30,3 +30,15 @@ deferred-restart:
...
+
+sync-restart:
+  description: Example restart with a custom callback function. Used in testing
+  params:
+    delay:
+      description: "Introduce an artificial delay (for testing)."
+      type: integer
+      default: 0
+    timeout:
+      description: "Time (seconds) to wait before given up."
```
Gu1nness:
```suggestion
      description: "Time (seconds) to wait before giving up."
```

Gu1nness (commented):
Long review again, but I really love the way this is coming up!
It's mostly nitpicking and improving a bit the logging/stability.
Notes on the data modelling but I don't think we have time to rewrite the whole data modelling so I guess we'll go with that for now? Maybe except the `from_string`, `to_string` that really looks like hacking around json/poorly using json.

Nearly there!

rollingops/src/charmlibs/rollingops/etcd/_rollingops.py
patriciareinoso: now it's a bit more centralized. And I added constants that are used also i the charm event definition so we avoid possible mismatches

rollingops/src/charmlibs/rollingops/etcd/_rollingops.py
patriciareinoso: Added type and help. Action defaults to `store` and I think is enough in these cases

rollingops/src/charmlibs/rollingops/common/_models.py
patriciareinoso: I changed it to UNAVAILABLE

In the case of peer: it means the peer relation does not exist
In the case of etcd: it means , peer relation or etcd relation do not exist , or etcd i not reachable.

But in fact, from the external API, if etcd is UNAVAILABLE we fall back to peer and return the status on the peer. 

I improved this docstring and the one from RollingOpsState

rollingops/src/charmlibs/rollingops/peer/_worker.py
patriciareinoso: All the errors are caught inside `stop`

rollingops/src/charmlibs/rollingops/etcd/_backend.py
patriciareinoso: In this case we will fallback.
But thinking about it again, If you integrated with etcd it's because you meant to use etcd and you should fallback only because etcd is unavaible, not because you failed to install the etcdctl. Maybe this should be more noisy?

Gu1nness:
@patriciareinoso About the data modelling comment, and after discussion with Mehdi I paste here a more detailed input:

I think we're hacking around json and building a lot of complexity by re-inventig data serialization/deserialization instead of using proper tools to do that.

An example that we have is:
```python3
@dataclass
class Operation:
    …
    def from_string(cls, data: str) -> Operation:
        obj = json.loads(data)
        return cls.from_dict(obj)

    def to_string() -> str:
        return json.dumps(self._to_dict(), separators=(',', ':'))
       

class OperationQueue: # Beware, not event a dataclass
    def __init__(self, operations: list[Operation] | None = None):
        self.operations: list[Operation] = list(operations or [])

    def to_string(self):
        items = [op.to_string() for op in self.operations]
        return json.dumps(items, separators=(',', ':'))

   def from_string(cls, data: str):
        try:
            items = json.loads(data)
        except:
            …
        if not isinstance(items, list) or not all(isinstance(s, str) for s in items):
            raise …
        operations = [Operation.from_string(s) for s in items]
        return cls(operations)
```
And that builds something weird.

Instead of having:
```
[
    {"callback_id": "...", "requested_at": "…",}, …
]
```

We have:
```
[
    "{\"callback_id\": \"…\", \"requested_at\": \"…\"}", …
]
```
This implies:
 * two json deser instead of one
 * Non standard format
 * Custom parsing
 * Risks in the future if all those nested quotes become re-escaped by something downhill
 
Possibilities to improve this:
1. Improve the from_string and to_string so they are are less complex, using the power of json and dataclasses (aka build proper json of nested structures).
2. (better IMHO)Use dedicated library for data serialization and deserialization that automatically does the parsing and validation (because I haven't even talked about the manual validation that we end up doing with this implem). An idea could be pydantic which is already used in the background by Data interfaces V1 on which you rely already.

rollingops/src/charmlibs/rollingops/etcd/_etcdctl.py
patriciareinoso: In case of success:
```
CompletedProcess(args=['etcdctl', 'txn'], returncode=0, stdout='SUCCESS\n\nOK\n\n1\n', stderr='')
```
in case of failure
```
CompletedProcess(args=['etcdctl', 'txn'], returncode=0, stdout='FAILURE\n', stderr='')
```
so there was no risk of of having `SUCCESS` in the input string.

If the transaction fails, the return code would be positive so we check the stdout to decide. 

If the process fails (time out, etcd unreachable ... ) the exception is bubbled up

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
patriciareinoso: fixed after using pydanctic for operations

rollingops/src/charmlibs/rollingops/etcd/_backend.py
patriciareinoso: the etcd keys would look like this 
`/rollingops/default//model-unit/`

etcd can handle that, it's not ideal but it operations won't fail

rollingops/src/charmlibs/rollingops/common/_models.py
patriciareinoso: fixed after using pydantic

rollingops/src/charmlibs/rollingops/common/_models.py
patriciareinoso: ensure that we only load at construction.

rollingops/src/charmlibs/rollingops/common/_models.py
patriciareinoso: Ack.
I prioritized Operation and OperationQueue since the serialization was production incorrect json values. This can be fixed in another PR

rollingops/src/charmlibs/rollingops/common/_models.py
patriciareinoso: To be revisited if we move forward making this class pydantic

rollingops/src/charmlibs/rollingops/common/_exceptions.py
patriciareinoso: Nice, it could be implemented later

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
patriciareinoso: added a validation but not using regex. wdyt?

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
patriciareinoso: added

rollingops/src/charmlibs/rollingops/common/_base_worker.py
patriciareinoso: I am not a fan of using a given value to represent a missing value. Specially after killing all the processes in my computer because I forgot the pid > 0 validation and ended up doing os.kill(-1, SIGTERM). 

I isolated the None vs ""  to the pid properties.

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
patriciareinoso: I am reviewing this part of the code to ensure we do not have zombie processes here.

patriciareinoso:
@Gu1nness 
Thank you for the review. 

The main change is the use of pydantic for the `Operation`  and `OperationQueue` classes. 
These were the most critical since the stored object was incorrect.
This change also simplify all the `/` escaping when writing to etcd so thanks for that catch.

As for the rest of the databag modelling I think it can wait because the fields are simple and the change on the logic should not change the content of it. So I'll try to do it later.

While testing I caught some issues which I fixed: we were killing the process before the lease was revoke and the lock was released in case of `rollingops_etcd_failed` hook. And I improved the update-status reconciliation.

I will review the lease process because if the parent process die, it will not die.  So that comment is not yet answered.

rollingops/src/charmlibs/rollingops/etcd/_backend.py
Gu1nness: I think it should be yes.
If you want to use etcd but forget to install it, you should know about it

rollingops/src/charmlibs/rollingops/etcd/_etcdctl.py
Gu1nness: Perfect !

rollingops/src/charmlibs/rollingops/common/_models.py
```diff
@@ -34,7 +42,25 @@
...
+    - RETRY_RELEASE:
+        The operation failed or timed out and should be retried later.
+        The operation is re-queued and the lock is released so that
+        other units may proceed before this operation is retried.
+
+    - RETRY_HOLD:
+        The operation failed or timed out and should be retried immediately.
+        The operation is re-queued and the lock is kept by the current
+        unit, allowing it to retry immediately.
+    """
```
Gu1nness, line 63: Now that's excellent documentation!

rollingops/src/charmlibs/rollingops/common/_models.py
Gu1nness, line 1: Very solid implementation, way better
Thanks for that, I have more trust now!

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
Gu1nness: That's good, it respects the structure, we know exactly the lease_id and we're sure we've parsed what we wanted.

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -207,14 +224,14 @@ def move_head(self, to_queue_prefix: str) -> bool:
 
         op_id = kv.key.split('/')[-1]
         new_key = f'{to_queue_prefix}{op_id}'
-        data = json.dumps(kv.value)
-        value_escaped = data.replace('\\', '\\\\').replace('"', '\\"').replace('\n', '\\n')
+        op = Operation.model_validate(kv.value)
+        data = op.to_string()
```
Gu1nness, line 262: Damn I love that so much better.

rollingops/src/charmlibs/rollingops/etcd/_worker.py
```diff
@@ -35,8 +35,8 @@ class EtcdRollingOpsAsyncWorker(BaseRollingOpsAsyncWorker):
     manage its own worker lifecycle.
     """
 
-    _pid_field = WORKER_PID_FIELD
-    _log_filename = 'etcd_rollingops_worker'
+    _pid_field = 'etcd-rollingops-worker-pid'
```
Gu1nness, line 38: Nit: Why did this one slip to a hardcoded string from a constant ?

Gu1nness (approved):
This is so much better now!
Congrats!

sinclert-canonical:
👋🏻 Hey @patriciareinoso 

I know we _just_ spoke offline about the rollingops v0/v1 charmlib and its port to a standard Python package, which I assume this PR is doing. Please, let me know if that is not the case.

I wanted to share a possible improvement regarding log file locations, in the context of root-less Kubernetes charms (those where the workload container user does not have access to `/var/lib`, `/var/log`... unless explicitly granted). The improvement would be to allow the clients to specify the location for their log-file, instead of hard-coding them. I think it would make the library easier to integrate with that type of Kubernetes charms.

---

**Disclaimer:**

I do not know in which review stage this PR is. Feel free to completely ignore my feedback if the current approach has already been agreed upon. The last thing I want is to delay the porting of the _charmlib_ to a standard Python package.

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
```diff
@@ -0,0 +1,465 @@
...
+    back to peer-based processing when etcd becomes unavailable or
+    inconsistent.
+    """
+
+    def __init__(
+        self,
+        charm: CharmBase,
+        peer_relation_name: str,
+        etcd_relation_name: str,
+        cluster_id: str,
```
patriciareinoso, line 71: These 2 parameter should be optional.

Mehdi-Bendriss (approved):
Thank you Patricia! This is massive and really good work. 
I have 2 points that are not blocking the merge - as it targets a feature branch:
- Will you follow with a next PR for patching `setup_logging` with a logfile path (and making the etcd relation name optional)?
- Can you fix the docs CI steps that are failing?

patriciareinoso:
@Mehdi-Bendriss check  https://github.com/canonical/charmlibs/pull/445

It answers boths questions

PR #420, opened 2026-04-13
feat(istio-ingress-config): adds the istio-ingress-config interface

adhityaravi:
## Add istio-ingress-config interface library

Migrates the `istio-ingress-config` interface library from charmhub into charmlibs.

### New packages

- **`istio-ingress-config`** (`charmlibs.interfaces.gateway_metadata`): Interface for exchanging external authorizer configuration between an Istio ingress charm and the Istio control plane charm.

### What is this library for?

The [`istio-ingress-k8s`](https://github.com/canonical/istio-ingress-k8s-operator) charm needs to pass information about the external authentication provider to the [`istio-k8s`](https://github.com/canonical/istio-k8s-operator) core charm so the Istio control plane can be configured with an external authorizer.

The `istio-ingress-config` interface connects the `istio-ingress-k8s` charm to the `istio-k8s` charm, allowing the ingress to inform the control plane about the external auth decision provider with information including the decision service address, port, and header forwarding rules (which headers to include in auth checks, which headers to forward upstream/downstream on allow/deny decisions).

### Context

This library is already released to charmhub and its also being actively used. So this would be a pure migration.

### Related


interfaces/istio-ingress-config/src/charmlibs/interfaces/istio_ingress_config/__init__.py
```diff
@@ -0,0 +1,89 @@
...
+The `istio-ingress-k8s <https://github.com/canonical/istio-ingress-k8s-operator/>`_ charm
+needs to pass information about the external authentication provider to the
+`istio-k8s <https://github.com/canonical/istio-k8s-operator/>`_ core charm so the Istio
+control plane can be configured with an external authorizer.
+
+The ``istio-ingress-config`` interface connects the ``istio-ingress-k8s`` charm to the
+``istio-k8s`` charm, allowing the ingress to inform the control plane about the external
+auth decision provider with information including the decision service address, port,
+and header forwarding rules (which headers to include in auth checks, which headers to
+forward upstream/downstream on allow/deny decisions).
```
james-garner-canonical, line 33: Thanks for clearly explaining the purpose of the relation, this is very helpful. This relation feels very internal to the charmed Istio solution -- is there a vision that either of these charms could be swapped out for another charm?

adhityaravi: Oh you are right. This is infact an istio internal only interface. And probably would stay that way. Would that mean that it shouldn't go into charmlibs?

james-garner-canonical: That has been our approach so far -- internal interfaces don't need to be publicly documented on Charmhub and in the `charmlibs` docs, and their libraries can stay in team or product specific packages. Would that work well with your team's workflows?

adhityaravi: Thats fine, I can move this into our internal library. thanks

adhityaravi: Ill close the issue after I have moved it and added a reference to the move, if thats okay

adhityaravi:
This interface library is moved into the unified [`canonical-service-mesh`](https://github.com/canonical/service-mesh/pull/49) library. Hence this PR is closed.

PR #428, opened 2026-04-15
feat: add `valkey_client` interface

reneradoi:
This PR adds the new `valkey_client` interface to Charm interface specifications. The interface is owned by the Data-NoSQL team and is based on the existing `data-interfaces` library.


interfaces/valkey_client/interface/v0/tests/test_provider.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2026 Canonical
```
james-garner-canonical: Did you add interface tests because you find them useful or are using them in your own CI? If not, let's drop them from the PR, since we're planning to move away from them in `charmlibs` in future.

CODEOWNERS
```diff
@@ -80,6 +80,7 @@
 /interfaces/tempo_cluster/ @canonical/tracing-and-profiling
 /interfaces/tls-certificates/ @canonical/tls
 /interfaces/tracing/ @canonical/tracing-and-profiling
+/interfaces/valkey_client/ @canonical/data-nosql
```
james-garner-canonical: You did this correctly at the time of submission, but your PR made me realise that we should do this differently.
```suggestion

# /interfaces/valkey_client/
/interfaces/valkey_client/interface/ @canonical/data-nosql
/interfaces/valkey_client/ruff.toml @canonical/data-nosql
```
PS: I've applied this change for the interfaces already in `main` as well in #433, which has resulted in a conflict with `main` for you, sorry.

james-garner-canonical (commented):
Thanks for your PR, @reneradoi. A couple of questions:

What is the status of this interface? The `interface.yaml` has a 0 version and draft status, but the documentation and examples are very thorough. Is the interface still considered experimental and prone to change under users, or is it essentially stable?
- If it's essentially stable, let's update the version and status accordingly.
- If it's still early days and changes are possible, we can provide some review of the databag format. We've *almost* published our how-to guide covering our recommendations on this, but you can [read the preview version](https://canonical-ubuntu-documentation-library--372.com.readthedocs.build/charmlibs/how-to/design-relation-interfaces/), which shouldn't change much before publishing.

Out of curiosity, is the plan for the foreseeable future for charmers to use the data platform interfaces lib directly, or are you considering adding thin, wrapper libs to `charmlibs`, like `charmlibs.interfaces.valkey_client`?

reneradoi:
Thank you for the quick review @james-garner-canonical! The interface shouldn't change much anymore, spec/design are done and so is implementation. I've updated the status accordingly.

Regarding your question: Yes, the idea is that clients can use `data-interfaces` directly. It supports everything needed, and we are providing documentation (as part of our [charm docs](https://canonical-charmed-valkey.readthedocs-hosted.com/latest/how-to/clients/#implement-the-client-interface)) on how to do that easily.

interfaces/valkey_client/interface/v0/tests/test_provider.py
reneradoi: Nope, I've just added them to adhere to current practices. They are removed now.

CODEOWNERS
reneradoi: Thank you, I've merged your changes and updated accordingly.

james-garner-canonical (approved):
Thanks, @reneradoi!

james-garner-canonical (changes requested):
Actually there is one final thing I need to request, sorry. Could you please add an entry to the [interface libraries CSV](https://github.com/canonical/charmlibs/blob/main/.docs/reference/interface-libs.csv) so that `valkey_client` shows up in the [interface library listing table in the docs](https://documentation.ubuntu.com/charmlibs/reference/interface-libs/).

reneradoi:
> Actually there is one final thing I need to request, sorry. Could you please add an entry to the [interface libraries CSV](https://github.com/canonical/charmlibs/blob/main/.docs/reference/interface-libs.csv) so that `valkey_client` shows up in the [interface library listing table in the docs](https://documentation.ubuntu.com/charmlibs/reference/interface-libs/).

I've added a line to the CSV. The `valkey_client` interface is the first one published with the new Python package for `data-interfaces`, so the reference to it looks a bit different than for the other Data products/client interfaces.

interfaces/index.json
```diff
@@ -609,6 +609,16 @@
     "description": "The `tracing` interface supports push-based tracing.\nThe requirer publishes the list of protocols it wants to use to send traces.\nThe provider replies with an API URL and an endpoint for each protocol, omitting those it does not support.",
     "status": "published"
   },
+  {
+    "name": "valkey_client",
+    "version": "1",
+    "lib": "charms.data_platform_libs.data_interfaces",
```
james-garner-canonical:
```suggestion
    "lib": "dpcharmlibs.interfaces",
```

james-garner-canonical (commented):
Thanks @reneradoi, would be great if we can update the metadata for the other interfaces as they get updated to use the new lib as well.

interfaces/valkey_client/interface/v1/interface.yaml
```diff
@@ -0,0 +1,17 @@
+name: valkey_client
+version: 1
+status: published
+lib: charms.data_platform_libs.data_interfaces
```
james-garner-canonical:
```suggestion
lib: dpcharmlibs.interfaces
```

james-garner-canonical (approved):

PR #432, opened 2026-04-16
ci: use an environment for publishing

james-garner-canonical:
This PR updates the `build-n-publish` job of the `publish` workflow to use a Github environment.

CI in the PR can't really test this, but I'll be making a library release shortly after landing this PR which will test it.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -53,6 +53,7 @@ jobs:
   build-n-publish:
     needs: [unify-inputs,test-package]
     if: ${{ needs.unify-inputs.outputs.include != '[]' }}
+    environment: pypi
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yaml
```diff
@@ -53,6 +53,7 @@ jobs:
   build-n-publish:
     needs: [unify-inputs,test-package]
     if: ${{ needs.unify-inputs.outputs.include != '[]' }}
+    environment: pypi
```
tonyandrewmeyer: For what it's worth, ops/ops-scenario/ops-tracing, jubilant, and charmhub-listing-review use `publish-pypi`, although pytest-jubilant does use `pypi`. I don't think it's important to always have the same name, so I'm fine with this, but noting in case you wanted to follow the pattern.

tonyandrewmeyer (approved):

.github/workflows/publish.yaml
james-garner-canonical: Thanks!
```suggestion
    environment: publish-pypi
```

PR #433, opened 2026-04-16
ci: narrow the CODEOWNERS entries for interfaces without a library

james-garner-canonical:
This PR narrows the `CODEOWNERS` entries for interfaces without a library to own the `interface/` sub-directory and the `ruff.toml` file specifically, rather than the whole directory. This ensures that charmlibs-maintainers will be requested for review when a library is added to such an interface.


tonyandrewmeyer:
@james-garner-canonical do you want to resolve the conflicts before I review? It seems like it's conflicting on the main change so that seems more sensible than the other way around.

CODEOWNERS
```diff
@@ -22,6 +22,11 @@
 /.package/ @canonical/charmlibs-maintainers
 /interfaces/.package/ @canonical/charmlibs-maintainers
 
+# interfaces/index.json doesn't require CODEOWNERS review as its contents are validated in CI
+/interfaces/index.json
```
james-garner-canonical, line 26: This is moving up from the end of the file, so that the file ends with the massive list of interface entries.

james-garner-canonical:
Good catch, thanks @tonyandrewmeyer. I've merged `main` and reorganised the file a bit to minimise conflicts in future.

tonyandrewmeyer (approved):
The file is valid and the adjustment approach makes sense to me. I didn't do any validation that everything that was there previously is still; I'm happy to trust that you organised that correctly.

PR #437, opened 2026-04-16
feat(otlp): Update otlp to use generic rules class

swetha1654:
Tandem PR for the `cos-lib` [changes](https://github.com/canonical/cos-lib/pull/192).

Downstream tandem PR:

1. otel-collector: [test latest otlp interface](https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/246)

This pull request updates the OTLP interface to use the new `GenericRules` API from a feature branch of the `cos-lib` library, and refactors rule handling to align with recent upstream changes.

**Rule handling refactor:**

* Refactored `RuleStore` to use `GenericRules` with explicit Loki and Prometheus backends, replacing the previous `Rules` class and associated APIs. [[1]](diffhunk://#diff-86dd7e51fef4e1260abe3a4fe9af82aea0953746387faac2a34e2b4dff0ad21dL27-R41) [[2]](diffhunk://#diff-86dd7e51fef4e1260abe3a4fe9af82aea0953746387faac2a34e2b4dff0ad21dL59-R71)
* Updated rule validation and injection logic to use the new `validate` method since `otlp` is the only user of the `inject_and_validate` method and we can be sure that rules that come to the `OtlpProvider` will always have topology injected. [[1]](diffhunk://#diff-86dd7e51fef4e1260abe3a4fe9af82aea0953746387faac2a34e2b4dff0ad21dL491-R506)
* Updated backend-specific label injection, including using `CosTool` for PromQL label injection. [[1]](diffhunk://#diff-86dd7e51fef4e1260abe3a4fe9af82aea0953746387faac2a34e2b4dff0ad21dL322-R331)

**Testing updates:**

* Updated unit tests in `test_rules.py` to match the new rule handling logic, since we now don't do topology injection on the provider side anymore. 
* Adjusted imports to ignore private import usage warnings for compatibility with the new API.


interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -24,14 +24,21 @@
...
+from cosl.backends.loki import LokiRuleBackend
+from cosl.backends.prometheus import PrometheusRuleBackend
+from cosl.cos_tool import CosTool
 from cosl.juju_topology import JujuTopology
-from cosl.rules import HOST_METRICS_MISSING_RULE_NAME, Rules, generic_alert_groups
-from cosl.types import OfficialRuleFileFormat, SingleRuleFormat
+from cosl.rules import (
+    HOST_METRICS_MISSING_RULE_NAME,  # pyright: ignore[reportPrivateImportUsage]
+    GenericRules,
+    generic_alert_groups,  # pyright: ignore[reportPrivateImportUsage]
```
swetha1654: @MichaelThamm there is still a partial breaking change from my PR in cos-lib. Linting fails in the downstream libraries that use the latest version of cos-lib since `HOST_METRICS_MISSING_RULE_NAME` and `generic_alert_groups` are defined in `common_rules.py` and imported in `rules.py` and linter throws an error:
```
"HOST_METRICS_MISSING_RULE_NAME" is not exported from module "cosl.rules"
  Import from "cosl.common_rules" instead
```
This doesn't necessarily break the code itself, it still works. It just causes a linting failure. Users can either choose to ignore the linting failure or update the code to import these rules from `cosl.common_rules` instead.

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -319,7 +326,9 @@ def _duplicate_rules_per_unit(
                     for juju_unit in sorted(peer_unit_names):
                         rule_copy = copy.deepcopy(rule)
                         rule_copy.get('labels', {})['juju_unit'] = juju_unit
-                        rule_copy['expr'] = self._rules.promql.tool.inject_label_matchers(
+                        rule_copy['expr'] = CosTool(
+                            default_query_type='promql'
+                        ).inject_label_matchers(
```
swetha1654, line 331: `self._rules.promql` is of type `GenericRules` now and doesn't expose a tool attribute anymore so we can either use `self._rules.promql.backend.tool.inject_label_matchers` or we can use CosTool directly. I decided to use CosTool to keep it simple.

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
```diff
@@ -488,13 +497,13 @@ def rules(self) -> dict[int, RuleStore]:
...
             rules = RuleStore(self._topology)
-            logql_result = rules.logql.inject_and_validate_rules(
-                requirer.rules.logql, requirer.metadata
+            logql_result = rules.logql.validate(
+                cast('Mapping[str, list[OfficialRuleFileItem]]', requirer.rules.logql)
             )
-            promql_result = rules.promql.inject_and_validate_rules(
-                requirer.rules.promql, requirer.metadata
+            promql_result = rules.promql.validate(
+                cast('Mapping[str, list[OfficialRuleFileItem]]', requirer.rules.promql)
```
swetha1654, line 506: The new `GenericRules` class does not expose `inject_and_validate` but only validate. We can be sure that the rules obtained from `OtlpRequirer` databag will always have juju topology injected. This is because the OtlpRequirer class uses the `add` API from the upstream cosl before it dumps into the databag and this method takes care of adding the juju topology.

interfaces/otlp/tests/unit/test_rules.py
```diff
@@ -247,23 +232,8 @@ def test_provider_rules(
...
-                    # AND the rules are labeled with the provider's topology
-                    assert rule.get('labels', {}).get('juju_model') == MODEL_NAME
-                    assert rule.get('labels', {}).get('juju_model_uuid') == MODEL_UUID
-                    assert rule.get('labels', {}).get('juju_application') == app
-                    assert rule.get('labels', {}).get('juju_charm') == charm
-
-                    # AND the expressions are labeled
-                    assert f'juju_model="{MODEL_NAME}"' in rule['expr']
-                    assert f'juju_model_uuid="{MODEL_UUID}"' in rule['expr']
-                    assert f'juju_application="{app}"' in rule['expr']
```
swetha1654, line 266: the provier does not do topology injection anymore so we can remove this part of the test

interfaces/otlp/src/charmlibs/interfaces/otlp/_otlp.py
swetha1654: This is not the case anymore now that we have reverted the rules from common_rules to rules.py

james-garner-canonical:
@swetha1654 I see the [cos-lib PR](https://github.com/canonical/cos-lib/pull/192) is closed. Should this PR be closed too?

PR #438, opened 2026-04-20
chore: bump copyright year in the templates

dimaqq:
Normally we don't bump copyright year in source code files.

These however are copied to newly initialised libraries, and should correspond to when that's done.

We could parametrise it, or fix it up like this and buy us another year.


james-garner-canonical (approved):

james-garner-canonical:
Thanks! I think doing this manually makes sense for now, since if we add the year automatically, then CI will complain from the first CI run of the year onwards until we update the templates.

PR #441, opened 2026-04-21
feat(istio-ingress-route): adds istio-ingress-route lib

adhityaravi:
## Add istio-ingress-route interface library

Migrates the `istio-ingress-route` interface library from [`charmhub`](https://charmhub.io/istio-ingress-k8s/libraries/istio_ingress_route) into charmlibs.

### New packages

- **`istio-ingress-route`** (`charmlibs.interfaces.istio_ingress_route`): Interface for advanced ingress routing with custom path prefixes, mutli-port support etc. through the `istio-ingress-k8s` charm.

### What is this library for?

The `istio-ingress-k8s` charm supports the standard `ingress` interface, but that interface has limitations: it cannot open multiple ports, configure custom path prefixes, apply URL rewrites or redirects, or set up gRPC routing.

The `istio-ingress-route` interface fills this gap. It allows requiring charms to publish rich routing configurations like multi-port listeners, HTTP path matching, gRPC method routing, URL rewrite filters, request redirect filters that the `istio-ingress-k8s` charm translates into Kubernetes Gateway API resources (Gateway listeners, HTTPRoutes, and GRPCRoutes). 

This interface is specifically targeted at the `istio` class K8s Gateway API. (Even though one could argue that the library could be generically applied for all Gateway API classes, this probably wont be the case because of the plans for a better `ingress` V3 library by the platform engineering team)

### Context

This library is already in use by the Kubeflow and the COS charms. So this would be a pure migration.

### Related


adhityaravi:
@dstathis @james-garner-canonical 

A question:

This is a library already used by a few other charms. I tried to apply the same pattern from [here](https://github.com/canonical/charmlibs/pull/410) i.e., model the top level databag, remove the `StoredState`, to keep all the libs following the same pattern. The refactors kinda break the backwards compat. 

- Do you think that's fine and we should do the refactors now along with the migration?
- Or should we migrate and open issues to tackle these as v2?

I am thinking we should open issues for separation of concerns. But I am also thinking this is a migration why the hell not do it now.

james-garner-canonical:
> @dstathis @james-garner-canonical
> 
> A question:
> 
> This is a library already used by a few other charms. I tried to apply the same pattern from [here](https://github.com/canonical/charmlibs/pull/410) i.e., model the top level databag, remove the `StoredState`, to keep all the libs following the same pattern. The refactors kinda break the backwards compat.
> 
> * Do you think that's fine and we should do the refactors now along with the migration?
> * Or should we migrate and open issues to tackle these as v2?
> 
> I am thinking we should open issues for separation of concerns. But I am also thinking this is a migration why the hell not do it now.

I would say keep it exactly compatible with the existing library since there are already users.

I wouldn't even suggest trying to change the code or databag modeling in a backwards compatible way in this PR.

Since it sounds like this interface is a bit of an interim solution until the `ingress` v3 lib, I wouldn't worry about planning a v2 for this interface to switch to the recommended/consistent pattern -- it's not too big a deal for the libs to be inconsistent for now IMO.

dstathis:
> > @dstathis @james-garner-canonical
> > 
> > A question:
> > 
> > This is a library already used by a few other charms. I tried to apply the same pattern from [here](https://github.com/canonical/charmlibs/pull/410) i.e., model the top level databag, remove the `StoredState`, to keep all the libs following the same pattern. The refactors kinda break the backwards compat.
> > 
> > * Do you think that's fine and we should do the refactors now along with the migration?
> > * Or should we migrate and open issues to tackle these as v2?
> > 
> > I am thinking we should open issues for separation of concerns. But I am also thinking this is a migration why the hell not do it now.
> 
> I would say keep it exactly compatible with the existing library since there are already users.
> 
> I wouldn't even suggest trying to change the code or databag modeling in a backwards compatible way in this PR.
> 
> Since it sounds like this interface is a bit of an interim solution until the `ingress` v3 lib, I wouldn't worry about planning a v2 for this interface to switch to the recommended/consistent pattern -- it's not too big a deal for the libs to be inconsistent for now IMO.

@adhhityaravi can you explain the breaking change a bit more?

What we mostly care about here is whether the databag contents change. We don't want the cos charms to break because istio-ingress gets an update.

On the other hand, I see no reason why we can't break the library api as switching from the old charmlib is a manual change anyway. Additionally, the only users according to charmhub are observability, whom we communicate with every day.

adhityaravi:
@dstathis the relation databag stores plain strings at the moment. When we model it and use `relation.load` and `relation.save` it changes the databag content for bool items. 

So the problem is the opposite, the library API can stay the same. it doesnt need any update but the databag contents change. so when we update the `istio-ingress-k8s` charm with the charmlibs based interface its not anymore compatible with the COS charms. or the Kubeflow charms.

james-garner-canonical:
IMO databag backwards compatibility is more important than idiomatic databag modeling.

A few random ideas: could you use enums for modeling the string values? Would you be better off with passing a custom (de)serializer to save/load?

Caveat: I haven't looked at the actual code yet.

adhityaravi:
@adhityaravi Remove the `StoredState` usage which is completely a transparent change. So no changes to the databag or the API as discussed with @dstathis

adhityaravi:
@james-garner-canonical so we decided to neither break the API nor the databag for reasons similar to what you said

- it doesn't add enough value to do it now
- this is a pure migration task
- this is kinda sorta a transient solution

I just made a couple of pure cosmetic changes like making some classes private, dropping the use of `StoredState` which didn't make sense in the first place. I don't think you need to deep dive but a quick glance wouldn't hurt in my opinion. 

And ofcourse, thanks for being on top of this!

james-garner-canonical:
Looks reasonable to me. Happy to merge with @dstathis's approval on the code changes.

dstathis (approved):
+1 We discussed these changes in person.

james-garner-canonical (approved):
Thanks folks, looking good, I'll merge and publish now.

PR #442, opened 2026-04-22
feat(service-mesh): adds service-mesh library

adhityaravi:
## Add service-mesh interface library

Migrates the `service-mesh` interface library from [`charmhub`](https://charmhub.io/istio-beacon-k8s/libraries/service_mesh) into charmlibs.

### New packages

- **`service-mesh`** (`charmlibs.interfaces.service_mesh`): Interface for adding your charmed application to a service mesh, leveraging the `service_mesh` and `cross_model_mesh` endpoints to provide secure, policy-driven traffic management between applications

### What is this library for?

This library is for enrolling a charm onto a Charmed Service Mesh solution and automatically provisioning network policies that restrict cluster-internal network traffic between charms.

Service meshes provide capabilities for routing, controlling, and monitoring traffic between applications. A key feature is the ability to enforce authorization policies that govern which pods can communicate with each other and on which ports, paths, and HTTP methods. For example, you can define that a metrics scraper pod is allowed to `GET /metrics` on port `9090` from a producer pod, while preventing all other pods from accessing it.

The `ServiceMeshConsumer` subscribes a charm to a related service mesh by declaring access policies based on the charm's Juju relations. Since application relations often reflect traffic flow patterns (e.g. a database consumer connecting to a database provider), the consumer automatically generates the appropriate mesh traffic rules. It also handles labelling the charm's Kubernetes resources to enroll them in the mesh, and supports cross-model relations for multi-model deployments.

The `ServiceMeshProvider` publishes mesh enrollment labels and the mesh type to consumers, and collects the aggregated policies requested by all related consumer charms so the mesh control plane can enforce them.

This interface is designed to be mesh agnostic but at the moment is implemented only by the Istio ambient charms.

### Context

This library is already in use by the Kubeflow and the COS charms. So this would be a pure migration.

### Related


adhityaravi:
FYI: this interface library uses utilities from [canonical-service-mesh](https://github.com/canonical/service-mesh/pull/49) package. Hence merging is blocked until the `canonical-service-mesh` package PR is merged and released to PyPi

dstathis (approved):
Seems to be a direct migration.

adhityaravi:
@james-garner-canonical work from our side is done 

`canonical-service-mesh` package is not pinned anymore. Can be reviewed :)

interfaces/service-mesh/pyproject.toml
```diff
@@ -0,0 +1,75 @@
...
+    "License :: OSI Approved :: Apache Software License",
+    "Intended Audience :: Developers",
+    "Operating System :: POSIX :: Linux",
+    "Development Status :: 5 - Production/Stable",
+]
+dynamic = ["version"]
+dependencies = [
+    "ops",
+    "pydantic>=2",
+    "canonical-service-mesh",
```
james-garner-canonical: I may have missed flagging this in other PRs (if so, my apologies) -- but we should really set bounds on these dependencies to avoid breakages on major version releases. Perhaps:

```suggestion
    "ops>=3,<4",
    "pydantic>=2,<3",
    "canonical-service-mesh<1",
```

But I don't actually know what lowest `ops` version you need.

james-garner-canonical (approved):
Looks great, thanks for this. Just wanted to flag one issue for your attention before merging and releasing.

interfaces/service-mesh/pyproject.toml
adhityaravi: good catch! I will find the other libs and add pins to them as well! many thanks @james-garner-canonical!

PR #444, opened 2026-04-23
ci: add --output-all option to ls.py

james-garner-canonical:
This PR updates the `ls.py` script with an `--output-all` option. Currently you have to pass `--output` repeatedly for each required output to get the same result. This was OK at first, but as we've added more output fields, this has become painful, especially when running the script for local exploration.

I've also added a `--path-only` output option. This is the current default behaviour, and it seems like a good idea to be able to specify it explicitly.

### DIFF
--- a/.scripts/ls.py
+++ b/.scripts/ls.py
@@ -86,12 +86,14 @@ def _main() -> None:
     parser.add_argument('--indent-json', action='store_true')
     parser.add_argument('--regex', default=None)
     group = parser.add_mutually_exclusive_group()
-    group.add_argument(
-        '--output', action='append', choices=[f.name for f in dataclasses.fields(Info)]
-    )
+    all_fields = [f.name for f in dataclasses.fields(Info)]
+    group.add_argument('--output', action='append', choices=all_fields)
+    group.add_argument('--output-all', action='store_const', const=all_fields, default=[])
     group.add_argument('--name-only', action='store_true')
+    group.add_argument('--path-only', action='store_true')  # default behaviour
     args = parser.parse_args()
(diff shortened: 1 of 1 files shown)


.scripts/ls.py
```diff
@@ -86,12 +86,14 @@ def _main() -> None:
...
-        '--output', action='append', choices=[f.name for f in dataclasses.fields(Info)]
-    )
+    all_fields = [f.name for f in dataclasses.fields(Info)]
+    group.add_argument('--output', action='append', choices=all_fields)
+    group.add_argument('--output-all', action='store_const', const=all_fields, default=[])
     group.add_argument('--name-only', action='store_true')
+    group.add_argument('--path-only', action='store_true')  # default behaviour
     args = parser.parse_args()
-    single_output = 'name' if args.name_only else 'path'  # used if --output isn't specified
+    output = args.output or args.output_all
```
dwilding, line 95: This means that `--output version --output-all` will result in `version` only? I think I would have expected `--output-all` to override the individual `--output`s. But I don't think it's critical either way!

dwilding (approved):

.scripts/ls.py
james-garner-canonical: Because these are options for a mutually exclusive group, only one can be provided without an error, so I didn't really think about the order here (which ultimately doesn't matter).

dwilding: Ah got you, now I see `add_mutually_exclusive_group()`. Thanks for explaining that

james-garner-canonical: I'm now wondering if it would make sense to be able to pass --output-all to override, but I'll stick with the current implementation for now.

PR #445, opened 2026-04-23
feat: parameterize logs location and make etcd relation optional

patriciareinoso:
## Description

This PR

1. Allows to define the `base_dir` where all the files related to rolling ops will be written. 

We write:
    - Log file for the background proces
    - TLS certificates
    - ETCDCTL configuration file

The new parameter is optional, if not provided we default to `/var/lib/rollingops`

2. Makes the `etcd_relation_name` and `cluter_id` optional parameters. If not provided they default to `None`. In this case we would only be using the peer backend.

3. Add a `is_waiting` function.

There are no functional changes in:
- `rollingops/src/charmlibs/rollingops/etcd/_certificates.py`
- `rollingops/src/charmlibs/rollingops/etcd/_etcdctl.py`

But the diff is not nicely shown in github. I changed those modules to replace static functions with objects


rollingops/src/charmlibs/rollingops/_rollingops_manager.py
```diff
@@ -66,11 +67,12 @@ class RollingOpsManager(Object):
...
         charm: CharmBase,
-        peer_relation_name: str,
-        etcd_relation_name: str,
-        cluster_id: str,
         callback_targets: dict[str, Any],
+        peer_relation_name: str,
+        etcd_relation_name: str | None = None,
+        cluster_id: str | None = None,
         sync_lock_targets: dict[str, type[SyncLockBackend]] | None = None,
+        base_dir: str = '/var/lib/rollingops',
```
sinclert-canonical: I understand the ordering of the arguments is required so the mandatory ones come before the optional ones. I wonder whether we could make the `callback_targets` optional as well. What would happen if a particular application is only interested in emitting _synchronous_ lock events?

An additional benefit would be that we could place both `_targets` suffixed args next to each other:

```python
def __init__(
        self,
        charm: CharmBase,
        peer_relation_name: str,
        etcd_relation_name: str | None = None,
        cluster_id: str | None = None,
        callback_targets: dict[str, Any] | None = None,
        sync_lock_targets: dict[str, type[SyncLockBackend]] | None = None,
        base_dir: str = '/var/lib/rollingops',
):
```

---

The exact same argument applies to the  `peer_relation_name` argument (i.e. _what if someone is only interested in the cross-application rolling-ops?_). We could potentially make it optional as well, **unless** such relation is relied upon by the ETCD one to perform the cross-application rolling-ops.

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
sinclert-canonical, line 117: We _could_ make this relation optional as well, what would happen if a particular application is only interested in cross-application rolling-ops?

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
```diff
@@ -110,17 +114,23 @@ def __init__(
...
-            peer_relation_name=peer_relation_name,
-            etcd_relation_name=etcd_relation_name,
-            cluster_id=cluster_id,
-            callback_targets=callback_targets,
-        )
-        self.framework.observe(
-            charm.on[self.etcd_relation_name].relation_broken, self._on_etcd_relation_broken
-        )
+        self.etcd_backend: EtcdRollingOpsBackend | None = None
+        if etcd_relation_name and cluster_id:
```
sinclert-canonical, line 124: Absolutely mandatory to ease the onboarding for new consumer charms. Thanks!

sinclert-canonical (commented):
🧡  Thanks for the quick iteration on MySQL's feedback. This is looking great.

---

Please consider that, even if making the ETCD relation optional is the **greatest requirement lift** you could make for library consumers (thanks again!), it is only a piece of a greater pattern:

This library is, rightfully, offering a way to deal with both:
- Intra-application rolling-ops (thought the peer relation).
- Inter-application rolling-ops (thought the ETCD relation).

The library is also, rightfully, offering a way to emit / react to both:
- Synchronous events (less common within charms).
- Asynchronous events (most common within charms).

In both scenarios, library consumers may want to deal with **one, the other, or both**. Therefore, the _big pattern_ here would be to make all `RollingOpsManager` arguments that deal with these functionalities, optional, so each consumer can tweak the instantiation to their particular needs.

rollingops/src/charmlibs/rollingops/common/_base_worker.py
```diff
@@ -48,20 +48,28 @@ class BaseRollingOpsAsyncWorker(Object):
     _pid_field: str
     _log_filename: str
 
-    def __init__(self, charm: CharmBase, handle_name: str, peer_relation_name: str):
+    def __init__(
+        self,
+        charm: CharmBase,
+        handle_name: str,
+        peer_relation_name: str,
+        base_dir: str,
```
Gu1nness: Use `pathops.LocalPath` here so you're sure this is a valid path pointing locally instead of converting it later.

rollingops/src/charmlibs/rollingops/common/_utils.py
```diff
@@ -61,7 +62,8 @@ def datetime_to_str(dt: datetime) -> str:
 
 
 def setup_logging(
-    log_file: str,
+    base_dir: str,
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_backend.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         etcd_relation_name: str,
         cluster_id: str,
         callback_targets: dict[str, Any],
+        base_dir: str,
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_certificates.py
```diff
@@ -38,136 +38,138 @@
...
-    """
-    if not _exists():
-        return False
-    try:
-        stored = SharedCertificate.from_paths(
-            CLIENT_CERT_PATH,
-            CLIENT_KEY_PATH,
-            CA_CERT_PATH,
+class CertificateStore:
+    def __init__(self, base_dir: str):
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -36,14 +36,15 @@
 class EtcdLease:
     """Manage the lifecycle of an etcd lease and its keep-alive process."""
 
-    def __init__(self):
+    def __init__(self, base_dir: str):
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -154,9 +155,10 @@ class EtcdLock:
     automatically released if the owner stops refreshing the lease.
     """
 
-    def __init__(self, lock_key: str, owner: str):
+    def __init__(self, lock_key: str, owner: str, base_dir: str):
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -219,21 +221,22 @@ class EtcdOperationQueue:
     the value contains the serialized operation data.
     """
 
-    def __init__(self, prefix: str, lock_key: str, owner: str):
+    def __init__(self, prefix: str, lock_key: str, owner: str, base_dir: str):
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -378,10 +381,14 @@ class WorkerOperationStore:
     - requeue or delete completed operations
     """
 
-    def __init__(self, keys: RollingOpsKeys, owner: str):
-        self._pending = EtcdOperationQueue(keys.pending, keys.lock_key, owner)
-        self._inprogress = EtcdOperationQueue(keys.inprogress, keys.lock_key, owner)
-        self._completed = EtcdOperationQueue(keys.completed, keys.lock_key, owner)
+    def __init__(self, keys: RollingOpsKeys, owner: str, base_dir: str):
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_etcd.py
```diff
@@ -475,10 +482,14 @@ class ManagerOperationStore:
     Queue transitions and storage details remain encapsulated behind this API.
     """
 
-    def __init__(self, keys: RollingOpsKeys, owner: str):
-        self._pending = EtcdOperationQueue(keys.pending, keys.lock_key, owner)
-        self._inprogress = EtcdOperationQueue(keys.inprogress, keys.lock_key, owner)
-        self._completed = EtcdOperationQueue(keys.completed, keys.lock_key, owner)
+    def __init__(self, keys: RollingOpsKeys, owner: str, base_dir: str):
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_etcdctl.py
```diff
@@ -48,360 +47,358 @@
...
-        cert_path=str(client_cert_path),
-        key_path=str(client_key_path),
-    )
-
-    try:
-        with_pebble_retry(lambda: BASE_DIR.mkdir(parents=True, exist_ok=True))
-        with_pebble_retry(
-            lambda: CONFIG_FILE_PATH.write_text(json.dumps(asdict(config), indent=2), mode=0o600)
+class Etcdctl:
+    def __init__(self, base_dir: str):
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_relations.py
```diff
@@ -48,10 +48,11 @@
 class SharedClientCertificateManager(Object):
     """Manage the shared rollingops client certificate via peer relation secret."""
 
-    def __init__(self, charm: CharmBase, peer_relation_name: str) -> None:
+    def __init__(self, charm: CharmBase, peer_relation_name: str, base_dir: str) -> None:
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_relations.py
```diff
@@ -193,11 +194,14 @@ def __init__(
         relation_name: str,
         cluster_id: str,
         shared_certificates: SharedClientCertificateManager,
+        base_dir: str,
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_worker.py
```diff
@@ -35,11 +36,18 @@ class EtcdRollingOpsAsyncWorker(BaseRollingOpsAsyncWorker):
...
     """
 
-    _pid_field = 'etcd-rollingops-worker-pid'
+    _pid_field = WORKER_PID_FIELD
     _log_filename = ETCD_LOG_FILENAME
 
-    def __init__(self, charm: CharmBase, peer_relation_name: str, owner: str, cluster_id: str):
-        super().__init__(charm, 'etcd-rollingops-async-worker', peer_relation_name)
+    def __init__(
+        self, charm: CharmBase, peer_relation_name: str, owner: str, cluster_id: str, base_dir: str
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/etcd/_rollingops.py
```diff
@@ -64,6 +64,12 @@ def main():
     before exiting.
     """
     parser = argparse.ArgumentParser(description='RollingOps etcd worker')
+    parser.add_argument(
+        '--base-dir',
+        type=str,
```
Gu1nness: type here to correct type.

rollingops/src/charmlibs/rollingops/peer/_backend.py
```diff
@@ -197,7 +197,11 @@ class PeerRollingOpsBackend(Object):
     """
 
     def __init__(
-        self, charm: CharmBase, relation_name: str, callback_targets: dict[str, Callable[..., Any]]
+        self,
+        charm: CharmBase,
+        relation_name: str,
+        callback_targets: dict[str, Callable[..., Any]],
+        base_dir: str,
```
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/peer/_rollingops.py
```diff
@@ -24,6 +24,12 @@
 def main():
     """Juju hook event dispatcher."""
     parser = argparse.ArgumentParser(description='RollingOps peer worker')
+    parser.add_argument(
+        '--base-dir',
+        type=str,
```
Gu1nness: type here to `pathops.LocalPath`

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
Gu1nness: Same here

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
```diff
@@ -443,11 +472,34 @@ def state(self) -> RollingOpsState:
             operations=operations.queue,
         )
 
+    def is_waiting(self, callback_id: str, kwargs: dict[str, Any] | None = None) -> bool:
```
Gu1nness, line 479: Is this what should help us workaround the TLS scale-up loop ? 
I'm wondering how that works since the issue was that the leader could not see the unit new unit on the `rollingops-peers` relation.
If so: my idea was that the leader could receive a unit as parameter and check if that unit was in the pool of units in the `rollinops-peers` relation.

Gu1nness (commented):
Mostly moving from str to pathops.LocalPath IMHO.
Otherwise, all good.

About the `is_ready ` is it `is_waiting` that you implemented ? Do that solve our problem ?

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
patriciareinoso: non, this is related to this comment https://github.com/canonical/mongo-single-kernel-library/pull/299#discussion_r3118107775

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
patriciareinoso: The cross-application rollingops relies on both the peer relation and the etcd relation. 
If something goes wrong with the etcd lock (we are not able to reach the etcd, the relation was removed, ... ) we fallback to the peer solution. 

To be able to fallback without loosing any operation we use the peer relation as a backup operation storage. 
So we cannot make the peer relation optional in that case

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
patriciareinoso: I understand your point but from the beginning the API was not designed to support only sync operations. Async locks remain the main scenario and the callback_targets is the core of the library. I would prefer to keep it mandatory. If a particular application does not need async callbacks, passing an empty dictionary is always an option. 

As for grouping the _targets arguments together, I agree it reads nicely, but I’d still prioritize keeping the signature aligned with the original intent and consistency of the API over reordering for aesthetics.

The answer about the peer relation name is answered in the comment below

sinclert-canonical: Alright, the PR is perfectly fine as is. I just wanted to commend on it as anticipation for future use-cases.

sinclert-canonical (approved):
Thanks for your tireless efforts. This is looking great!

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
patriciareinoso: I added it as an idea for future improvements [here](https://docs.google.com/document/d/12OA-k7GRHHS4AGkefThLdihmBS6RuXOtQU1Qyn46rkI/edit?tab=t.0#heading=h.merb02txzct4)

rollingops/src/charmlibs/rollingops/_rollingops_manager.py
Gu1nness: Ah yes I forgot.
Thanks!

Gu1nness (approved):

Mehdi-Bendriss (approved):

PR #450, opened 2026-04-28
docs: describe how pathops uses tag

james-garner-canonical:
This PR adds the two line-comments from the template to the pathops pyproject.toml file, and an inline comment explaining what pathops uses the tag variable for.


tromai (approved):
Thanks for the change. It looks great.

PR #452, opened 2026-04-29
chore(service-mesh): pins deps across service mesh related libraries

adhityaravi:
Adds major version pins for dependencies across the following service mesh related libraries

- `istio-ingress-route`
- `istio-request-auth`
- `service-mesh`


adhityaravi:
@dstathis Should we pin the ops lib already to v3? currently all our charms use v2. but i think v2 is old.

james-garner-canonical:
Ah I may have steered you in the wrong direction here with `ops>=2,<3` in the other PR. Ideally libraries should be a bit more tolerant, since Ops 2 and Ops 3 are very compatible (it's just that Ops 3 doesn't support Python 3.8 and may grow new features that Ops 2 won't get), so your libs probably work just fine with Ops 2 or Ops 3 based charms right now. For example, `charmlibs.pathops` specifies `ops>=2.19,<4`.

WDYT about updating `service-mesh` as well in this PR to use a wider range of supported versions? Also consider bumping the library patch version for each library to release these changes.

adhityaravi:
@james-garner-canonical if you say ops v3 is backwards compatible with v2 as long as we're py >= 3.8, I am more than happy to do that pin.

james-garner-canonical:
> @james-garner-canonical if you say ops v3 is backwards compatible with v2 as long as we're py >= 3.8, I am more than happy to do that pin.

Sounds good. Let's go with `>=2.23.1,<4` (the typed relation data feature landed in 2.23).

adhityaravi:
@james-garner-canonical done!

dstathis (approved):

PR #453, opened 2026-04-29
docs: better example of Ops dependency bounds in template and examples

james-garner-canonical:
This PR updates the template and examples with better (but still commented out) dependency bounds for Ops, along with an explanation of the general principle, and the specifics for Ops.

### DIFF
--- a/.example/pyproject.toml
+++ b/.example/pyproject.toml
@@ -17,9 +17,11 @@ dynamic = ["version"]
 dependencies = [
     # Libraries should use clear but wide dependency bounds to avoid overly constraining charms.
     # Ops 3 is backwards compatible with Ops 2, but drops Python 3.8 support.
-    # Ops 3 has new features not present in Ops 2.
-    # To review the changes in Ops 3, see https://github.com/canonical/operator/releases.
-    # Set the dependency bounds according to the features your library needs.
+    # Ops 3 has new features not present in Ops 2.23 (the last Ops 2 feature release).
+    # See the Ops 2.23 maintenance branch changelog and the Ops 3 changelog for current information:
+    # - https://github.com/canonical/operator/blob/2.23-maintenance/CHANGES.md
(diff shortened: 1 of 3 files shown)


.example/pyproject.toml
```diff
@@ -15,7 +15,11 @@ classifiers = [
 ]
 dynamic = ["version"]
 dependencies = [
-    # "ops",
+    # Libraries should use clear but wide dependency bounds to avoid overly constraining charms.
+    # Set the dependency bounds according to the features your library needs.
+    # Ops 3 is backwards compatible with Ops 2, but drops Python 3.8 support.
+    # Ops 3 may have new features not present in Ops 2.
```
dwilding: Is the "may" here because we don't know which specific version of 3.x someone might put as an upper bound? I feel like we'd be fine to be more definitive:


```suggestion
    # Ops 3 has new features not present in Ops 2.
```

dwilding: How about mentioning the releases page, since we don't have a changelog in the docs?


```suggestion
    # Ops 3 may have new features not present in Ops 2.
    # To review the changes in Ops 3, see https://github.com/canonical/operator/releases.
```

dwilding (commented):
Good idea to add this. I'm proposing a couple of tweaks.

dwilding (approved):

.example/pyproject.toml
james-garner-canonical: It's because I couldn't remember what we've added in Ops 3 that didn't make it into Ops 2. Maybe we should link to the changelog here.

james-garner-canonical: Or the releases page, good idea.

james-garner-canonical: Seems like `ops.hookcmds` is the only big non-testing feature, but since there's something I'll use your wording, thanks.

james-garner-canonical: I've gone with linking to the changelogs.

PR #454, opened 2026-04-29
feat: path to etcdctl

patriciareinoso:
We add a part in the charmcraft file to include the etcdctl. 
We install it in the <charm-dir>/usr/bin/etcdctl so we need to retrieve that path in the ETCD related classes

### DIFF
--- a/rollingops/tests/integration/charms/k8s/charmcraft.yaml
+++ b/rollingops/tests/integration/charms/k8s/charmcraft.yaml
@@ -16,6 +16,20 @@ parts:
     plugin: uv
     build-snaps: [astral-uv]
 
+  etcdctl:
+    plugin: make
+    source: https://git.launchpad.net/etcd
+    source-type: git
+    source-tag: lp-v3.6.7
+    build-snaps:
+      - go/latest/stable
+    override-build: |
+      set -eux
+      make build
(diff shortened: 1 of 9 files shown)


rollingops/tests/integration/charms/k8s/charmcraft.yaml
```diff
@@ -16,6 +16,20 @@ parts:
     plugin: uv
     build-snaps: [astral-uv]
 
+  etcdctl:
+    plugin: make
+    source: https://git.launchpad.net/etcd
+    source-type: git
+    source-tag: lp-v3.6.7
```
Mehdi-Bendriss: you can use `lp-v3.6.10`

rollingops/tests/integration/charms/machine/charmcraft.yaml
```diff
@@ -16,6 +16,20 @@ parts:
     plugin: uv
     build-snaps: [astral-uv]
 
+  etcdctl:
+    plugin: make
+    source: https://git.launchpad.net/etcd
+    source-type: git
+    source-tag: lp-v3.6.7
```
Mehdi-Bendriss: lp-v3.6.10

Mehdi-Bendriss (commented):
Thanks Patricia, looks good. Minor comments on the etcd version

PR #456, opened 2026-04-30
ci: upload charmcraft logs

james-garner-canonical:
We see regular `charmcraft pack` failures in CI, which appear to be due to `snapd` startup issues when packing from a cold start.

Here's an [example failure](https://github.com/canonical/charmlibs/actions/runs/25139482207/job/73685763313). Here's the `charmcraft` issue: https://github.com/canonical/charmcraft/issues/2515

---

This PR  adds a follow-up step to upload the `charmcraft` logs after running integration tests (whether we succeed or fail -- so long as either the pack or integration step ran).

In order for the uploaded artifact to have a unique name, it is prefixed with the name of the package under test. The package name provided as input to the workflow may contain a slash (e.g. `interfaces/tls-certificates`), which is invalid for an artifact name. The cleanest way to get a valid and unique name is to get the distribution package name via `.scripts/ls.py`.

To avoid having to perform `jq` wizardry to extract the bare name from the JSON list that `ls.py` would output, I've updated `ls.py` with a `--no-json` argument (which prints each entry on a separate line instead -- so the output for a single field is just `<field value>\n`). As a driveby improvement, I've collapsed the `--name-only` and `--path-only` arguments into the more general `--output-only <field>`, which simplifies the handling of the various `--output` arguments and makes querying any single field possible without needing to add a custom argument.

---

There's a charmcraft issue tracking what appears to be the same failure that we see: https://github.com/canonical/charmcraft/issues/2515

### DIFF
--- a/.scripts/ls.py
+++ b/.scripts/ls.py
@@ -85,15 +85,13 @@ def _main() -> None:
     parser.add_argument('--only-if-version-changed', action='store_true')
     parser.add_argument('--indent-json', action='store_true')
     parser.add_argument('--regex', default=None)
+    parser.add_argument('--no-json', action='store_true')
     group = parser.add_mutually_exclusive_group()
     all_fields = [f.name for f in dataclasses.fields(Info)]
-    group.add_argument('--output', action='append', choices=all_fields)
-    group.add_argument('--output-all', action='store_const', const=all_fields, default=[])
-    group.add_argument('--name-only', action='store_true')
-    group.add_argument('--path-only', action='store_true')  # default behaviour
+    group.add_argument('--output', action='append', choices=all_fields, dest='output')
+    group.add_argument('--output-all', action='store_const', const=all_fields, dest='output')
+    group.add_argument('--output-only', action='store', default='path')  # default behaviour
     args = parser.parse_args()
-    output = args.output or args.output_all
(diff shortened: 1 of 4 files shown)


.scripts/ls.py
```diff
@@ -85,15 +85,13 @@ def _main() -> None:
     parser.add_argument('--only-if-version-changed', action='store_true')
     parser.add_argument('--indent-json', action='store_true')
     parser.add_argument('--regex', default=None)
+    parser.add_argument('--no-json', action='store_true')
```
dwilding, line 88: I had to think a while to understand what `--no-json` signals, as eg `.scripts/ls.py interfaces --no-json --output name` will still output a JSON object per line.

How about using a different name - something like `--no-array` or `--lines`?

dwilding (approved):

.scripts/ls.py
james-garner-canonical: Technically it outputs a Python dictionary per line, but in practice since it's a `dict[str, str]` it looks very similar. However, since it's formatted with `'` rather than `"`, piping an entry into `jq` will fail, so I think we do meet the bar for `--no-json` not outputting valid JSON.
```
jq: parse error: Invalid string literal; expected ", but got ' at line 1, column 8
```

PR #460, opened 2026-05-01
feat: charmlibs.snap 2.0

james-garner-canonical:
This PR rewrites the `charmlibs.snap` library to drop its caching layer, and use the API exclusively rather than an ad-hoc mixture of CLI and API use. All snap operations are now provided via top-level functions rather than a mixture of top-level functions and methods on snap objects.

The design principle is to wrap the API in a clean and Pythonic way, basing function naming and error behaviour on the snap CLI to take advantage of user familiarity with the CLI.

This PR is scoped to install/remove/refresh, info, logs, and the ensure and ensure_revision helpers. Follow up PRs will add aliases, apps (services), config, and interfaces before the initial 2.0 release.


snap/src/charmlibs/snap/_client.py
```diff
@@ -0,0 +1,238 @@
+# Copyright 2021 Canonical Ltd.
```
tonyandrewmeyer: Nit: this seems new, so 2026.

snap/src/charmlibs/snap/_client.py
```diff
@@ -0,0 +1,238 @@
...
+
+from . import _client_sockets, _errors
+
+if typing.TYPE_CHECKING:
+    import http.client
+    from collections.abc import Generator
+
+logger = logging.getLogger(__name__)
+
+# we need the powerful snapd socket
```
tonyandrewmeyer: Nit: Is there a non-powerful snapd socket?

snap/src/charmlibs/snap/_client.py
```diff
@@ -0,0 +1,238 @@
...
+    """POST request to snapd REST API."""
+    return _request('POST', path, body=body)
+
+
+def put(path: str, body: dict[str, Any] | None = None):
+    """PUT request to snapd REST API."""
+    return _request('PUT', path, body=body)
+
+
+def _request(
```
tonyandrewmeyer: It's annoying that the "one obvious way" language has so many http request options...

I assume the and the raw method could be far shorter if they were done with `requests` (or `httpx` but async doesn't seem too useful here). I also assume we're preferring the stdlib to keep dependencies minimal. Is that the right trade-off here? `requests` is so well established and used by many charms already that I wonder if it's worth considering.

snap/src/charmlibs/snap/_client.py
```diff
@@ -0,0 +1,238 @@
...
+    if not isinstance(response_dict, dict):  # pyright: ignore[reportUnnecessaryIsInstance]
+        raise _errors.SnapBadResponseError(
+            message=f"Unexpected response type {type(response_dict).__name__!r} for path {path!r}, expected a 'dict'",  # noqa: E501
+            kind='charmlibs-snap',
+            value=str(response_dict),
+            status_code=response.status,
+            status=response.reason,
+        )
+    try:
+        match response_dict['type']:
```
tonyandrewmeyer: Old man yells at cloud: seems to me if/elif/else would be just as good here.

snap/src/charmlibs/snap/_errors.py
```diff
@@ -0,0 +1,154 @@
...
+
+    This class is private because the public disconnect function suppresses this error,
+    following the snap CLI's lead.
+    """
+
+
+class SnapOptionNotFoundError(SnapAPIError):
+    """Raised via the API when the specified snap config option is not found.
+
+    Note that ``SnapOptionNotFoundError.value`` is typically a ``dict``,
```
tonyandrewmeyer: This feels a bit ugly. Is it the only case where it's not a str?

snap/src/charmlibs/snap/_errors.py
```diff
@@ -0,0 +1,154 @@
...
+
+        {'SnapName': snap_name, 'Key': key}.
+    """
+
+
+class SnapChangeError(SnapAPIError):
+    """Raised when a snap change results in an error or has an unexpected status."""
+
+
+def _error_type_from_result_kind(kind: str) -> type[SnapAPIError]:  # pyright: ignore[reportUnusedFunction]
```
tonyandrewmeyer: Old man yells at cloud: would this not be better as a dict?

snap/src/charmlibs/snap/_functions.py
```diff
@@ -0,0 +1,68 @@
...
+"""High level helper functions that build on top of the basic snap operations."""
+
+from . import _errors, _snapd_snaps, _utils
+
+
+def ensure_revision(snap: str, revision: int, *, classic: bool = False) -> bool:
+    """Ensure the snap is installed at the specified revision.
+
+    Returns:
+        True if the snap was installed or updated, False otherwise.
```
tonyandrewmeyer: Bools are nice, but I wonder if sometimes you want to know whether it was an install or update, for example because you need to do more follow-up work on install. Same for `ensure`.

snap/src/charmlibs/snap/_functions.py
```diff
@@ -0,0 +1,68 @@
...
+      channel is specified), it will be refreshed only if update = ``True`` (default).
+
+    Returns:
+        True if the snap was installed or updated, False otherwise.
+    """
+    info = _snapd_snaps.info(snap, missing_ok=True)
+    if info is None:  # Not installed.
+        _snapd_snaps.install(snap, channel=channel, classic=classic)
+        return True
+    if channel is not None and info.channel != _utils._normalize_channel(channel):
```
tonyandrewmeyer: Elsewhere I've seen UK spelling. We should be consistent even in private methods. Sadly that probably means US English throughout.

snap/src/charmlibs/snap/_utils.py
```diff
@@ -0,0 +1,60 @@
...
+def _normalize_channel(channel: str) -> str:  # pyright: ignore[reportUnusedFunction]
+    """Normalize a snap channel string to the form "track/risk".
+
+    Channels may be specified as track or risk only, or as "track/risk" or "track/risk/branch".
+    Snapd uses default values internally, but will record the *requested* value in the snap info.
+    This function normalizes channels with no "/" to the form "track/risk" for easier comparison.
+    """
+    if '/' not in channel:
+        if channel not in ('edge', 'beta', 'candidate', 'stable'):
+            # track only, append default risk
```
tonyandrewmeyer: Nit: I know we have different options on this, but I would prefer comments as sentences.

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+        return cls(
+            name=info_dict['name'],
+            channel=info_dict['channel'],
+            revision=int(info_dict['revision']),
+            version=info_dict['version'],
+            classic=info_dict['confinement'] == 'classic',
+            hold=info_dict.get('hold'),
+        )
+
+    @property
```
tonyandrewmeyer, line 63: What's the reason for all the properties rather than just attributes?

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+def info(snap: str, *, missing_ok: Literal[True]) -> Info | None: ...
+def info(snap: str, *, missing_ok: bool = False) -> Info | None:
+    """Get information about an installed snap.
+
+    This function implements the semantics of the `snap list` command,
+    restricted to a single snap.
+
+    Args:
+        snap: the name of the snap.
+        missing_ok: if ``True``, return ``None`` if the snap is not installed.
```
tonyandrewmeyer: I don't really like this style in general; is there a reason we have it? Why not just always have the True behaviour and let the caller check for None? Or if you don't like that for typing, always have the False behaviour and ignore the tiny performance hit.

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+    strict: bool = False,
+) -> None:
+    """Install a snap.
+
+    Raises:
+        ValueError: if both channel and revision are specified.
+        SnapAlreadyInstalledError: if the snap is already installed.
+        SnapError: (or a subtype) if the snap could not be installed as requested.
+    """
+    if channel is not None and revision is not None:
```
tonyandrewmeyer, line 153: Couldn't this be an overload?

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+    return Info._from_dict(info_dict)
+
+
+def install(
+    snap: str,
+    *,
+    channel: str | None = None,
+    revision: int | None = None,
+    classic: bool = False,
+    strict: bool = False,
```
tonyandrewmeyer: Instead of an argument, what about returning a bool instead?

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+    # NOTE: Unlike the API, the CLI doesn't error if the snap isn't installed (just prints a msg).
+    try:
+        _client.post(f'/v2/snaps/{snap}', body=data)
+    except _errors.SnapNotFoundError:
+        if strict:
+            raise
+
+
+def refresh(
+    snap: str, channel: str | None = None, revision: int | None = None, strict: bool = False
```
tonyandrewmeyer: Strict should be kwonly, and maybe revision too?

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+) -> None:
+    """Refresh a snap.
+
+    Raises:
+        ValueError: if both channel and revision are specified.
+        SnapError: (or a subtype) if the snap could not be refreshed as requested.
+    """
+    if channel is not None and revision is not None:
```
tonyandrewmeyer, line 230: Also seems like an overload opportunity?

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+    _client.post(f'/v2/snaps/{snap}', body=data)
+
+
+def unhold(snap: str) -> None:
+    """Unhold a snap to allow it to be refreshed."""
+    # NOTE: Neither the API nor CLI error if the snap isn't installed or held.
+    _client.post(f'/v2/snaps/{snap}', body={'action': 'unhold'})
+
+
+def _list_snaps() -> list[Info]:  # pyright: ignore[reportUnusedFunction]
```
tonyandrewmeyer: I haven't found where this is used yet, but it feels like this and list channels would maybe be better as generators?

snap/src/charmlibs/snap/_snapd_logs.py
```diff
@@ -0,0 +1,69 @@
...
+    import datetime
+
+logger = logging.getLogger(__name__)
+
+
+# /v2/logs
+
+
+@dataclasses.dataclass
+class LogEntry:
```
tonyandrewmeyer, line 35: Maybe frozen? It doesn't seem like people should edit them.

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,115 @@
...
+    slot: str | None = None,
+    /,
+    *,
+    forget: bool = False,
+) -> None:
+    """Disconnect a plug from a slot."""
+    data: dict[str, Any] = {'action': 'disconnect'}
+    if slot_snap is None:
+        assert slot is None
+        # Called with 2 arguments, treat as as snap disconnect <snap>:<slot>
```
tonyandrewmeyer: It doesn't feel Pythonic to have the args switch like this. I assume it's to mimic the CLI, but I think it should be reconsidered.

tonyandrewmeyer (commented):
If I recall correctly, you've been doing some API planning, had talks with Ben about that, etc. I wonder if an explanation doc that outlines the thought behind the design, like an extended version of the PR description, might be good to add now. Partly it helps review, but it also helps long term.

You mention AI use in the description. I'm obviously good with that, but it would be good to know if there are parts that are predominately AI generated (other than the mentioned docstrings) as I'd review those with less trust.

This PR is very large. Is the plan landing it like this? Maybe stacked PRs or logical commits would be better (tbh I haven't checked the commits, maybe I can already review that way). I care most about API shape, and it's not simple to see in a change this big, plus I do want to properly review the rest.

I haven't had time to play with actually using it, sorry.

I didn't read any of the test code yet, sorry.

I wonder about the always-wait decision. Pebble does this too, and it works well for charms there. In some cases, it feels like giving snapd a bunch of work to do and then waiting for it all is handy (in my experience with concierge and manual use). It's not always parallelised, though. So I'm unsure on this - would be be able to reverse this decision if we wanted to later, without needing a v3?

Overall, this feels very comprehensive. How targeted at charms is this, vs this is just a nice Python API for snapd?

snap/src/charmlibs/snap/_client.py
james-garner-canonical: I was trying to make a cheeky reference to the snap CLI's [source code](https://github.com/canonical/snapd/blob/dee46fb8fe1b0d4f1f7c6d82270bdd43996011a7/cmd/snap/main.go#L382), but probably best to drop it.
```go
	// we need the powerful snapd socket
	Socket: dirs.SnapdSocket,
```

snap/src/charmlibs/snap/_client.py
james-garner-canonical: I'm not super convinced that using `requests` or `httpx` would actually get us all that much here. I did start with `requests` and `requests-unixsocket` for initial prototyping, which was nice, but:
- We'd still need a custom unix socket handler implementation (or use another third party dep like `requests-unixsocket`).
- We don't want to just let errors bubble up (whether explicit errors returned by the API, or timeouts/etc), so we'd still need similar logic for catching/inspecting/wrapping them.
- We'd still need the same amount of complexity for interpreting the actual responses.

If I'm correct, I think that means that we'd really only get to make these lines nicer, which doesn't seem worth the cost of the dependency to me.
```
    # _request
    headers = {'Accept': 'application/json'}
    if body is not None:
        data = json.dumps(body).encode('utf-8')
        headers['Content-Type'] = 'application/json'
    else:
        data = None
    response = _request_raw(method, path, query=query, headers=headers, data=data)
```

```
    # _request_raw
    request = urllib.request.Request(
        f'http://localhost{path}'
        + (f'?{urllib.parse.urlencode(query, doseq=True)}' if query else ''),
        method=method,
        headers=headers or {},
        data=data,
    )
    opener = urllib.request.OpenerDirector()
    opener.add_handler(_client_sockets.UnixSocketHandler(_SOCKET_PATH))
    opener.add_handler(urllib.request.HTTPRedirectHandler())
    try:
        return opener.open(request, timeout=_REQUEST_TIMEOUT)
```

snap/src/charmlibs/snap/_client.py
james-garner-canonical:
```suggestion
# Copyright 2026 Canonical Ltd.
```

snap/src/charmlibs/snap/_client.py
james-garner-canonical: You're right, the benefits of match-case here are marginal. I do like that we really know up front what we're branching on (you'd have to modify the `match` statement to match on additional information, vs just editing an `elif` branch), but it's not really a big deal. However, I don't see any reason we should avoid language features like match-case.

snap/src/charmlibs/snap/_errors.py
james-garner-canonical: It's the only case I've encountered in testing so far. I encountered it quite late, and debated several different ways of handling it, including implementing a custom constructor for this case that turned `value` into something like `SnapName:Key`. Ultimately I decided that was a bad idea for two reasons: (1) we don't know when some other part of the API might decide to return a value like this; (2) the error `repr` ends up being a bit less obvious to readers.

What I am still wondering about is whether we should unconditionally convert `value` to `str`. So here we'd have the string `"{'SnapName': snap_name, 'Key': key}"`. It feels a bit wrong to 'throw away' information like that, but maybe it's the right approach. I do think that if users wanted to do something with `value`, their safest bet would be to convert it to a `str` for operations like `'foo' in str(e.value)'`, so maybe we'd just be making the UX nicer.

snap/src/charmlibs/snap/_errors.py
james-garner-canonical: This would work as a `dict`, but with the following downsides:
- (minor) can be less clear when we're collapsing multiple `kind` strings into a single error type since they're separate (not necessarily adjacent) dictionary entries.
- (major) the default value is determined separately at the `get` call site rather than at the dictionary definition site like the other values (unless we use a `defaultdict`, but I still think they're too error prone and would prefer to avoid them).

snap/src/charmlibs/snap/_functions.py
james-garner-canonical: This is a very good point, but I'm not sure how best to address it.

---

We could return, for example "installed", "refreshed", None.

Then with `ensure` we'd need to consider distinguishing refreshed to a different channel from refreshed on the same channel. We also wouldn't know whether a refresh to a different channel actually changed the installed version, or just the changed the channel being tracked (e.g. both channels currently point to the same revision), though since we query `info` first we could check. So we could have values representing "installed", "refreshed-to-a-different-revision", "refreshed-but-same-revision", "no-change" (e.g. `None`).

Or maybe we only want to care about whether it was previously installed or not in both cases, and a ternary value is fine.

---

Switching to enumerated types *would* be backwards compatible with a 'returns truthy values on change' API, though technically incompatible with 'returns exactly a `bool`'. Adding a new value to our enumerated types later would be backwards incompatible unless it was somehow a sub-value of an existing value (I don't think this is really possible without scary `__eq__` shenanigans).

---

Another thing a user might care about (more than revisions, potentially) is whether the 'version' field has changed. The only reasonable way to accommodate that on top of the rest that I can think of is some sort of `Change` type (or `None` if there are no changes), like this:
```py
Change(
    installed=False,  # Already installed.
    refreshed=True,  # We did a refresh.
    revision=True,  # The new revision is different.
    version=False,  # But actually the installed version is the same.
)
```

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: Here's why I think this approach is reasonable:
- `range` works similarly, with argument position meaning a different thing depending on how many args there are: `range(stop)` and `range(start, stop)`. That's similar to `disconnect(slot)` and `disconnect(plug, slot)`.
- The arguments are positional only, so the user should never have to think about the names switching or anything.
- We follow the same four argument pattern for `connect`, with the same semantics for the 3 and 4 argument cases, but the opposite semantics for the short form (`connect(plug)`, `disconnect(slot)`).

---

That said, I definitely don't love it. But I'm not really sure what a good alternative would be.

---

For example, if we switched to keyword-only arguments, `disconnect` would need both the plug and slot arguments to be optional,  making it impossible for the type checker to tell if you're making an invalid 1 or 0 argument call.

It might be worth grouping the argument pairs into `Plug` or `Slot` types, or just `tuple`s, though that makes the three argument cases a bit uglier (`disconnect((snap, plug), (snap, None))`).

If we went with objects, we could do something like `Plug(snap, plug)` with `connect` and `disconnect` methods. And `Plug(None, None).disconnect(snap, slot)` would allow the current two argument form of `disconnect`.

---

Perhaps a better alternative is different functions for the 2 and 3-4 argument cases.
- `connect` with 3 required and 1 optional argument, and `connect_plug` with only 2 arguments.
- `disconnect` with 3 required and 1 optional argument, and `disconnect_slot` with only 2 arguments.

snap/src/charmlibs/snap/_snapd_logs.py
james-garner-canonical: Well spotted. What I'd actually like to do is use a non-`dataclass` with read-only properties so we don't opt into the whole `dataclass` API.

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: Alluded to in my response to the `LogEntry` comment, but:
1. If this was a `dataclass` it would make sense for it to be `frozen` since `my_snap_info.name = 'something-else'` doesn't mean anything to the library.
2. I think it's worth paying some boilerplate cost in the library implementation to avoid opting into the whole `dataclasses` API + `object.__setattr__` if we have `frozen=True`.

Should probably add `__repr__` at least and maybe `__eq__` (since equality via identity will be surprising to users).

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: - I think it's important for the library to have an ergonomic way to probe if a snap is installed, and I don't think that catching an exception is ergonomic enough, so I definitely want to enable the `if some_function(my_snap) ...` pattern.
- As you say, the optional return typing is painful, we don't want users to always have to write `assert my_info is not None`. I figure it makes sense to default to `missing_ok=False` so the case where we want less boilerplate is as boilerplate free as possible, while the explicit 'check if installed' case is a little more verbose.
- If we were to implement a separate `is_installed` function, we'd hit the same API and throw away the response, which feels wasteful, since if the snap *is* installed it would probably be common to call `info` next. More importantly though, users might prefer `try: snap_info = info(snap)` over `if is_installed(snap): info(snap)` anyway (similar boilerplate, minor performance benefit, less racy), so we only get the ergonomic win from `info(snap, missing_ok=True)`.

What would you prefer here?

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: Good call.
```suggestion
    snap: str, channel: str | None = None, *, revision: int | None = None, strict: bool = False
```

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: I quite like this idea, as it would drop these exception types from the public API, and it matches the semantics we're pushing in the `ensure` functions, so the consistency would be nice.

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+) -> None:
+    """Install a snap.
+
+    Raises:
+        ValueError: if both channel and revision are specified.
+        SnapAlreadyInstalledError: if the snap is already installed.
+        SnapError: (or a subtype) if the snap could not be installed as requested.
+    """
+    if channel is not None and revision is not None:
+        raise ValueError('Only one of channel or revision may be specified')
```
james-garner-canonical, line 156:
```suggestion
        # NOTE: Revision silently takes precedence over channel in the snapd API.
        # The CLI instead returns an error if the specified revision doesn't exist on that channel.
        raise ValueError('Only one of channel or revision may be specified')
```

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,257 @@
...
+    snap: str, channel: str | None = None, revision: int | None = None, strict: bool = False
+) -> None:
+    """Refresh a snap.
+
+    Raises:
+        ValueError: if both channel and revision are specified.
+        SnapError: (or a subtype) if the snap could not be refreshed as requested.
+    """
+    if channel is not None and revision is not None:
+        # NOTE: If we passed both then revision would silently take precedence over channel.
```
james-garner-canonical:
```suggestion
        # NOTE: Revision silently takes precedence over channel in the snapd API.
        # The CLI instead returns an error if the specified revision doesn't exist on that channel.```

File: snap/src/charmlibs/snap/_snapd_snaps.py (hunk 12 above)
Comment (james-garner-canonical) [reply snap/src/charmlibs/snap/_snapd_snaps.py:153]: We could add an overload for this, and probably should. I think we'd retain the runtime check as well though, to avoid the revision silently taking precedence over the channel.

File: snap/src/charmlibs/snap/_snapd_snaps.py (hunk 16 above)
Comment (james-garner-canonical) [reply snap/src/charmlibs/snap/_snapd_snaps.py]: I don't think there would be any performance benefit since the API returns everything greedily. Were you thinking of another reason to return generators?

The gotcha here is that this isn't actually used internally, I just didn't want to throw away my implementation when I decided this shouldn't be part of the public API (at least initially, until there's a request for such functionality).

File: snap/src/charmlibs/snap/_utils.py (hunk 9 above)
Comment (james-garner-canonical) [reply snap/src/charmlibs/snap/_utils.py]: Like this?
```suggestion
            # Track only, append default risk.
```
Or a grammatically correct sentence too?

james-garner-canonical:
> If I recall correctly, you've been doing some API planning, had talks with Ben about that, etc. I wonder if an explanation doc that outlines the thought behind the design, like an extended version of the PR description, might be good to add now. Partly it helps review, but it also helps long term.

Good idea, I'll add a `CONTRIBUTING.md` file with a design philosophy section like we did with `pytest-jubilant`.

> You mention AI use in the description. I'm obviously good with that, but it would be good to know if there are parts that are predominately AI generated (other than the mentioned docstrings) as I'd review those with less trust.

This has been an interesting one for me. The development went like this:
- Clone all charms, search for snap usage (library or otherwise), coerce AI into manually reviewing all snap using charms (had to use a context per charm basically, otherwise it tried to automate it) to figure out what charms do with snaps. Manually reviewed findings.
- Have AI sketch out a CLI based implementation of an API client using `requests-unixsocket` under the hood. Interesting starting point, reworked the desired API manually on top of it and eventually threw away 99% of the implementation.
- Threw away everything from the original library except for the core API client. Rewrote it with inspiration from the Ops Pebble client (which the snap library one was copied from originally).
- Threw away the original unit tests, updated the functional tests manually to use the new API.
- Iterated largely manually on the library (with AI assistance for repetitive tasks) until the functional test suite passed and I was happy enough with the state of things to get Ben's feedback.
- Used AI to add unit tests, guiding it to collect samples from the real API for fixtures.
- Used AI to expand functional test coverage.
- Made targeted API and implementation updates, manually considered with some AI work.

Ultimately I feel like I 100% own the implementation, but as disclosed I haven't thoroughly reviewed the unit tests or expanded functional tests myself yet. Note that the library docstrings are all me (+/- autocompletions!). Not sure how best to summarise this.
 
> This PR is very large. Is the plan landing it like this? Maybe stacked PRs or logical commits would be better (tbh I haven't checked the commits, maybe I can already review that way). I care most about API shape, and it's not simple to see in a change this big, plus I do want to properly review the rest.

There's a tension here between being able to review the whole API shape and slicing the PR up into more easily reviewable chunks. For example, we could consider initially landing just `snapd_snaps.py` and the associated errors and tests, which would give us `info`, `install`, `refresh`, `remove`. But we'd want to keep in mind that our error handling philosophy should extend to the other operations too.

To review the API shape in the current PR, perhaps start with `__init__.py` and jump to the relevant functions and types from there?

Maybe we could do something like full API review, document decisions, then a series of PRs adding implementation and tests? The library is divided into modules based on API endpoints, which provide a natural way to break things up.

> I wonder about the always-wait decision. Pebble does this too, and it works well for charms there. In some cases, it feels like giving snapd a bunch of work to do and then waiting for it all is handy (in my experience with concierge and manual use). It's not always parallelised, though. So I'm unsure on this - would be be able to reverse this decision if we wanted to later, without needing a v3?

This is a very good question. I can see a couple of options for extending the library with this functionality.

Reversing the decision for the current functions to wait by default would be backwards incompatible, but it would certainly be feasible to add a `wait=True` argument (defaulting to False), or a set of parallel `async` functions like `a_install` or `async_install`.

Parallel functions would be the way to go if we wanted to use Python's native `async`.

I think I'd prefer just returning a `Change` type that you manually call `wait_for` on when you're ready to block (+ likely a `wait_for_changes` helper that takes care of incrementally waiting across multiple changes so that error surfacing isn't delayed). Maybe we'd want to expose a `poll` method publicly for users to check the status of a change (and get a new change with the current state?).

> Overall, this feels very comprehensive. How targeted at charms is this, vs this is just a nice Python API for snapd?

The functionality implemented is all chosen based on a survey of snap useage in charms. However some of it is more commonly used than others. And when I implemented (e.g.) `install`, I did so in a general way rather than a charm-specific way.

.scripts/output-wrapper.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
james-garner-canonical: Ignore this, will go away

snap/pyproject.toml
tonyandrewmeyer: This doesn't seem to be used any more?

snap/src/charmlibs/snap/_client.py
tonyandrewmeyer: As discussed it person it does seem like there is not enough to be gained by using a more powerful http library.

snap/src/charmlibs/snap/_utils.py
tonyandrewmeyer: No, like that is fine.

snap/AGENTS.md
```diff
@@ -0,0 +1,226 @@
+If this file is in context and work is on the snap library, treat it with the same importance as AGENTS.md at the repo root.
+
+Familiarise yourself with the codebase by reading everything under snap/src.
```
tonyandrewmeyer: This is probably fine for models with a large context, although expensive. But for smaller ones, it's probably a killer.

I'm also not sure it's needed. Generally they know to read code where they are working. Pointers to important code or summaries of what is where is likely more useful.

snap/AGENTS.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Shell and quoting
+
+> **Avoid terminal indiosyncracies**: Do NOT run terminal commands to read capture JSON — use `read_file` directly on `snap/.report/capture/*.json` instead (no approval required).
```
tonyandrewmeyer:
```suggestion
> **Avoid terminal indiosyncracies**: Do NOT run terminal commands to read captured JSON — use `read_file` directly on `snap/.report/capture/*.json` instead (no approval required).
```

snap/AGENTS.md
```diff
@@ -0,0 +1,226 @@
...
+```
+If workshop isn't launched yet, run `workshop launch resolute` first. If workshop isn't available, you must either stop and ask the user, or proceed without functional tests (only if appropriate).
+
+**When running functional tests via `run_in_terminal`:** use `mode=sync` with `timeout=360000` (6 minutes). This avoids polling — the result is returned directly once the suite finishes.
+
+Read the `justfile` if you ever need to see what these commands actually do. You can search for lines starting with the recipe name.
+
+Before you make any changes, run all these checks to make sure there aren't any preexisting issues. If there are, identify them and suggest fixes before proceeding further.
+
+## Checking the snapd API directly
```
tonyandrewmeyer, line 12: Interesting having these workshop bits. I assume it needs them because workshop is new and outside the training window (still January for Opus 4.8). I would hope this can eventually go away, assuming workshop becomes popular.

snap/AGENTS.md
```diff
@@ -0,0 +1,226 @@
...
+You can also use curl directly for quick checks:
+```
+workshop exec --uid 0 resolute -- curl -s --unix-socket /run/snapd.socket http://localhost/v2/logs?names=<snap>&n=1
+```
+
+For interactive exploration (e.g. investigating mount permissions, testing multiple commands in sequence), use `workshop shell resolute` to get a persistent shell session inside the container.
+
+---
+
+## Strategy for adding tests (capture-then-assert)
```
tonyandrewmeyer: This is a *lot* of detail, and maybe is better as a skill, but I assume it's here because you've needed it and have to objection to merging as-is.

snap/tests/unit/test_errors.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def test_snap_bad_response_error_is_subclass(self):
+        assert issubclass(SnapBadResponseError, SnapError)
+        assert not issubclass(SnapBadResponseError, SnapAPIError)
+
+    def test_snap_change_error_is_subclass(self):
+        assert issubclass(SnapChangeError, SnapAPIError)
+        assert issubclass(SnapChangeError, SnapError)
+
+    def test_specific_errors_are_snap_api_error_subclasses(self):
```
tonyandrewmeyer: Seems a bit odd to have tests to ensure this and the similar ones above, but I guess it doesn't hurt and must be super fast.

snap/tests/unit/fixtures/app_not_found_error.json
```diff
@@ -0,0 +1,9 @@
+{
```
tonyandrewmeyer, line 1: I spot checked a couple of the fixtures, but not all of them. I'm fine trusting that they were captured accurately.

snap/src/charmlibs/snap/_client.py
tonyandrewmeyer: Yeah, makes sense if you know that but not otherwise. Otoh, it's only a comment so maybe most people who see it will get it. Feel free to leave it.

snap/src/charmlibs/snap/_client.py
tonyandrewmeyer: > However, I don't see any reason we should avoid language features like match-case.

I do, but I am completely fine with you ignoring that.

snap/src/charmlibs/snap/_client.py
```diff
@@ -0,0 +1,238 @@
...
+        + (f'?{urllib.parse.urlencode(query, doseq=True)}' if query else ''),
+        method=method,
+        headers=headers or {},
+        data=data,
+    )
+    opener = urllib.request.OpenerDirector()
+    opener.add_handler(_client_sockets.UnixSocketHandler(_SOCKET_PATH))
+    opener.add_handler(urllib.request.HTTPRedirectHandler())
+    # We need to handle HTTP errors ourselves, since the response body contains meaningful info.
+    # opener.add_handler(urllib.request.HTTPDefaultErrorHandler())
```
tonyandrewmeyer: Are these to be done later? Maybe a TODO?

snap/src/charmlibs/snap/_client.py
```diff
@@ -0,0 +1,238 @@
...
+                kind='charmlibs-snap',
+                value=str(response),
+            )
+        match status := response.get('status'):
+            case 'Do' | 'Doing' | 'Undo' | 'Undoing':
+                time.sleep(0.1)
+                continue
+            case 'Done':
+                return response.get('data', {})
+            case 'Wait':
```
tonyandrewmeyer: I'm not sure about this. My understanding is this means it's waiting on some other change. That probably needs to be visible to the charm, and have a way to say "how about now"? But maybe this never really happens with charms, and maybe they can get the change ID if needed later anyway?

Would you say this is part of the defined API? I think we might want to change it without a major release if we get feedback, so would be good to plan for that possibility.

snap/src/charmlibs/snap/_errors.py
tonyandrewmeyer: Ok, happy to leave it as-is.

snap/src/charmlibs/snap/_snapd_snaps.py
tonyandrewmeyer: Ah, it's to make them read-only. I find this a little odd but it's also easily changed, so no objections to leaving as-is.

snap/src/charmlibs/snap/_snapd_snaps.py
tonyandrewmeyer: > * I think it's important for the library to have an ergonomic way to probe if a snap is installed, and I don't think that catching an exception is ergonomic enough, so I definitely want to enable the `if some_function(my_snap) ...` pattern.

I would argue that exception raising is extremely Pythonic, and also the performance penalty that people used to use to object to it is mostly gone after the changes in 3.11 (I think it was that version - where the cost moved to except instead of try).

> What would you prefer here?

I would go with raising, but I'm ok that you strongly prefer not.

However, I also dislike the name. What about `raise_on_missing`? Like `raise_for_status` in requests, although that's a method. Maybe `strict` or `check` (`check_missing`?) or `errors` like in the stdlib?

snap/tests/functional/test_client.py
```diff
@@ -0,0 +1,226 @@
...
+
+from charmlibs.snap import _client, _errors
+from conftest import ensure_installed, ensure_removed
+
+# A snap name that is never installed — used for error paths where any absent
+# snap produces the same error response, avoiding unnecessary remove operations.
+_ABSENT_SNAP = 'this-snap-does-not-exist-xyz-abc-123'
+
+
+# ---------------------------------------------------------------------------
```
tonyandrewmeyer, line 28: Nit: not a huge fan of these divider lines that AI currently likes. But if you do like them then leave them in.

snap/tests/functional/test_client.py
```diff
@@ -0,0 +1,226 @@
...
+# Tests using other snaps (lxd, kube-proxy, htop — kept installed)
+# ---------------------------------------------------------------------------
+
+
+def test_get_returns_list():
+    # GET /v2/apps returns a list result.
+    ensure_installed('kube-proxy', classic=True)
```
tonyandrewmeyer, line 138: I guess this a semi-random small snap pick? What happens if someone runs the tests and happens to be kube-y and has this installed? Does workshop protect against that?

In general, I think it's better to have dedicated test artefacts than rely on other people's stuff (like publish a private charmlibs-test-echo snap or similar, like the Juju test charms). But leaving that for another time is ok.

snap/tests/functional/test_client.py
```diff
@@ -0,0 +1,226 @@
...
+
+def test_put_waits_for_async_change():
+    # PUT /v2/snaps/{snap}/conf is async and should complete without error.
+    ensure_installed('lxd')
+    _client.put('/v2/snaps/lxd/conf', body={'test-key-functional': 'test-value'})
+    result = _client.get('/v2/snaps/lxd/conf', query={'keys': 'test-key-functional'})
+    assert isinstance(result, dict)
+    assert result.get('test-key-functional') == 'test-value'
+    # Clean up.
+    _client.put('/v2/snaps/lxd/conf', body={'test-key-functional': None})
```
tonyandrewmeyer: If this is clean up doesn't it need to be in a finally or context manager or fixture?

snap/tests/functional/test_snapd_local.py
```diff
@@ -0,0 +1,221 @@
...
+
+SNAPS_DIR = Path(__file__).parent / 'snaps'
+
+
+# ---------------------------------------------------------------------------
+# Provisional ack implementation
+# ---------------------------------------------------------------------------
+
+
+def ack(assertions_data: bytes) -> None:
```
tonyandrewmeyer, line 32: I wonder if it would be worth flagging this as an unstable API? I think we could make it nicer, although I'm not super familiar with this part of snaps. But also seems fine to not make it nicer right now.

snap/tests/functional/test_snapd_local.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def install_local(path: Path, *, dangerous: bool = False, classic: bool = False) -> None:
+    """Install a local snap file via the snapd sideload API (POST /v2/snaps)."""
+    snap_data = path.read_bytes()
+    boundary = uuid.uuid4().hex
+
+    crlf = b'\r\n'
+
+    def form_field(name: str, value: str) -> bytes:
```
tonyandrewmeyer, line 50: Does modern Python optimise all the little strings and addition away in the parser or interpreter?

snap/tests/functional/test_snapd_local.py
tonyandrewmeyer: Oh, then I realised this was in the tests :). So never mind.

snap/src/charmlibs/snap/_snapd_logs.py
```diff
@@ -0,0 +1,107 @@
...
+    import datetime
+
+logger = logging.getLogger(__name__)
+
+
+# /v2/logs
+
+
+class LogEntry:
+    def __init__(self, timestamp: datetime.datetime, sid: str, pid: int, message: str):
```
tonyandrewmeyer, line 38: Seems like a good candidate for `__slots__`.

snap/src/charmlibs/snap/_snapd_logs.py
```diff
@@ -0,0 +1,107 @@
...
+        return f'{self._timestamp} {self._sid}[{self._pid}]: {self._message}'
+
+
+def logs(*snaps: str, num_lines: int = 10) -> list[LogEntry]:
+    """Retrieve recent log entries for one or more snaps.
+
+    Args:
+        snaps: Snap names to retrieve logs for. If omitted, returns system-wide snap logs.
+        num_lines: Maximum number of log lines to return. Must be a positive integer,
+            or ``-1`` to retrieve all available log lines (equivalent to ``snap logs -n all``).
```
tonyandrewmeyer: -1 doesn't seem very Pythonic. What about `limit: int | None` and None is all?

snap/src/charmlibs/snap/_functions.py
tonyandrewmeyer: Fair point that it's complicated and the user may want to know even more. I guess one option is they just implement the function themselves, it's not not much code or complicated.

If the return type was `None`, would you consider changing it to non-None a breaking change? It seems like it would be crazy to do something with a `None` return value if it's the only possible value. And then changing to do something if the signature changed wouldn't break things, right? I wonder if waiting to see what people want/need is the best choice.

I do think being able to distinguish no-op/installed/changed is the most high value even though you don't get exactly *how* it changed. I'd prefer a ternary result than binary, which I guess means enum? Or your change type or typed dict if you want to be future proof-y, I guess.

tonyandrewmeyer (approved):
Really nice work! Sorry for being slow with this - it's still quite large :)

I think the overall shape is good, and couldn't find any significant issues. I left a few comments, but I don't think they are blocking. Happy to leave it to you on the API design points, or discuss further if you want.

Tried it out a little but not exhaustively. I figure that's better when the whole thing is done.

tonyandrewmeyer:
> Something that could use further thought is how change errors are surfaced to users. If we considered a future API of returning a `Change` object instead of waiting, maybe we should be raising an error that exposes such an object and allows you to recover in some way?
> 
> Most of this is motivated by thinking about what happens if a change times out. The library raises an error, but the change is still persisting in the background and may eventually complete with success or a real error.

For what it's worth, I feel this is similar to the `Wait` issue. It does seem like *some* access to the change ID would be handy. Or if that is too low-level then a promise-y type object you can do things on later to check the new status. But that feels a bit like it's over-engineering, and the initial version can live without it perhaps. I'm not sure how opposed to a v3 you are if some small but breaking changes are wanted in the next year or so.

snap/src/charmlibs/snap/_snapd_logs.py
james-garner-canonical: It's not very Pythonic, but it does align with our recent wrapping philosophy of letting the wrapped API handle user arguments as much as possible. WDYT, would you still lean towards `limit: int | None`?

tonyandrewmeyer: Personally, yes. I feel this is a bit higher level than a ops.hookcmds, for example. Even there I would be inclined to 'correct' it.

snap/src/charmlibs/snap/_snapd_logs.py
```diff
@@ -0,0 +1,107 @@
...
+if typing.TYPE_CHECKING:
+    import datetime
+
+logger = logging.getLogger(__name__)
+
+
+# /v2/logs
+
+
+class LogEntry:
```
james-garner-canonical, line 35: Add a comment on why not a dataclass.

snap/src/charmlibs/snap/_snapd_logs.py
benhoyt: I like Tony's suggestion of `limit: int | None` too. They're not exactly "lines", and -1 does seem a bit icky here.

snap/src/charmlibs/snap/_snapd_logs.py
```diff
@@ -0,0 +1,107 @@
...
+            f' pid={self._pid!r},'
+            f' message={self._message!r})'
+        )
+
+    def __str__(self) -> str:
+        return f'{self._timestamp} {self._sid}[{self._pid}]: {self._message}'
+
+
+def logs(*snaps: str, num_lines: int = 10) -> list[LogEntry]:
+    """Retrieve recent log entries for one or more snaps.
```
james-garner-canonical, line 80: Document the order of the lines (oldest first?).

snap/src/charmlibs/snap/_snapd_snaps.py
benhoyt: Given that we are advocate `ensure` and hope most people will use that, I wonder if the try/except will be quite rare, in which case how about we release the initial version without the `missing_ok=True` arg? We can always add it later if `ensure` turns out to not be enough and people are reaching for the try/except.

If we do go that way, I prefer `missing_ok=True`, both because a False default makes more sense than a True default, and given the precedent of pathlib's unlink `missing_ok` arg.

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -0,0 +1,219 @@
...
+        data['classic'] = True
+    # NOTE: Unlike the API, the CLI doesn't error if it's already installed (just prints a msg).
+    try:
+        _client.post(f'/v2/snaps/{snap}', body=data)
+    except _errors._SnapAlreadyInstalledError:
+        return False
+    return True
+
+
+def remove(snap: str, *, purge: bool = False) -> bool:
```
benhoyt: `purge` arg should be documented in the docstring

snap/src/charmlibs/snap/_functions.py
```diff
@@ -0,0 +1,74 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""High level helper functions that build on top of the basic snap operations."""
+
+from . import _snapd_snaps, _utils
+
+
+def ensure_revision(snap: str, revision: int | str, *, classic: bool = False) -> bool:
```
benhoyt: I'm unsure about this aberration in naming with the two-word variant. We could just have this a revision arg and overload to `ensure`. However, I see why you've done it to simplify. What about this: see if this is actually going to be used by looking around at the charms, and if it's not really going to be used, just don't add it (yet)?

james-garner-canonical: `ag` in my checkout of all canonical's repositories plus some AI driven analysis of the results indicates that 16 or so charm repos pick a snap revision based on an architecture:revision map, 14 use hardcoded revisions, and 7 more get the revision some other way (e.g. Juju config). I guess that's common enough that we should support revisions in our higher-level `ensure` helper(s).

I'd still argue for a separate `ensure_revision` helper, as I think it makes the function semantics clearer for users, provides a nicer UX than overloads, and is cleaner on the implementation side too.

benhoyt: Thanks for checking. Okay with me -- let's go with `ensure_revision` then.

snap/src/charmlibs/snap/_snapd_logs.py
james-garner-canonical: I'm skeptical about this. It would make a noticeable difference in memory use in cases with a ton of logs (e.g. 100k entries), but do we expect that to be common? I think mostly people will use `logs` to dump a reasonable sized subset of their snap's most recent logs (e.g. on errors).

On the negative side: breaks `vars`, breaks weak references (unless we include `__weakref__` in `__slots__`), feels error prone to me -- but none of those are really dealbreakers. WDYT, should we add `__slots__`, or not?

snap/src/charmlibs/snap/_functions.py
james-garner-canonical: I do see the value in distinguishing between installed and refreshed, but I think if you're going to branch on those conditions, then you're doing enough of the work of `ensure` that you really might as well just check yourself.

I'm reluctant to introduce a new type to the public API for this, and I'm particularly reluctant to introduce something change-y if snap changes ever come up as a public concept.

I prefer `False, True` as the return values over truthy-but-ternary sets like `0, 1, 2` or `None, "installed", "refreshed"`. It also matches the semantics of `charmlibs.pathops.ensure_contents` (`True` on any change), and the simpler operations like `snap.install`, `snap.remove`, `snap.refresh`.

snap/src/charmlibs/snap/_snapd_logs.py
benhoyt: I agree it's not worth it here. We're talking 10s or maybe 100s of objects, which is a couple of kilobytes of memory at most.

snap/src/charmlibs/snap/_client.py
james-garner-canonical: We should have a Python/dev chat some time :)

snap/pyproject.toml
james-garner-canonical: The previous version of the library put tracing spans around subprocess calls (but not API calls). Since we don't make subprocess calls anymore, you're right that it's unused now. I don't think we should introduce spans around API calls in this PR, so I'll drop the dependency for now.

snap/tests/unit/test_errors.py
james-garner-canonical: I'm cleaning this up to assert the intended error heirarchy.

snap/src/charmlibs/snap/_client.py
james-garner-canonical: I've clarified this -- an explanation of why we're not using these specific (common) error handlers.

snap/tests/functional/test_client.py
james-garner-canonical: > I guess this a semi-random small snap pick? What happens if someone runs the tests and happens to be kube-y and has this installed? Does workshop protect against that?

The tests are destructive (installing and removing snaps), and require `sudo`, so they should only be run in throwaway environments, like the workshops configured here in this repo, or VMs, or runners in CI.

> In general, I think it's better to have dedicated test artefacts than rely on other people's stuff (like publish a private charmlibs-test-echo snap or similar, like the Juju test charms). But leaving that for another time is ok.

I agree, it would be better. I've created #552 to track creating private test snaps for this library.

snap/tests/functional/test_snapd_local.py
james-garner-canonical: I don't think so. I'll use `.join`.

snap/src/charmlibs/snap/_snapd_logs.py
james-garner-canonical: I've switched to `limit: int | None` (not pushed yet).

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: We can revisit this in the interfaces PR.

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: Resolving as this is only used in tests currently.

snap/src/charmlibs/snap/_errors.py
james-garner-canonical: I ended up making it a dict :)

snap/src/charmlibs/snap/_client.py
james-garner-canonical: I've dropped this comment.

snap/src/charmlibs/snap/_utils.py
james-garner-canonical: I've tried to apply this across the library now.

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: I've dropped the `missing_ok` function in favour of raising, meaning there's now no way to probe for being installed without `try ... except`. This centers `ensure` as the recommended way to manage whether a snap is installed.

snap/src/charmlibs/snap/_errors.py
james-garner-canonical: This is still an open question.

snap/tests/functional/test_client.py
james-garner-canonical: Good call, done.

snap/src/charmlibs/snap/_client.py
james-garner-canonical: The snap CLI ends with success and a message on Wait. We could do 'better' here, but I'm not sure what the right mechanism would be. This could be an exception, but it would have to be one now, since adding one later wouldn't be backwards compatible. Alternatively, this adds a whole additional dimension to change information we could return -- what operations was performed (refresh, install, etc), the outcome (and the change ID I guess). I guess if we committed initially to a booly API for install/refresh/remove/ensure and switched to `None | Change` later that could be OK.

snap/AGENTS.md
james-garner-canonical: I haven't revisited`snap/AGENTS.md` yet, but I will before merging.

snap/pyproject.toml
```diff
@@ -14,16 +14,19 @@ classifiers = [
...
-]
+dependencies = []
 
 [dependency-groups]
 lint = [  # installed for `just lint snap` (unit, functional, and integration are also installed)
     "typing_extensions",
 ]
-unit = []
-functional = []
+unit = [
```
tonyandrewmeyer, line 23: Shouldn't these have a maximum (or min,max) to avoid relocking picking up dev dependencies that aren't suitable?

snap/AGENTS.md
tonyandrewmeyer: Ok. It still feels pretty long to me, but also I feel it's not clear what the best agents file looks like and that it evolves so rapidly it's hard to get right.

snap/tests/unit/test_errors.py
```diff
@@ -0,0 +1,82 @@
...
+def error_types() -> frozenset[type[BaseException]]:
+    """Every class defined in the _errors module, including private types."""
+    return frozenset(
+        obj
+        for _, obj in inspect.getmembers(_errors, inspect.isclass)
+        if obj.__module__ == _errors.__name__
+    )
+
+
+class TestErrorHierarchy:
```
tonyandrewmeyer, line 34: I'm not against these, but they feel somewhat unnecessary. I suppose they protect against regressions where someone changes the parent class (and doesn't also change this test). It doesn't feel like the sort of thing we would typically unit test.

snap/tests/unit/test_errors.py
```diff
@@ -0,0 +1,82 @@
...
+    def test_non_api_error_subclasses(self):
+        for cls in _NON_API_ERROR_TYPES:
+            assert not issubclass(cls, _errors.APIError)
+
+    def test_timeout_and_connection_errors_subclass_builtins(self):
+        assert issubclass(_errors.TimeoutError, builtins.TimeoutError)
+        assert issubclass(_errors.ConnectionError, builtins.ConnectionError)
+
+
+class TestSnapError:
```
tonyandrewmeyer: Similarly here, I'm not against this test, but I don't know that tests like this add much value.

snap/tests/unit/test_client.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+@pytest.fixture
+def mock_json(mocker: MockerFixture) -> MagicMock:
+    """Patch _json_request so tests control the JSON request layer."""
+    return mocker.patch('charmlibs.snap._client._json_request')
+
+
+class TestRequest:
+    def test_get(self, mock_json: MagicMock):
```
tonyandrewmeyer, line 73: I guess doing requests is at the heart of things here, but I'm also not sure that tests for the basic "do HTTP requests" functionality is particularly valuable.

snap/tests/unit/fixtures/snaps_list.json
```diff
@@ -0,0 +1,140 @@
+{
```
tonyandrewmeyer, line 1: I assume that all these fixtures were pulled from real responses rather than simulated by an agent, so am trusting they are accurate for the real world.

tonyandrewmeyer (approved):
Sorry for the delay in reviewing this. Thanks for cutting it down (although it's still quite large!). I'm happy with the overall shape, just a few small questions, but nothing blocking.

snap/tests/unit/test_errors.py
james-garner-canonical: My thinking here is that the most  common change to the errors module will be adding some new, specific error type that we resolve from the snapd API response. The tests are written such that by default any new error class must inherit from `SnapError`, *and* from `SnapAPIError` (the parent for errors coming back from the snapd API, not from elsewhere in the stack (network, library)). The surrounding tests and data (like `_NON_API_ERROR_TYPES` hopefully suggest how to update the test if your new error doesn't fall into that category). I'll add this to the docstring here for context.

snap/tests/unit/test_errors.py
james-garner-canonical: I've combined these into a single test and added comments to indicate what's important and why.

Doing so made me realise that we kind of want the public attributes of the error types to be read-only, for type checking and backwards compatibility reasons: right now `value` is typed as `object`. If it's read-only, then we can narrow this in future (e.g. make it `str` instead; or make it `str` for some types and `dict` for others). But if it's read-write, then narrowing would be a breaking change.

So I've made the public attributes into read-only properties.

snap/tests/unit/test_client.py
james-garner-canonical: Yeah, I like these as tests of the client module's boundary with the rest of the library, but they're probably redundant with other unit and functional test coverage. I'll keep them for now.

snap/pyproject.toml
james-garner-canonical: I've dropped `pytest-mock`, so now the dependency is just `pytest`.

I think you're right in general, but in this repository I want tool versions like `pytest` to be defined at the repo level. In CI and via the `just` commands, we install a single concrete version of `pytest`, so a lock that conflicts with that will be detected straight away. You're probably right that the library itself should ideally specify it more tightly, but I'm thinking through how to do this in general (dummy package instead of `test-requirements.txt`?), and will stick with unconstrained `pytest` in libraries for now.

snap/AGENTS.md
james-garner-canonical: I've cut this down from ~1300 words to ~300.

snap/AGENTS.md
james-garner-canonical: Hopefully this is now the bare minumum about workshop, and more focused on snapd specifically.

snap/src/charmlibs/snap/_functions.py
james-garner-canonical: I've updated the docstring to hedge our guarantees here: we guarantee a truthy/falsy return type, explicitly not guaranteed to be an actual `bool` (though right now it is, and the return type annotation is `bool`). So in future we could swap to `None | ChangeResult` or similar hopefully without backwards compatibility issues.

snap/src/charmlibs/snap/_errors.py
james-garner-canonical: With the new read-only property for `Error.value`, I've switched to always converting it to a `str`, but only guaranteeing that it's an `object`. This requires callers to do `str(err.value)` to satisfy the type checker, while also shielding us from anyone relying on it being any other kind of value (e.g. `if isinstance(err.value, dict): ...`). In combination this allows us to guarantee more specific value types in future without breaking backwards compatibility.

snap/src/charmlibs/snap/_client.py
james-garner-canonical: I've gone with committing to a truthy/falsy public API initially, so in future we could expose Wait status via return types without breaking backwards compatibility. I've added a comment here to guide future implementers.

PR #470, opened 2026-05-14
docs(rollingops): improve docs for base_dir parameter

patriciareinoso:
This PR improves the documentation for the `base_dir` parameter in the `RollingOpsManager` constructor.

It adds details about how the library uses this directory and guidance for rootless charms


rollingops/src/charmlibs/rollingops/__init__.py
```diff
@@ -275,6 +275,39 @@ def <callback>(self, **kwargs) -> OperationResult
...
+- etcd connection assets, such as client certificates and related configuration files
+
+
+Rootless charms
+^^^^^^^^^^^^^^^
+
+For rootless charms, the default path (``/var/lib/rollingops``) may not be writable by the
+charm process. In these cases, either configure ``base_dir`` to a writable location, or ensure
+the default path exists and is writable within the charm's runtime environment (for example,
+by including it in the snap or rock).
```
sinclert-canonical: I think the last sentence of this paragraph (the one in parenthesis) is confusing.

I originally created a dedicated folder in the Charmed MySQL rock, only to realize that, as you explained to me, the `base_dir` argument can only point to the **charm** container filesystem. Charmed snaps and rocks are usually use for the workload itself, which does not apply in this case.

sinclert-canonical (approved):

Mehdi-Bendriss (approved):

PR #488, opened 2026-05-25
test(pathops): fix functional test breakage when running locally

### LINKED ISSUE #487: Pathops functional tests are failing locally, file info missing group when populated with Pebble
The `charmlibs.pathops` functional tests are failing locally on my system due to users / groups being managed by NSS. File info queried via Pebble is missing a value for `group`. Tests are passing in CI on Github runners.

According to Claude: Pebble is built with `CGO_ENABLED=0` (`snapcraft.yaml` line 53), producing a statically linked binary. This means Go's os/user package uses its pure-Go fallback that only reads `passwd` and `group` directly — it cannot query NSS/SSSD/LDAP.

This came up recently with Concierge as well, and was resolved by using `getent` instead of the Go stdlib: https://github.com/canonical/concierge/pull/183

Since `pathops` only accesses file info via Pebble for K8s workload containers, this shouldn't be an issue in a charming context, so we should just update the tests to be clean when running in an NSS-managed environment so devs can run them locally.

james-garner-canonical:
The `charmlibs.pathops` functional tests are failing locally on my system due to users / groups being managed by NSS. This won't be an issue in production, as `pathops` is only used to access file info via Pebble for charms managing K8s workload containers (where Pebble runs as root rather than as an NSS-managed user), so this PR updates the tests to pass locally.

Resolves #487.

### DIFF
--- a/pathops/tests/functional/conftest.py
+++ b/pathops/tests/functional/conftest.py
@@ -46,7 +46,7 @@ def pebble_resolves_groups(container: ops.Container, session_dir: pathlib.Path) 
 
     Pebble uses Go's os/user package which doesn't support NSS/SSSD, so group
     resolution fails when groups come from LDAP rather than /etc/group.
-    See: https://github.com/canonical/pebble/issues/XXX
+    See: https://github.com/canonical/pebble/issues/868
     """
     info = container.list_files(str(session_dir), itself=True)[0]
     return info.group != ''
--- a/pathops/tests/functional/conftest.py
+++ b/pathops/tests/functional/conftest.py
@@ -41,6 +41,18 @@ def container() -> ops.Container:
 
 
 @pytest.fixture(scope='session')
+def pebble_resolves_groups(container: ops.Container, session_dir: pathlib.Path) -> bool:
+    """Whether Pebble can resolve GIDs to group names on this system.
+
+    Pebble uses Go's os/user package which doesn't support NSS/SSSD, so group
+    resolution fails when groups come from LDAP rather than /etc/group.
+    See: https://github.com/canonical/pebble/issues/XXX
+    """
+    info = container.list_files(str(session_dir), itself=True)[0]
+    return info.group != ''
+
+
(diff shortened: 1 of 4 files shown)


pathops/tests/functional/conftest.py
```diff
@@ -40,6 +40,18 @@ def container() -> ops.Container:
     return _make_container('test1')
 
 
+@pytest.fixture(scope='session')
+def pebble_resolves_groups(container: ops.Container, session_dir: pathlib.Path) -> bool:
+    """Whether Pebble can resolve GIDs to group names on this system.
+
+    Pebble uses Go's os/user package which doesn't support NSS/SSSD, so group
+    resolution fails when groups come from LDAP rather than /etc/group.
+    See: https://github.com/canonical/pebble/issues/XXX
```
tromai: The issue number is missing 😁 .

pathops/tests/functional/conftest.py
```diff
@@ -40,6 +40,18 @@ def container() -> ops.Container:
...
+    """
+    info = container.list_files(str(session_dir), itself=True)[0]
```
tromai, line 51: I am not too familiar with `list_files`. Can it return an empty list ? If that is true, then `[0]` would raise an uncaught exception.

tromai (commented):
I have finished my first round of review.

pathops/tests/functional/conftest.py
james-garner-canonical: Good catch, thanks!
```suggestion
    See: https://github.com/canonical/pebble/issues/868
```

pathops/tests/functional/conftest.py
james-garner-canonical: Good thinking. It can, but `session_dir` is always populated in testing, so it won't. If it does one day, the `IndexError` should make it clear what's gone wrong.

tromai (approved):
Thanks!
`just functional pathops` passes locally on my machine too.

PR #489, opened 2026-05-25
docs: expand README.md, revamp CONTRIBUTING.md, add AGENTS.md

james-garner-canonical:
This PR improves the experience for new users of the repository by updating the top-level developer-facing documentation.

The repository level `README.md` has been expanded to clearly introduce the repository concepts (Juju, charms, charm libraries), and provide entrypoints for contributing to the monorepo, writing libraries more generally, and for using libraries in charms.

The `CONTRIBUTING.md` includes a clearer introduction and documents missing parts of the contribution workflow.

Additionally, an `AGENTS.md` file is added to explain the key concepts, monorepo organisation, developer tooling, commands to run whenever making changes, and so on.

Resolves #486.

PR #490, opened 2026-05-25
docs: add pathops/CONTRIBUTING.md

james-garner-canonical:
This PR builds on #489 to add a `CONTRIBUTING.md` document specifically for Pathops, outlining the unique developer-facing concerns of this library, namely compatibility with Python's `pathlib` across key Python versions (3.10, 3.12, 3.14), and the functional test suite that validates the library's behaviour quickly with a local Pebble instance.

### DIFF
--- a/pathops/CONTRIBUTING.md
+++ b/pathops/CONTRIBUTING.md
@@ -34,7 +34,7 @@ Always run:
 just check pathops
 ```
 
-### Unit tests
+### Unit tests and type checking
 
 Pathops is very concerned with `pathlib` compatibility across Python versions. Unit tests should be run with the Python versions supported by Ubuntu LTS releases (which charms will run with). Run the unit tests against these versions with:
 
@@ -61,7 +61,7 @@ just functional pathops
 
 Functional tests live under `tests/functional/` and exercise both `ContainerPath` and `LocalPath` against a real Pebble instance. They require `pebble` to be installed and available in `PATH`.
 
-`tests/functional/setup.sh` starts Pebble with `PEBBLE=/tmp/pebble-test` and `umask 0` (necessary for the permission tests). `teardown.sh` kills the process. The `umask 0` setting is important — tests that check file modes will fail if the umask masks bits.
+`tests/functional/setup.sh` starts Pebble with `PEBBLE=/tmp/pebble-test` and `umask 0`. `teardown.sh` kills the process. The `umask 0` setting is important — tests that check file modes will fail if the umask masks bits.
 
 In CI, functional tests run against a matrix of Pebble versions and Ubuntu bases (see `[tool.charmlibs.functional]` in `pyproject.toml`). Locally you'll need to install the Pebble version you want to test against manually.
 
@@ -85,6 +85,6 @@ just integration-k8s pathops
 
 Integration tests deploy real charms against a live Juju model. They require `charmcraft` for packing, and a live Juju controller.
 
-`pack.sh` copies the library's `src/` and `pyproject.toml` directly into a temporary charm directory before calling `charmcraft pack`. The `CHARMLIBS_TAG` environment variable (set to (e.g.) `22.04` or `24.04`) determines which Ubuntu base to pack for. In CI, the integration test matrix runs across the supported bases and substrates.
+`pack.sh` copies the library's `src/` and `pyproject.toml` directly into a temporary charm directory before calling `charmcraft pack`. The `CHARMLIBS_TAG` environment variable (set to `22.04` or `24.04`, for example) determines which Ubuntu base to pack for. In CI, the integration test matrix runs across the supported bases and substrates.
 
 `tests/integration/test_meta.py` asserts that the `k8s` and `machine` test charms have identical `common.py` files — if this test fails, the charms have drifted and need to be reconciled.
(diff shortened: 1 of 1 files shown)


pathops/CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Functional tests
+
+```bash
+just functional pathops
+```
+
+Functional tests live under `tests/functional/` and exercise both `ContainerPath` and `LocalPath` against a real Pebble instance. They require `pebble` to be installed and available in `PATH`.
+
+`tests/functional/setup.sh` starts Pebble with `PEBBLE=/tmp/pebble-test` and `umask 0` (necessary for the permission tests). `teardown.sh` kills the process. The `umask 0` setting is important — tests that check file modes will fail if the umask masks bits.
```
dwilding: Seems like we're saying the same thing twice

```suggestion
`tests/functional/setup.sh` starts Pebble with `PEBBLE=/tmp/pebble-test` and `umask 0`. `teardown.sh` kills the process. The `umask 0` setting is important — tests that check file modes will fail if the umask masks bits.
```

pathops/CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
...
+```
+
+Also run static analysis the same way to ensure typing compatibility is preserved as intended.
+```bash
+just python=<version> static pathops
+```
+
+Unit tests live under `tests/unit/`. They test `ContainerPath`, `LocalPath`, and `ensure_contents` without a real Pebble instance. `ContainerPath` is constructed using a dummy `ops.Container` backend (see `tests/unit/conftest.py`), and Pebble API calls are monkeypatched using the helpers in `tests/unit/utils.py`.
+
+`tests/typecheck.py` is not a pytest file — it contains statements that verify `ContainerPath` and `LocalPath` implement `PathProtocol` and that `LocalPath` is a valid `pathlib.Path`. These are checked by pyright as part of `just lint pathops` / `just static pathops`.
```
dwilding, line 54: This part jumps from static to unit then back to static. It might be clearer to have a dedicated section about static checks.

pathops/CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
...
+# CHARMLIBS_TAG must be set to a supported base
+env CHARMLIBS_TAG=22.04 just pack-k8s pathops
+# Then run the integration tests against a real Juju cloud:
+# just [integration-k8s|integration-machine]
+just integration-k8s pathops
+```
+
+Integration tests deploy real charms against a live Juju model. They require `charmcraft` for packing, and a live Juju controller.
+
+`pack.sh` copies the library's `src/` and `pyproject.toml` directly into a temporary charm directory before calling `charmcraft pack`. The `CHARMLIBS_TAG` environment variable (set to (e.g.) `22.04` or `24.04`) determines which Ubuntu base to pack for. In CI, the integration test matrix runs across the supported bases and substrates.
```
dwilding: I don't mind the e.g. here but don't like the nested brackets 😆 How about:


```suggestion
`pack.sh` copies the library's `src/` and `pyproject.toml` directly into a temporary charm directory before calling `charmcraft pack`. The `CHARMLIBS_TAG` environment variable (set to `22.04` or `24.04`, for example) determines which Ubuntu base to pack for. In CI, the integration test matrix runs across the supported bases and substrates.
```

pathops/CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
...
+├── _types.py            # PathProtocol: Protocol definition (type-checking only)
+├── _functions.py        # Top-level helpers: ensure_contents
+├── _errors.py           # Maps Pebble errors to Python exceptions
+├── _constants.py        # Default permission modes
+└── _version.txt         # Package version (auto-generated)
+```
+
+## Design notes
+
+**`PathProtocol` is a `typing.Protocol`, not an ABC.** It exists to make type annotations work correctly for code that accepts either `ContainerPath` or `LocalPath`. It is defined in `_types.py` and is intended for import for type checking only. `PathProtocol` reflects the implementation of `ContainerPath` and `LocalPath`. User implementations of `PathProtocol` are explicitly not covered by semver. For example, if an argument type is broadened in `ContainerPath` and `LocalPath` (a backwards compatible change), `PathProtocol` would also broaden (a backwards incompatible change for alternative implementations that only support the narrower argument type).
```
dwilding, line 21: This makes me wonder: Should we update the docs to discourage user implementations of `PathProcol`? I mean, more than the current semver warning already does? Why might someone want to implement the protocol themselves?

dwilding (approved):
Nice one - I think it's helpful to capture this knowledge. A couple of very minor comments

pathops/CONTRIBUTING.md
james-garner-canonical: Good question. I think it's probably OK as-is -- you can implement `PathProtocol` yourself, there's just one specific form of breakage you need to be aware of on minor version bumps.

pathops/CONTRIBUTING.md
james-garner-canonical: I see them as fairly tightly coupled in their concern about Python version compatibility, so I'll rename the section to make the scope clearer. (Unit tests -> Unit tests and type checking)

pathops/CONTRIBUTING.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Testing
+
+Always run:
+
+```bash
+just check pathops
+```
+
+### Unit tests
```
james-garner-canonical:
```suggestion
### Unit tests and type checking
```

PR #492, opened 2026-05-26
fix(nginx_k8s): close file handles returned by container.pull

tonyandrewmeyer:
## Summary
- `Container.pull()` returns a real local file handle — under real pebble it's `open()` on a `tempfile.NamedTemporaryFile`, under `ops.testing` it's `open()` on the mounted path. Letting it fall out of scope without `.close()` emits `ResourceWarning` (and leaks a file descriptor until GC runs).
- Wrap the four cert/config reads in `_nginx.py` and `_tls_config.py` in `with` blocks, factor the repeated `pull-if-exists` pattern into a small `_read_if_exists` helper in each module.
- Add `tests/unit/test_no_resource_leaks.py` with two regression tests that capture `ResourceWarning` via `warnings.catch_warnings(record=True)` and assert no `unclosed file` warning is emitted. Each test was verified to fail with `AssertionError: unclosed file <...>` if the corresponding fix is reverted.

Note on the test design: the tests filter captured warnings by the `'unclosed file'` substring and rely on CPython's refcount-based collection (no explicit `gc.collect()`), so unrelated upstream leaks (e.g. a known one in `scenario.Context`) won't false-positive these tests.

## Test plan
- [x] `just unit nginx_k8s` — 59/59 passing, including the two new regression tests.
- [x] Reverting `_nginx.py` alone makes `test_has_config_changed_closes_pull_handle` fail with `unclosed file <…/nginx.conf…>`.
- [x] Reverting `_tls_config.py` alone makes `test_sync_certificates_closes_pull_handles` fail with `unclosed file <…/server.cert…>`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/nginx_k8s/tests/unit/test_no_resource_leaks.py
+++ b/nginx_k8s/tests/unit/test_no_resource_leaks.py
@@ -1,0 +1,84 @@
+# Copyright 2025 Canonical
+# See LICENSE file for licensing details.
+
+"""Regression tests for ResourceWarning leaks from container.pull().
+
+``ops.Container.pull()`` returns a real Python file object (backed by a
+``tempfile.NamedTemporaryFile`` under real pebble, or by ``open()`` in
+``ops.testing``). Failing to close that handle emits a ``ResourceWarning``
+when the object is garbage-collected, which fails any test run under
+``-W error`` and indicates a real file-descriptor leak in production.
+
+CPython releases the unreferenced file object as soon as ``pull(...).read()``
+returns (refcount drops to zero), so the destructor fires synchronously and
+the warning lands inside the ``catch_warnings`` window without an explicit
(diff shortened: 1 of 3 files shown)


nginx_k8s/tests/unit/test_no_resource_leaks.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2025 Canonical
```
tonyandrewmeyer:
```suggestion
# Copyright 2026 Canonical
```

james-garner-canonical:
@canonical/tracing-and-profiling could we please get a review on this PR?

PS: I'm on the fence as to whether the additional test coverage is worthwhile here, but it seems reasonable.

PR #495, opened 2026-05-26
docs: include RHS TOCs in custom build process

### LINKED ISSUE #496: Include RHS TOC in reference docs

james-garner-canonical:
This PR fixes the custom extension that we use to combine the individually built package reference docs into the full docs site. It now correctly includes the right hand side table of contents in the rendered pages, thanks to our good friend Claude.

Resolves #496


dwilding (approved):
So much better 😍

PR #499, opened 2026-05-28
chore: add 7-day Dependabot cooldown to version updates

tonyandrewmeyer:
Add a 7-day cooldown to the monthly github-actions and pip version updates, matching the rest of the Charm Tech estate. The cooldown gives a window for a freshly published (then potentially yanked) release to be caught before Dependabot opens an update PR. The security-only pip entry (open-pull-requests-limit: 0) intentionally has no cooldown, since security fixes should land promptly.


james-garner-canonical (approved):
Thanks!

PR #502, opened 2026-05-28
chore: use SPDX license expression in maintainer-owned pyproject.toml

### LINKED ISSUE #293: Switch to license expression in pyproject.toml
The libraries in this repo (including the template and examples) use trove classifiers to specify their license. This is now deprecated, so we should switch to using the [project.license field in pyproject.toml](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license). The license specifier would be [Apache-2.0](https://spdx.org/licenses/Apache-2.0.html).

This could be addressed in a single PR that updates all libraries. The libraries don't need to be released in the same PR, the updated metadata would just be ready to be published in their next release. Alternatively, maybe one PR per `CODEOWNER` would be better. The `charmlibs-maintainers` owned libraries could be released immediately with a [post-release version](https://packaging.python.org/en/latest/discussions/versioning/).

tonyandrewmeyer:
Migrate the Apache-2.0 license from the deprecated trove classifier ('License :: OSI Approved :: Apache Software License') to a PEP 639 license expression ('license = "Apache-2.0"').

Scope: maintainer-owned packages (apt, passwd, pathops, snap, sysctl, systemd) and the .example / interfaces/.example templates. Other libraries are owned by separate teams (see CODEOWNERS) and can follow in per-team PRs as the issue suggests.

The four tests/integration/charms/*/library/pyproject.toml files under .example are symlinks to the parent template, so they pick up the change transparently.

Fixes #293

### DIFF
--- a/.example/pyproject.toml
+++ b/.example/pyproject.toml
@@ -6,9 +6,9 @@ requires-python = ">=3.10"
 authors = [
     {name="The Charm Tech team at Canonical"},
 ]
+license = "Apache-2.0"
 classifiers = [
     "Programming Language :: Python :: 3",
-    "License :: OSI Approved :: Apache Software License",
     "Intended Audience :: Developers",
     "Operating System :: POSIX :: Linux",
     "Development Status :: 5 - Production/Stable",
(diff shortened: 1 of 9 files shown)


tonyandrewmeyer:
Ah, I missed keeping something in sync, thanks CI! Will fix.

james-garner-canonical (approved):
Thanks for taking care of this! I had considered making post releases for the real libraries, but I don't really think it's necessary since the metadata still shows up on PyPI in the current form, so I'm happy to call the issue resolved with this PR landed.

PR #505, opened 2026-06-01
feat: support `str | os.PathLike` in `PathProtocol.glob` and `LocalPath.glob`

### LINKED ISSUE #9: Look into supporting `str | os.PathLike` in `PathProtocol.glob` and `LocalPath.glob`
Python 3.13 adds support for `pathlib.Path.glob` to accept a `str | os.PathLike` pattern. Previous versions required the pattern to be a `str` only.

`ContainerPath.glob` supports a `str | os.PathLike` pattern, which is quite useful. `pathlib.Path` is already incompatible with `PathProtocol` due to the extended `write_{bytes,text}` and `mkdir` arguments, so we don't need to worry about breaking compatibility.

We could therefore streamline `PathProtocol.glob` and provide some convenience for `LocalPath.glob` by supporting `str | os.PathLike` as well -- and the implementation would be simple:
```py
class LocalPath(pathlib.PosixPath):
    ...
    def glob(self, pattern: str | os.PathLike[str]) -> Generator[Self]:
        return super().glob(os.fspath(pattern))
```

tonyandrewmeyer:
Python 3.10/3.11/3.12 will hit the new code path; Py 3.13+ will pass through to the now-PathLike-accepting base, which is harmless.

Existing `pathops` functional tests in `tests/functional/test_container_path.py::TestGlob` only ever pass `str` patterns. If you want functional/integration coverage of the `LocalPath` side too, I could add a parametrize over `[str, pathlib.PurePosixPath]` on the `test_ok`. I left it since the unit tests already cover the `LocalPath` shim and the protocol signature.

The `pyright: ignore[reportIncompatibleMethodOverride]` is needed because pyright 1.1.408's stubs for `pathlib.Path.glob` declare `Generator[Path, None, None]` while `PathProtocol.glob` declares `Iterator[Self]`. Newer typeshed actually uses `Self`, so once the pinned `pyright==1.1.408` in `test-requirements.txt` is bumped the ignore may become unused. I couldn't remember whether we want the specific match in the ignore brackets, but I think we decided we did.

The issue mentions `Generator[Self]` as the return; I kept the protocol return as `Iterator[Self]` (unchanged) and the `LocalPath` override as `Iterator[Self]`. In `pathops 1.2.1` both were relaxed to `Iterator` for the same `iterdir`/`glob` pair, so this matches that direction.

Fixes #9


james-garner-canonical (approved):
Thanks for taking care of this. I think this warrants a release, could you bump the minor version and update the changelog in this PR, please?

---

> Python 3.10/3.11/3.12 will hit the new code path; Py 3.13+ will pass through to the now-PathLike-accepting base, which is harmless.

Everything will hit the new codepath since it's unconditional though, right? That is, we pre-convert to a string even on new Pythons. That seems perfectly fine though.

> Existing `pathops` functional tests in `tests/functional/test_container_path.py::TestGlob` only ever pass `str` patterns. If you want functional/integration coverage of the `LocalPath` side too, I could add a parametrize over `[str, pathlib.PurePosixPath]` on the `test_ok`. I left it since the unit tests already cover the `LocalPath` shim and the protocol signature.

Agree that we don't need functional tests here (they're more about exercising the Pebble side, which is untouched by this PR).

> The `pyright: ignore[reportIncompatibleMethodOverride]` is needed because pyright 1.1.408's stubs for `pathlib.Path.glob` declare `Generator[Path, None, None]` while `PathProtocol.glob` declares `Iterator[Self]`. Newer typeshed actually uses `Self`, so once the pinned `pyright==1.1.408` in `test-requirements.txt` is bumped the ignore may become unused.

Interesting, good info. I guess I still don't have dependabot configured how I want since we're a few `pyright` releases behind now.

> I couldn't remember whether we want the specific match in the ignore brackets, but I think we decided we did.

Yeah, we did, thanks.

> The issue mentions `Generator[Self]` as the return; I kept the protocol return as `Iterator[Self]` (unchanged) and the `LocalPath` override as `Iterator[Self]`. In `pathops 1.2.1` both were relaxed to `Iterator` for the same `iterdir`/`glob` pair, so this matches that direction.

Good call.

PR #511, opened 2026-06-02
feat: Cache private key in TLS library

swetha1654:
Addresses: #512


saltiyazan (approved):
Hi @swetha1654 !

Thanks for making the change, looks good to me.
Could you please bump the patch version in `_version.py` before we merge so this is actually published?

swetha1654:
@saltiyazan done

kayra1 (approved):

PR #518, opened 2026-06-08
ci: thin justfile

### LINKED ISSUE #405: Make justfiles more approachable for contributors
The `justfile` does what it's supposed to, but it's grown in complexity over time and can be a bit unapproachable, especially as many editors don't highlight `justfile`s nicely by default.

One approach might be to push the complexity to Python scripts under `.scripts` and have the `justfile` largely be a thin wrapper around those scripts.

### LINKED ISSUE #449: `just <recipe> <package>` shouldn't use an unsupported Python version by default
The `justfile` uses Python 3.10 for recipes by default, as it's the lowest version supported in this repo. But a library can set a higher minimum version, like `>=3.12`. Ideally, `just unit <that package>` wouldn't run with Python 3.10 and fail, but would instead run with the lowest supported major Python version of the package instead.

Perhaps worth handling downstream of #405.

james-garner-canonical:
This PR moves the implementation of the repository's developer tooling from inline `justfile` recipes to a separate `.scripts/just.py` script. This allows us to cover the recipes with `ruff` and `pyright`, and makes many things less tortuous (having a unified approach to `uv run`, for example), and several things that were previously too painful to write are now possible and have been implemented:
- Groups are only added to `uv run` if the library defines them -- this means that CI won't break if a library deletes an unused dependency group.
- The default Python version is not statically defined, but determined based on the library's minimum `requires-python`, which resolves #449
- The separate `interface.just` module is no longer required for `just interface init`, instead we can write `just init --interface`, since recipes can now trivially define boolean flags.
- Accept an explicit `--unsafe-fixes` flag for the `format` recipe, forwarded to `ruff check`.
- We only run `ruff check --diff` to print the diff if `ruff check` fails during `fast-lint`.
- Detailed help for each recipe and argument.

The `justfile` remains in place for compatibility and because it provides the useful feature that `just <recipe>` from anywhere in the repository will execute (and run from the repository root). Each recipe is now just a single line call to `just.py`.

Resolves #405

---

I experimented with having a separate script for each recipe, but I found that this was just a lot less readable due to the extra boilerplate each file requires (shebang, Python version, license, imports, `if __name__ == '__main__'`), and the extra hoops involved in sharing code between recipes (separate `_common.py` module), or having recipes call each other (e.g. `lint` calls `fast-lint` and `static`; `check` calls `lint` and `unit`, and builds the docs).


tonyandrewmeyer:
Do I need to worry about all the failing CI when I review?

tonyandrewmeyer:
> * Accept an explicit `--unsafe-fixes` flag for the `format` recipe, forwarded to `ruff check`.

Not for charmlibs or this PR, but I wonder if we should make `--unsafe-fixes` the default in our other Python repos. We're manually reviewing the changes anyway, so should notice anything that is actually a bad change (and I've yet to see a bad unsafe fix). Saves doing `tox -e format -- --unsafe-fixes` and whatever the Make version is.

.docs/how-to/migrate.md
```diff
@@ -263,7 +263,7 @@ By default each test is treated as compatible with both substrates.
 
 Your library's reference documentation is automatically built from its docstrings and source code. On top of that, you can add tutorials, how-to guides, and explanations under `<library path>/docs`, and they'll be included in this documentation site under the respective categories.
 
-Read more: {ref}`how-to-add-library-docs`
+Read more: {ref}`how-to-add-library-docs`
```
tonyandrewmeyer, line 266: Strange, I cannot find a difference here...

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
+#!/usr/bin/env -S uv run --script --no-project
+
+# /// script
+# requires-python = ">=3.12"
+# dependencies = []
+# ///
+
+# ruff: noqa: I001  # tomllib is first-party in 3.11+
```
tonyandrewmeyer, line 8: Is this needed because ruff can't understand the requires-python at the top?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+
+# Copyright 2025 Canonical Ltd.
```
tonyandrewmeyer:
```suggestion
# Copyright 2026 Canonical Ltd.
```

Claude does this to me too (I'm assuming).

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+
+def _register(fn: _F) -> _F:
+    RECIPES[fn.__name__.replace('_', '-')] = fn
+    return fn
+
+
+def main(argv: list[str]) -> int:
+    """Dispatch `argv` to the named recipe, returning its exit code."""
+    if not argv or argv[0] in ('-h', '--help'):
+        print(QUICK_START)
```
tonyandrewmeyer, line 93: The colouring is nice for humans but not ideal for agents or piping. Could we strip it when not a tty or similar? Also in the exit 2 case below.

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+    func = RECIPES.get(name)
+    if func is None:
+        print(f'Unknown recipe: {name!r}\n', file=sys.stderr)
+        print(QUICK_START)
+        return 2
+    return func(args)
+
+
+@_register
+def help(argv: list[str]) -> int:  # noqa: A001
```
tonyandrewmeyer, line 105: It feels a bit odd to be implementing a help function when argparse (optparse, getopt, click, typer, ...) already exist. I guess it's small, but it makes me wonder what just is giving us here, in exchange. You're already using argparse, can't it do what you need for help?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+def init(argv: list[str]) -> int:
+    """Scaffold a new charmlibs package interactively (forwards extra args to cookiecutter)."""
+    parser = _parser(init)
+    parser.add_argument(
+        '--interface',
+        action='store_true',
+        help='Scaffold a charmlibs.interfaces package instead of a general charmlibs package.',
+    )
+    args, cookiecutter_args = parser.parse_known_args(argv)
+    if args.interface:
```
tonyandrewmeyer, line 129: Same comment as above about agents/piping and ANSI codes.

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+    parser = _parser(add)
+    parser.add_argument('package', help='Path from the repo root to the package, e.g. `pathops`.')
+    args, uv_args = parser.parse_known_args(argv)
+    cmd = ['uv', 'add', '--constraints', TEST_REQUIREMENTS, *uv_args]
+    _run(cmd, cwd=REPO_ROOT / args.package)
+    return 0
+
+
+@_register
+def check(argv: list[str]) -> int:
```
tonyandrewmeyer, line 167: Nit: wouldn't this be tidier in the linting recipes section below?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+
+
+@_register
+def functional(argv: list[str]) -> int:
+    """Run functional tests with `coverage` for a package."""
+    args, pytest_args = _package_parser(functional).parse_known_args(argv)
+    python = _resolve_python(args.package, args.python)
+    cmds = _coverage_cmds(args.package, 'functional', python, pytest_args or ['-rA'])
+    joined = ' && '.join(shlex.join(str(part) for part in cmd) for cmd in cmds)
+    script = textwrap.dedent(
```
tonyandrewmeyer, line 278: This sort of thing would normally use a heredoc, no? Would that be useful here?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+- {CYAN}just functional <package>{NORMAL} (may require extra software like {CYAN}pebble{NORMAL})
+
+{BOLD}Run integration tests{NORMAL} (requires a Juju controller and a cloud):
+- Pack: {CYAN}just pack-k8s <package>{NORMAL} or {CYAN}just pack-machine <package>{NORMAL}
+- Run: {CYAN}just integration-k8s <package>{NORMAL} or {CYAN}just integration-machine{NORMAL}
+
+{BOLD}Build the docs: {CYAN}just docs{NORMAL}
+- For specific packages only: {CYAN}just docs html <packages>{NORMAL}
+""".strip()
+# Mapping of recipe names to their functions, populated by the `_register` decorator.
```
tonyandrewmeyer, line 80: Maybe worth mentioning that order matters here and what future additions should do in that regard?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+    """Pack machine charm(s) for a package's Juju integration tests."""
+    return _pack(pack_machine, 'machine', argv)
+
+
+def _pack(fn: FunctionType, substrate: str, argv: list[str]) -> int:
+    """Pack the package's integration-test charm(s) for the given substrate."""
+    parser = _parser(fn)
+    parser.add_argument(
+        '--tag',
+        default=os.environ.get('CHARMLIBS_TAG', ''),
```
tonyandrewmeyer, line 365: I can't remember the exact system here: is it correct that if there's no tag we pass an empty string in rather than leaving it out?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+        return 2
+    return func(args)
+
+
+@_register
+def help(argv: list[str]) -> int:  # noqa: A001
+    """Describe usage and list the available recipes."""
+    _parser(help).parse_args(argv)
+    print('All recipes require `uv` to be available.\n')
+    print('Available recipes:')
```
tonyandrewmeyer, line 109: This doesn't include the docs ones, which live elsewhere. Maybe worth including?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+- Run {CYAN}ruff{NORMAL} for all packages: {CYAN}just fast-lint{NORMAL}
+
+{BOLD}Run individual checks for a package:{NORMAL}
+- {CYAN}just lint <package>{NORMAL} (fix errors with {CYAN}just format <package>{NORMAL})
+- {CYAN}just unit <package>{NORMAL}
+- {CYAN}just functional <package>{NORMAL} (may require extra software like {CYAN}pebble{NORMAL})
+
+{BOLD}Run integration tests{NORMAL} (requires a Juju controller and a cloud):
+- Pack: {CYAN}just pack-k8s <package>{NORMAL} or {CYAN}just pack-machine <package>{NORMAL}
+- Run: {CYAN}just integration-k8s <package>{NORMAL} or {CYAN}just integration-machine{NORMAL}
```
tonyandrewmeyer, line 74:
```suggestion
- Run: {CYAN}just integration-k8s <package>{NORMAL} or {CYAN}just integration-machine <package>{NORMAL}
```

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+def combine_coverage(argv: list[str]) -> int:
+    """Combine a package's `coverage` reports."""
+    args = _package_parser(combine_coverage).parse_args(argv)
+    pkg_dir = REPO_ROOT / args.package
+    python = _resolve_python(args.package, args.python)
+    env = _coverage_env()
+    data_files = [
+        f
+        for test_id in ('unit', 'functional')
+        if (pkg_dir / (f := f'.report/coverage-{test_id}-{python}.db')).exists()
```
tonyandrewmeyer: I feel like a regular generator would be clearer than this walrus. But your call.

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+
+def _resolve_python(package: str, python: str | None) -> str:
+    """Return the Python version to test `package` with.
+
+    If `python` is `None`, return the higher of 3.10 and the package's minimum Python version.
+    """
+    if python:
+        return python
+    minimum = _requires_python_minimum(REPO_ROOT / package)
+    return max('3.10', minimum, key=lambda s: tuple(int(p) for p in s.split('.')))
```
tonyandrewmeyer: Maybe make 3.10 a module level constant so it's easier to find when we move to 3.12?

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+    """Return the `major.minor` lower bound of `pkg_dir`'s `requires-python`, e.g. `'3.10'`."""
+    pyproject_toml = tomllib.loads((pkg_dir / 'pyproject.toml').read_text())
+    requires_python = pyproject_toml['project']['requires-python']
+    regex = (
+        r'(?:>=|~=)'  # a `>=` or `~=` operator: the ones that set a lower bound
+        r'\s*'  # optional whitespace between the operator and the version
+        r'(\d+\.\d+)'  # capture just `major.minor`, stopping before any patch component
+    )
+    match = re.search(regex, requires_python)
+    assert match is not None
```
tonyandrewmeyer: Probably should be a ValueError raise, although why anyone would run this with asserts out I don't know.

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+        return python
+    minimum = _requires_python_minimum(REPO_ROOT / package)
+    return max('3.10', minimum, key=lambda s: tuple(int(p) for p in s.split('.')))
+
+
+def _requires_python_minimum(pkg_dir: pathlib.Path) -> str:
+    """Return the `major.minor` lower bound of `pkg_dir`'s `requires-python`, e.g. `'3.10'`."""
+    pyproject_toml = tomllib.loads((pkg_dir / 'pyproject.toml').read_text())
+    requires_python = pyproject_toml['project']['requires-python']
+    regex = (
```
tonyandrewmeyer, line 507: Misses requires-python = "==3.10.*", "<4", or any other without >=/~=. Maybe add an explicit raise with a "couldn't parse requires-python = ..." message.

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+# --- Linting recipes ---------------------------------------------------------------------------
+
+
+@_register
+def lint(argv: list[str]) -> int:
+    """Run fast linting (`ruff`) and static analysis (`pyright`) for a package."""
+    args, pyright_args = _package_parser(lint).parse_known_args(argv)
+    python = _resolve_python(args.package, args.python)
+    failures = _fast_lint(args.package)
+    if _static(args.package, python, pyright_args) != 0:
```
tonyandrewmeyer, line 210: This loses the info from pyright's exit code. Probably ok, I never use it myself, but just mentioning in case it was accidental.

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+        sys.exit(failures)
+    for cmd in _coverage_cmds(args.package, 'unit', python, ['-rA']):
+        _run(cmd, cwd=REPO_ROOT / args.package, env=_coverage_env())
+    _run(['just', 'docs', 'html', args.package])
+    return 0
+
+
+@_register
+def format(argv: list[str]) -> int:  # noqa: A001
+    """Run `ruff check --fix` and `ruff format`, modifying files in place."""
```
tonyandrewmeyer, line 181: Usually better to check then format, so format formats anything check does.

tonyandrewmeyer (approved):
Overall, seems like it's nicer UX. Technically breaking changes, but I assume we don't care about that here, which seems reasonable to me (for now).

How much `just` is left here? Enough that there is value over Make?

I tested locally:

  - just (quick-start banner), just help
  - just init --help, just lint --help, just integration-k8s --help, and so on
  - just init --no-input project_slug=foo and just init --interface --no-input project_slug=baz
  - just fast-lint pathops
  - just static pathops
  - just lint --python 3.12 pathops and the default (3.10)
  - just lint --python 3.10 rollingops
  - just unit pathops -k ...
  - just scripts-unit
  - just interfaces-json
  - just meh
  - ruff + pyright pass on .scripts/just.py itself

The PR description mentions pyright coverage, but CI doesn't actually pyright-check .scripts/just.py. fast-lint runs ruff over the repo (which includes it), and scripts-unit runs the test suite, but from what I can see no job runs pyright on .scripts/just.py.

There's no tests for just.py itself, which is ~500 lines, has logic like Python resolving, minimum Python version calculation, uv group filtering. Maybe worth doing for selected pieces? Or some light smoke test?

Nit: init doesn't propagate cookiecutter's failure when cookiecutter exits 0 on "directory already exists". This is cookiecutter's quirk, not the PR's bug, and matches prior justfile behavior. Not a blocker.

Ignoring the CI failures and assuming you will sort that out.

james-garner-canonical:
> Do I need to worry about all the failing CI when I review?

No, this is just the beauty of Concierge + Juju + Charmcraft at (modest) scale :)

james-garner-canonical:
> Overall, seems like it's nicer UX. Technically breaking changes, but I assume we don't care about that here, which seems reasonable to me (for now).

The breaking changes are:
- Removal of `interfaces.just` so you have to run `just init --interface` instead of `just interface init` is breaking. I figure that one is infrequent enough to run that it's OK, though it would be low effort to keep `interfaces.just` around for compatibility -- I just didn't figure it was worth it.
- Using `--python` and `--tag` options instead of `just` variables (`just python=foo <recipe>` vs `just <recipe> --python=foo`). I think these deserved to be socialised around the release, but should be OK. I don't think anyone would be using `tag`, but `python` may be in use.

> How much `just` is left here? Enough that there is value over Make?

The killer features over `make` are:
1. Cleanly pass arguments to recipes without needing to use a variable and quote.
2. Run from anywhere in the repo instead of only at the root.

I think 1 alone justifies it for me, and 2 wouldn't cut it alone. I'd happily switch to an alternative task runner that provides 1, though I'm not convinced it would be worth the breakage and churn.

> The PR description mentions pyright coverage, but CI doesn't actually pyright-check .scripts/just.py. fast-lint runs ruff over the repo (which includes it), and scripts-unit runs the test suite, but from what I can see no job runs pyright on .scripts/just.py.

You're right, opening #521 to track this. For now I get type checking in my editor, which is nice.

> There's no tests for just.py itself, which is ~500 lines, has logic like Python resolving, minimum Python version calculation, uv group filtering. Maybe worth doing for selected pieces? Or some light smoke test?

Yeah I wasn't sure about this. We get a certain amount of end-to-end use in CI that catches outright breakage but not subtle errors. I'll add some tests to this PR.

> Nit: init doesn't propagate cookiecutter's failure when cookiecutter exits 0 on "directory already exists". This is cookiecutter's quirk, not the PR's bug, and matches prior justfile behavior. Not a blocker.

I'm probably misunderstanding this one, here's what I see:
```
> just init
✨IMPORTANT✨ The project name should be the import package name, without the charmlibs. namespace.
You can press enter to accept the default, shown in brackets.
['uvx', 'cookiecutter', '.template']
  [1/3] 🌈 The project name (example): pathops
  [2/3] 🐍 The minimum Python version supported by the package (3.10): 
  [3/3] 📝 The author field for PyPI (The Charm Tech team at Canonical): 
Error: "pathops" directory already exists
error: Recipe `init` failed on line 14 with exit code 1
```
(Should probably suppress that `['uvx', 'cookiecutter', '.template']` in this case :)

.docs/how-to/migrate.md
james-garner-canonical: I guess it's a line ending change

.scripts/just.py
james-garner-canonical: Yeah ruff doesn't seem to pay attention to the script metadata.

.scripts/just.py
james-garner-canonical: Editorial choice to keep the line short

.scripts/just.py
james-garner-canonical:
```suggestion
# Mapping of recipe names to their functions, populated by the `_register` decorator.
# `just help` prints the recipes in registration order.
```

.scripts/just.py
james-garner-canonical: Ah I should have said, I hope to migrate the `docs.just` shenanigans in a similar fashion to this after landing this PR and the Sphinx pack migration.

.scripts/just.py
james-garner-canonical: To do this via `argparse` we'd need to switch to a subparser based approach, which would mean separating the parsers from the recipe function definitions and still implementing a mapping of some sort (e.g. an ad-hoc attribute using `setdefault` on each subparser). On balance I prefer the `RECIPES` + `@_register` + local parsers approach with lightweight help recipe. Maybe this indicates that `click` would be a better fit here than `argparse`, but I'd prefer to avoid the dependency.

.scripts/just.py
james-garner-canonical: It's lint + unit + docs as a convenience entrypoint for a package, so I'd prefer to keep it near the top.

.scripts/just.py
james-garner-canonical: Yeah, we replace the exit codes of `ruff check` and `ruff format --diff` with the failure count in `_fast_lint`, and continue that here. When running multiple commands and failing afterwards if any of them failed, it's not clear to me that we should propagate the error code of any one in particular.

Maybe it's worth logging the exit codes on failure (when we don't immediately exit on failure), WDYT?

.scripts/just.py
james-garner-canonical: I'm not sure how that would be applied here tbh. But using a heredoc would require us to use `subprocess.run` with `shell=True` which doesn't seem desirable.

.scripts/just.py
james-garner-canonical: I'm not sure what you mean by a regular generator, but I agree that this is ugly. Maybe this is simplest:
```py
    data_files: list[str] = []
    for test_id in 'unit', 'functional':
        file = f'.report/coverage-{test_id}-{python}.db'
        if (pkg_dir / file).exists():
            data_files.append(file)
```

.scripts/just.py
```diff
@@ -0,0 +1,553 @@
...
+    """Combine a package's `coverage` reports."""
+    args = _package_parser(combine_coverage).parse_args(argv)
+    pkg_dir = REPO_ROOT / args.package
+    python = _resolve_python(args.package, args.python)
+    env = _coverage_env()
+    data_files = [
+        f
+        for test_id in ('unit', 'functional')
+        if (pkg_dir / (f := f'.report/coverage-{test_id}-{python}.db')).exists()
+    ]
```
james-garner-canonical:
```suggestion
    data_files: list[str] = []
    for test_id in 'unit', 'functional':
        file = f'.report/coverage-{test_id}-{python}.db'
        if (pkg_dir / file).exists():
            data_files.append(file)
```

.scripts/just.py
james-garner-canonical: Yeah, the `justfile` did the same thing: `tag := env('CHARMLIBS_TAG', '')`

.scripts/just.py
james-garner-canonical: Makes sense, but we'd probably still find all the 3.10's by grepping anyway, and given the size of this script I'd prefer to keep this inlined at the single site of use.

.scripts/just.py
james-garner-canonical: Good idea, I'll make the error more informative. I'm fine with requiring `>=` or `~=` for now as it matches current practice and seems like the best practice for libraries.

.scripts/just.py
james-garner-canonical:
```suggestion
    if match is None:
        raise ValueError('Couldn't find minimum version in requires-python: {requires_python!r}')
```

.scripts/just.py
tonyandrewmeyer: Oh yeah, I was thinking of running through a shell, but not doing that makes this unnecessary.

.scripts/just.py
tonyandrewmeyer: Yeah, logging seems a reasonable compromise.

.scripts/just.py
james-garner-canonical: That makes sense. The current approach is because `ruff check --fix` can exit with failure, which makes `just format` exit before even running `ruff format`, leaving you with nasty unformatted code that you're supposed to fix the `check` errors in (or drop down to running `ruff format` manually like an animal). I definitely see the value in running `format` after `check --fix` though, for the reason you say: so fixes are formatted correctly.

---

I was thinking about running `check --fix`, then `format`, and then afterwards exiting with failure if `check --fix` failed, but that makes the `check --fix` output less helpful -- line numbers may have changed and linting violations may even have been eliminated by correct formatting.

So I suppose the desired behaviour is `ruff format`, *then* `ruff check --fix`, and then a final `ruff format` only if there were changes but no errors from `check`. The problem there is that normally `ruff check` exits with 0 whether there were fixes applied or not, and with 1 only if there were unfixable errors. You can pass `--exit-non-zero-on-fix`, but unfortunately that changes the behavior to exit 1 either if there were fixes applied or if there were unfixable errors, preventing us from knowing if we should exit early or not.

I don't want to introduce the complexity of capturing and parsing the output of tools to distinguish the cases, and I don't want to format after unfixable errors were detected ... so maybe the correct approach is `format`, then `check`, then exit if there were unfixable errors, otherwise format a final time ... or maybe we need to format anyway since there might have been successful changes that need formatting, but then the line numbers are potentially changed so we'd want to re-run check a final time ... at which point I'm clearly overthinking things ...

james-garner-canonical: Hmm no wait, if we're doing two checks, then the first format in format, check, format, check must be redundant ...

.scripts/just.py
james-garner-canonical: Good call, I've created #523 to track this (this PR currently matches the existing `justfile` behaviour).

.scripts/just.py
james-garner-canonical: Thanks, see #523.

.scripts/just.py
james-garner-canonical: On further consideration I'm going to keep this as-is for now, as a human needs to look at the output either way and the existing messages from Pyright make it clear what the failure reason was.

james-garner-canonical:
> (Should probably suppress that `['uvx', 'cookiecutter', '.template']` in this case :)

I thought about adding a `suffix` argument so that the custom header I print can show up after the command, but that would make it come out on stderr instead of stdout, so I'll leave this as-is too.

.scripts/just.py
james-garner-canonical: I decided to add a constant for testing.

PR #522, opened 2026-06-09
feat(istio-metadata): adds the istio-metadata library

adhityaravi:
## Add istio-metadata interface library

Migrates the `istio-metadata` interface libraries from [`istio-k8s-operator`](https://github.com/canonical/istio-k8s-operator/) into charmlibs.

### New packages

- **`istio-metadata`** (`charmlibs.interfaces.istio_metadata`): Interface for sharing istio control plane metadata (root namespace for eg) between the istio charm and consuming charms.

### What is this library for?

This library provides the provider and requirer sides of the ``istio-metadata`` relation interface, used to transfer information about an instance of Istio (such as its root namespace) to charms that need to interface with Istio.

It might be necessary for complex charms the spin up dynamic namespaces and mesh related resources autonomously from the istio charms. For eg. the charmed kubeflow stack,

Currently its used by the [kiali-k8s-operator](https://github.com/canonical/kiali-k8s-operator/)

### Context

This library is already in use so this would be a pure migration.


interfaces/istio-metadata/src/charmlibs/interfaces/istio_metadata/__init__.py
```diff
@@ -0,0 +1,92 @@
...
+``IstioMetadataRequirer`` is designed for relating to a single application)::
+
+    requires:
+      istio-metadata:
+        limit: 1
+        interface: istio_metadata
+"""
+
+from ._istio_metadata import (
+    DEFAULT_RELATION_NAME,
```
james-garner-canonical: Note that exposing this constant here has different semantics than exposing it in the Charmhub-style lib. In the old, single-file module, the value would be read-write. Here it's essentially read only, as a user setting `charmlib.interfaces.istio_metadata.DEFAULT_RELATION_NAME` to a different value has no effect on the `_istio_metadata.DEFAULT_RELATION_NAME` variable used by the library objects.

- If users don't really need this value, I'd drop it from the public API.
- If it's needed in read-only form, I guess it's OK as-is, though potentially confusing -- if the use case is knowing what relation_name a library object is using, maybe a read-only property would be cleaner? That would introduce code changes, which requires a judgement call on your part as to the migration burden.
- If it's needed in read-write form, then maybe the best option would be to expose a read-write property on the library objects?

james-garner-canonical (commented):
Looks good, thanks. One judgement question for you about exporting the default relation name at the package level (currently a read-only value in practice -- see comment).

The metadata looks good. We can merge with @dstathis's approval of the lift-and-shift of the code and tests.

One request: could you please add interface metadata [like this](https://github.com/canonical/charmlibs/tree/main/interfaces/k8s_backup_target/interface/v0)? It's pretty easy to generate with an LLM now, but the real value comes from your domain expert sign off on its correctness. The metadata provided there becomes the source of truth for Charmhub. Note the pattern for the schema -- it's just a symlink to a `_schema.py` module in your library, exposing what your library encodes about the databag format at a standard location. The only restriction is that the only third party dependency of the `_schema.py` module is Pydantic.

interfaces/istio-metadata/src/charmlibs/interfaces/istio_metadata/__init__.py
adhityaravi: sure, I can drop this. just a copy paste artifact, not as important.

adhityaravi:
@james-garner-canonical the metadata format that you mention is not clear to me. Could you elaborate? Should i add the same interfaces/v1 with the `_schema.py`, include a README and then create a new `_schema.py`, to which the metadata will point to, in the actual src folder? At the moment there is no `_schema.py` in this PR. (or in any of our service mesh libraries in here)

james-garner-canonical:
> @james-garner-canonical the metadata format that you mention is not clear to me. Could you elaborate? Should i add the same interfaces/v1 with the `_schema.py`, include a README and then create a new `_schema.py`, to which the metadata will point to, in the actual src folder? At the moment there is no `_schema.py` in this PR. (or in any of our service mesh libraries in here)

In this case it would be interfaces/<interface name>/interface/v0. The v0 directory should contain a readme and interface.yaml file.

A schema.py file is also recommended. This could just be a duplicate of the schema used in the library itself. But if you want to avoid duplication, you can instead have it be a symlink to a library module with the expected contents, which is what the linked example does (though this would require splitting the schema used by the library into a separate module). It's currently up to you which approach you prefer. Duplication is fine.

interfaces/istio-metadata/src/charmlibs/interfaces/istio_metadata/__init__.py
adhityaravi: done

adhityaravi:
@james-garner-canonical I think I have addressed your comments. Could you give it another quick glance before I annoy Dylan to approve it for me? thanks!

dstathis:
This is a direct migration. I approve.

dstathis (approved):

james-garner-canonical:
> @james-garner-canonical I think I have addressed your comments. Could you give it another quick glance before I annoy Dylan to approve it for me? thanks!

Sorry I ran out of time today, just had a quick look and it looks good, I'll approve, merge, and release tomorrow.

adhityaravi:
@james-garner-canonical some love for this PR would be nice 🙂

james-garner-canonical (approved):

.docs/reference/libs.yaml
```diff
@@ -1830,6 +1830,30 @@ interfaces:
...
   tags:
   - networking
+- name: charmlibs.interfaces.istio_metadata
+  status: recommended
+  url: https://pypi.org/project/charmlibs-interfaces-istio-metadata
+  docs: https://documentation.ubuntu.com/charmlibs/reference/charmlibs/interfaces/istio-metadata
+  src: https://github.com/canonical/charmlibs/tree/main/interfaces/istio-metadata
+  kind: PyPI
+  rel_name: istio-metadata
+  rel_url_charmhub: ''
```
james-garner-canonical: Must match canonical name used in `charmcraft.yaml`.

interfaces/istio_metadata/interface/v0/interface.yaml
```diff
@@ -0,0 +1,18 @@
+name: istio_metadata
```
james-garner-canonical, line 1: Should match canonical name used in `charmcraft.yaml`.

interfaces/istio_metadata/src/charmlibs/interfaces/istio_metadata/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+            )
+
+        def _publish(self, _):
+            self.istio_metadata.publish(root_namespace='istio-system')
+
+The provider's ``charmcraft.yaml`` should declare::
+
+    provides:
+      istio-metadata:
+        interface: istio_metadata
```
james-garner-canonical, line 46: Is this the canonical interface name?

interfaces/istio-metadata/src/charmlibs/interfaces/istio_metadata/_schema.py
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Schema models for the istio-metadata interface.
+
+This module is the source of truth for the databag wire format of the
+``istio-metadata`` interface, and is symlinked from
```
james-garner-canonical: Should match canonical name used in `charmcraft.yaml`.

CODEOWNERS
```diff
@@ -92,6 +92,9 @@
 # /interfaces/istio-ingress-route/
 /interfaces/istio-ingress-route @canonical/service-mesh
 
+# /interfaces/istio-metadata/
+/interfaces/istio-metadata @canonical/service-mesh
```
james-garner-canonical: The directory name the library is stored at should match the canonical interface name, as used in `charmcraft.yaml`.

james-garner-canonical (approved):
There's one discrepancy that needs to be resolved before merging: is the canonical interface name `istio-metadata` or `istio_metadata`? There are a few places where we need to use the canonical interface name (as used in `charmcraft.yaml`): `libs.yaml`, the directory name `interfaces/<canonical name>`, `interface.yaml`, docs like the interface readme.

Note that in other locations, we normalize to hyphens (PyPI package name, docs URLs), and we use underscores where Python syntax requires (`charmlibs.interfaces.istio_metadata`).

Publishing is set up so the library should publish on merge.

.docs/reference/libs.yaml
adhityaravi: Ah I see, I wasnt aware of this requirement. Resolved

interfaces/istio_metadata/interface/v0/interface.yaml
adhityaravi: matches

interfaces/istio_metadata/src/charmlibs/interfaces/istio_metadata/__init__.py
adhityaravi: yup

CODEOWNERS
adhityaravi: done

PR #524, opened 2026-06-14
fix(pathops): support path-like pattern in match, reject ContainerPath

### LINKED ISSUE #369: Investigate supporting newer pathlib.Path.match features
Python 3.12 added support for `pathlib.Path.match` to accept a path-like object instead of just a string. Python 3.14 subtly adjusted the behaviour to also work for any object that implements `with_segments`, even if it isn't path-like.

Python 3.13 added the related method `pathlib.Path.full_match`.

We should investigate which (if any) of these changes `charmlibs.pathops` could (and should) support now, and clearly document when we should add support for other features.

The `match` behaviour change on Python 3.14 prompted the creation of this issue, as a `ContainerPath` will now be accepted at runtime by `pathlib.PurePath.match`, `LocalPath.match`, and `ContainerPath.match`. At the very least, `ContainerPath.match` should probably return `False` if the containers don't match, if not make it an error to use `ContainerPath` here since it isn't `os.PathLike`.

tonyandrewmeyer:
This PR widens `match()` on `PathProtocol`, `LocalPath`, and `ContainerPath` to accept `str | os.PathLike[str]`, matching `pathlib.Path.match` on Python 3.12+ and the signature already used by `glob()` (see #505).

`ContainerPath.match` additionally rejects `ContainerPath` arguments with a `TypeError`. This restores the pre-3.14 behaviour: Python 3.14's `pathlib.PurePath.match` accepts any object that implements `with_segments`, which would otherwise silently let a `ContainerPath` through and match the wrong filesystem.

`LocalPath` gains a thin `match` override that calls `os.fspath` before delegating, so the path-like pattern works on 3.10/3.11 too (where `pathlib.Path.match` is `str`-only). The override also forwards the `case_sensitive` keyword on 3.12+ to stay signature-compatible with the parent across all supported Python versions.

Resolves #369


pathops/src/charmlibs/pathops/_local_path.py
```diff
@@ -171,14 +171,19 @@ def glob(  # pyright: ignore[reportIncompatibleMethodOverride]
...
         # path-like). On Python 3.14, pathlib.PurePath.match accepts any object with
         # `with_segments`, which would silently allow a ContainerPath through. Normalising
         # via os.fspath gives us a consistent, narrow API: str | os.PathLike[str] on every
         # supported version, with a clear TypeError for anything else (including ContainerPath,
         # which is not os.PathLike).
-        return super().match(os.fspath(path_pattern))
+        kwargs: dict[str, bool] = {}
+        if case_sensitive is not None:
+            kwargs['case_sensitive'] = case_sensitive
+        return super().match(os.fspath(path_pattern), **kwargs)
```
james-garner-canonical: I'm not sure about this change. It does resolve the incompatible override error across all Python versions. But the error is because `match` added `case_sensitive` in Python 3.12. It's valid in typing terms for the override to add an extra argument (as we're now doing on Python 3.10), but we've implemented our handling of the extra argument in such a way that we hide from both callers of the function and from our own static analysis that it's an error to pass `case_sensitive` on Python <3.12.

tonyandrewmeyer: The problem is that we need to satisfy the checker that wants an ignore in one version and forbids having an ignore in another version. Any suggestions?

tonyandrewmeyer: I found something that seems to work.

pathops/src/charmlibs/pathops/_local_path.py
```diff
@@ -171,6 +172,24 @@ def glob(  # pyright: ignore[reportIncompatibleMethodOverride]
...
         return super().glob(os.fspath(pattern))
 
+    # On Python 3.11, pathlib.Path.match only accepts a str pattern and has no
+    # case_sensitive keyword (3.12 added both). On Python 3.14, pathlib.PurePath.match
+    # accepts any object with `with_segments`, which would silently allow a ContainerPath
+    # through. Normalising via os.fspath gives us a consistent, narrow API on every
+    # supported version, with a clear TypeError for anything else (including ContainerPath,
+    # which is not os.PathLike). The signature is version-gated so callers on <3.12 get a
+    # static error if they pass case_sensitive rather than a hidden runtime TypeError.
+    if sys.version_info >= (3, 12):
```
james-garner-canonical, line 182: Maybe conditional definition is the cleanest general approach. It's a bit tricky for the docs, since we'll only render the docs for the version we happen to build the docs with, but it should be clean in IDEs.

In this case, it seems pretty straightforward to support `case_sensitive` for older Python versions, since we're explicitly a sub-class of `PosixPath`. For example, something like this:
```py
if case_sensitive or case_sensitive is None:
    return super().match(os.fspath(path_pattern))  # PosixPath.match is case-sensitive
return pathlib.PosixPath(os.fspath(self).lower()).match(os.fspath(path_pattern).lower())
```

What's your assessment of the pros and cons of conditional definition vs backporting `case_sensitive` like this?

james-garner-canonical: @tonyandrewmeyer WDYT here? I actually think the conditional definition approach might be the cleaner path, and it establishes a pattern we could use elsewhere in this library -- plus it makes tearing out the code for older versions super easy when dropping support for them.

The `os.fspath` conversion I'm less sure about. I think it's fine for `LocalPath` to reject `ContainerPath` at the type-checking level (as `pathlib.Path` does too currently), but silently work with it the way `pathlib.Path` would.

james-garner-canonical: I guess one nasty thing with conditional definition is that you can end up with a bit of a proliferation of docstrings and such. More relevant if we used this approach for `ContainerPath` too, which we should consider.

james-garner-canonical:
Let's land only the `ContainerPath` signature expansion, and leave the protocol (and local path) expansion out, but document why (can't add an override to LocalPath without handling the case_sensitive argument that isn't present on older versions we support) and when the protocol can expand (when we're Python 3.X+).

pathops/src/charmlibs/pathops/_container_path.py
```diff
@@ -151,14 +151,30 @@ def is_absolute(self) -> bool:
...
-        return self._path.match(path_pattern)
+        # Python 3.14's pathlib.PurePath.match accepts any object with `with_segments`, which
+        # would silently allow a ContainerPath through. Reject it explicitly to match the
+        # behaviour of joinpath / __truediv__ and stay consistent across Python versions.
+        if isinstance(path_pattern, ContainerPath):
+            raise TypeError(
+                f'ContainerPath is not a valid pattern for ContainerPath.match: {path_pattern!r}'
+            )
+        # os.fspath normalises path-like to str; this also produces a clear TypeError for
+        # anything that is neither str nor os.PathLike, matching pathlib.Path.match on 3.12+.
```
james-garner-canonical:
```suggestion
        # Python <3.12 requires path_pattern to be a string. Python 3.12+ accepts any path-like.
        # When the library requires Python 3.12+, we can drop the os.fspath call and this comment.
```

pathops/src/charmlibs/pathops/_local_path.py
```diff
@@ -171,6 +171,12 @@ def glob(  # pyright: ignore[reportIncompatibleMethodOverride]
         # ContainerPath.glob accepts str | os.PathLike[str], so we normalise here to match.
         return super().glob(os.fspath(pattern))
 
+    # NOTE: there is deliberately no match override to go with ContainerPath.match's
+    # str | os.PathLike[str] pattern. pathlib.Path.match's signature varies across the
+    # versions we support -- the case_sensitive keyword only exists from 3.12 -- so an
+    # override would have to be version-gated to avoid either dropping the keyword or
+    # passing one that doesn't exist. Add it once the minimum supported Python is 3.12.
+
```
james-garner-canonical: I may be wrong, but once we're Python 3.12+ won't `LocalPath` just be fine as-is, requiring no further changes?

```suggestion
```

If there's something concrete that we'd be adding here for Python 3.12+, let's comment that instead.

pathops/src/charmlibs/pathops/_types.py
```diff
@@ -116,6 +116,12 @@ def is_absolute(self) -> bool:
         ...
 
     # NOTE: Not supported -- (Python 3.12) ``case_sensitive`` keyword argument
+    # NOTE: Not supported -- (Python 3.12) path-like ``path_pattern``
+    # ContainerPath.match accepts str | os.PathLike[str], but the protocol stays narrow.
+    # Widening it here would oblige us to shim LocalPath.match too, and we can't write that
+    # override without also handling the case_sensitive keyword, which pathlib.Path.match
+    # doesn't have before 3.12. Widen both -- and drop both NOTEs -- once the minimum
+    # supported Python is 3.12.
```
james-garner-canonical: I think drop the last sentence.
```suggestion
    # doesn't have before 3.12.
```

james-garner-canonical (approved):
Thanks for doing this one, and navigating the fraught waters of what we can and can't support. A few comment tidy ups to address before merging, but otherwise good to go!

PR #525, opened 2026-06-16
ci: also export library docs URLs for interfaces

james-garner-canonical:
This PR updates `.scripts/ls.py` and `.scripts/just.py` to export both URLs to the interface library docs and the low-level interface docs (currently we only export the URLs for the interface docs). The goal is to have the library docs URLs surfaced on Charmhub on the interface pages.


.scripts/ls.py
```diff
@@ -411,11 +436,14 @@ def _get_schema_path_str(category: str, root: pathlib.Path, path: pathlib.Path)
 def _get_dist_name(package: pathlib.Path, root: pathlib.Path = _REPO_ROOT) -> str:
     """Load distribution package name from pyproject.toml and normalize it."""
     name = _pyproject_toml(package, root=root)['project']['name']
-    return _normalize(name.strip())
+    # Normalize distribution package name according to PyPI rules.
+    # https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization
+    return re.sub(r'[-_.]+', '-', name).lower().strip()
```
james-garner-canonical, line 435: Drive-by: this is just in-lining a function that was used in only one place.

.scripts/ls.py
```diff
@@ -441,9 +469,43 @@ def _interface_yaml(path: pathlib.Path, root: pathlib.Path = _REPO_ROOT):
...
+@dataclasses.dataclass
+class _LibEntry:
+    name: str
+    status: str
+    url: str
+    docs: str
+    src: str
+    kind: str
+    description: str
+    tags: list[str]
```
james-garner-canonical, line 480: Slight refactor to stop entirely relying on `yaml.safe_load` being typed as `Any`.

.scripts/ls.py
```diff
@@ -452,29 +514,11 @@ def _lib_urls() -> dict[str, str]:
...
-def _libs_yaml():
-    return yaml.safe_load((_REPO_ROOT / '.docs' / 'reference' / 'libs.yaml').read_text())
-
-
-def _normalize(name: str) -> str:
-    """Normalize distribution package name according to PyPI rules.
-
-    https://packaging.python.org/en/latest/specifications/name-normalization/#name-normalization
-    """
-    return re.sub(r'[-_.]+', '-', name).lower()
```
james-garner-canonical, line 476: Drive-by: in-lined above.

.scripts/ls.py
```diff
@@ -361,14 +364,29 @@ def _get_lib_name(category: str, root: pathlib.Path, path: pathlib.Path) -> str:
...
+    """Return URL to the library docs for this package, or interface docs for this interface."""
     if category == 'packages':
         url = _pyproject_toml(path, root=root)['project']['urls'].get('Documentation', '')
         return url.strip()
     assert category == 'interfaces'
     return f'https://documentation.ubuntu.com/charmlibs/reference/interfaces/{path.name}/'
 
 
+def _get_lib_docs_url(category: str, root: pathlib.Path, path: pathlib.Path) -> str:
+    """Return URL to the docs for this library (or the library for this interface)."""
```
dwilding, line 376:
```suggestion
    """Return URL to the docs for this library (or the library docs for this interface)."""
```

dwilding (approved):

.scripts/ls.py
james-garner-canonical: Dang English and its unclear lexical scopes. The parenthetical is intended to apply to '*this* library (or some other library)'.

PR #532, opened 2026-06-16
feat: Implement TLS Provider Capability sharing

saltiyazan:
Implement sharing provider capabilities in the tls relation data.
This PR introduces a spec/plan using speckit and the implementation.


VercinGor (approved):
Spec looks in order, don't see anything out of place with implementation either, good job, thanks!

kayra1 (approved):
lgtm

PR #540, opened 2026-06-21
docs: create underscored versions of every page as a redirect target

### LINKED ISSUE #541: Automate docs redirects between underscored and hyphenated page names
Library names can be validly spelled with hyphens and underscores in different contexts -- the canonical distribution package name is hyphenated (per PyPI behaviour), while underscores are required when importing a package. In the docs, we made the decision to always use hyphenated URLs, matching PyPI, but page titles often use the import name, and it's a common 'mistake' to write 'reference/charmlibs/foo_bar' in  URL. Ideally these would just redirect to the canonical URL.

Interface name are different -- the hyphen vs underscore distinction is meaningful to Juju.The canonical library path in the monorepo and URL in the documentation uses the actual interface name, preserving hyphens and underscores as-is. However, we wouldn't permit adding an interface whose name only differed in punctuation from an existing interface. So a query for (e.g.) the non-existent `tls_certificates` interface should really redirect to `tls-certificates`.

james-garner-canonical:
This PR adds a custom Sphinx extension to automatically create redirects from URLs using underscores to the canonical page using hyphens.
- Library docs use hyphens for the canonical URL, but underscores would be perfectly reasonable too.
- Interface docs use the actual interface name -- hyphen vs underscore matters here, they're potentially different named interfaces from Juju's perspective, but we would never permit two interfaces to only differ by punctuation at the ecosystem level. 

Hyphens vs underscores are a common source of confusion in creating working links to library and interface docs (e.g. coming up frequently in PRs). Our contributor docs and PR reviews aim to avoid errors when adding URLs, but this is hard to check automatically (e.g. the docs URL in a PyPI README.md to the docs may not exist on the docs site at time of PR merge), and even if we get that solved, we can't control the links committed outside this repository. Automatically ensuring users end up on the right canonical page would alleviate this issue.

Resolves #541.


.docs/tests/test_automatic_redirects.py
```diff
@@ -0,0 +1,168 @@
+# Copyright 2025 Canonical Ltd.
```
tonyandrewmeyer:
```suggestion
# Copyright 2026 Canonical Ltd.
```

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
+# Copyright 2025 Canonical Ltd.
```
tonyandrewmeyer:
```suggestion
# Copyright 2026 Canonical Ltd.
```

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+
+def _automatic_redirects(
+    app: sphinx.application.Sphinx, env: sphinx.environment.BuildEnvironment
+) -> None:
+    # Don't create redirects during the per-package passes.
+    if app.config.package is not None:
+        return
+    redirects = _build_redirects(set(env.found_docs))
+    target = typing.cast('dict[str, str]', app.config.rediraffe_redirects)
+    assert isinstance(target, dict), f'rediraffe_redirects must be a dict, not {target}'
```
tonyandrewmeyer: Since the complaint is the type, do you think it would be clearer to have the type in the error rather than whatever the str is? Or even just the repr maybe?

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+    if not set(target).isdisjoint(redirects.keys()):
```
tonyandrewmeyer: It seems odd that one of these implicitly uses "iter dict is keys" and the other has it explicit. Is that to satisfy some type check?

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+        lines = ['Manually configured redirects would be clobbered by automatic redirects:']
+        lines.extend(
+            f'{k} -> {target[k]} would be overwritten by {k} -> {v}'
+            for k, v in redirects.items()
+            if k in target
+        )
+        raise ValueError('\n - '.join(lines))
```
tonyandrewmeyer, line 57: I don't love a multi-line message in a ValueError. Where does this show up, in the build output? Do you have an example of what that looks like? If it looks good and is the only place it would be expected, I would be convinced I think. Same for the ones below.

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+    import sphinx.environment
+
+
+def setup(app: sphinx.application.Sphinx) -> dict[str, str | bool]:
+    """Sphinx extension entrypoint -- register the ``env-updated`` hook."""
+    app.connect('env-updated', _automatic_redirects)
+    return {'version': '1.0.0', 'parallel_read_safe': True, 'parallel_write_safe': True}
+
+
+def _automatic_redirects(
```
tonyandrewmeyer, line 38: Is this the best name? It feels like most of what it's doing is validation.

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+    Categories area also separately aliased with both variants and with no separator.
+    """
+    redirects: dict[str, str] = {}
+    for docname in sorted(found_docs):
+        category, _, doc = docname.partition('/')
+        category_variants = {
+            category,
+            category.replace('-', '_'),  # how_to
+            category.replace('_', '-'),  # how-to
+            category.replace('_', '').replace('-', ''),  # howto
```
tonyandrewmeyer: Does this mean it's not just "how-to" -> "howto", but also all other `_` and `-` removal cases, like "my-first-trace" -> "myfirsttrace"?

I understand the motivation for hyphen/underscore redirects, but I'm less sure about removing them. Is this an issue we see in practice?

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+                    redirects[alias] = docname.removesuffix('index.html')
+                else:
+                    redirects[alias] = docname
+    return redirects
+
+
+def _separator_variant(docname: str) -> str:
+    """Return the docname with ``-`` and ``_`` swapped."""
+    if '-' in docname:
+        assert '_' not in docname, f"Docname {docname} should not contain both '-' and '_'"
```
tonyandrewmeyer: I feel this ought to be a ValueError not an assertion. Also below.

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+        for category_variant in sorted(category_variants):
+            for doc_variant in sorted({doc, _separator_variant(doc)}):
+                if (category_variant, doc_variant) == (category, doc):
+                    continue
+                alias = f'{category_variant}/{doc_variant}'
+                assert alias not in found_docs, f'Alias {alias} is a real page!'
+                assert alias not in redirects, (
+                    f'Alias {alias} already redirects to {redirects[alias]}'
+                )
+                if docname.endswith('/index.html'):
```
tonyandrewmeyer: Do we need to handle the llms.txt markdown files here too?

tonyandrewmeyer (commented):
Do you have some examples that would be easy to test this on?

.docs/extensions/automatic_redirects.py
james-garner-canonical: I think the actual object is clearer than just the type, but good call on using `repr`.

```suggestion
    assert isinstance(target, dict), f'rediraffe_redirects must be a dict, not {target!r}'
```

.docs/extensions/automatic_redirects.py
james-garner-canonical: I try to use the convention of having `setup` wire a single Sphinx event up to a function named after the the purpose of the extension itself. The top-level flow of the extension is: build the automatic redirects, load the manual redirects, check for conflicts, apply. Most of that by lines in this function is checking for conflicts, but I think that's OK.

.docs/extensions/automatic_redirects.py
james-garner-canonical: Just poor consistency on my part while incrementally building this up. Because of `.values()` being used, I think it's clearer to use `.keys()` explicitly, so I'll do that consistently.

```suggestion
    if not set(target.keys()).isdisjoint(redirects.keys()):
```

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,106 @@
...
+        lines = [
+            'The following manually configured redirects would chain with automatic redirects:'
+        ]
+        lines.extend(
+            f'{k} -> {v} should point to {redirects[v]}'
+            for k, v in target.items()
+            if v in redirects
+        )
+        raise ValueError('\n - '.join(lines))
+    if not set(target).isdisjoint(redirects.values()):
```
james-garner-canonical:
```suggestion
    if not set(target.keys()).isdisjoint(redirects.values()):
```

.docs/extensions/automatic_redirects.py
james-garner-canonical: Yeah, if you add manual `rediraffe_redirects` entries that would interfere badly with the automatic redirects, the build will fail, and you'll get an error at the end like this:
```
Extension error (automatic_redirects):
Handler <function _automatic_redirects at 0x7286bf35e560> for event 'env-updated' threw an exception (exception: Manually configured redirects would be clobbered by automatic redirects:
 - howto/add_library_docs -> foo would be overwritten by howto/add_library_docs -> how-to/add-library-docs
 - howto/customize_integration_tests -> bar would be overwritten by howto/customize_integration_tests -> how-to/customize-integration-tests)
error: Recipe `html` failed on line 14 with exit code 2
```
The final line comes from `just html` directly.

You also get a big nasty traceback above that of course, because Python.

<details>

<summary>More output</summary>

```
# Other happy output above ...
getting Git timestamps for dependencies... [100%] tutorials
WARNING: Skipping build combination due to build error
sphinx-sitemap: No pages generated for doc-sitemap.xml

Traceback (most recent call last):
  File "/home/james.garner@canonical.com/.cache/uv/archive-v0/4IXEopcRAOzo8WNu1my7G/lib/python3.14/site-packages/sphinx/events.py", line 98, in emit
    results.append(listener.handler(self.app, *args))
                   ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/home/james.garner@canonical.com/code/james-garner-canonical/charmlibs/26-06+docs+redirect-underscores/.docs/extensions/automatic_redirects.py", line 48, in _automatic_redirects
    raise ValueError('\n - '.join(lines))
ValueError: Manually configured redirects would be clobbered by automatic redirects:
 - howto/add_library_docs -> foo would be overwritten by howto/add_library_docs -> how-to/add-library-docs
 - howto/customize_integration_tests -> bar would be overwritten by howto/customize_integration_tests -> how-to/customize-integration-tests

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/james.garner@canonical.com/.cache/uv/archive-v0/4IXEopcRAOzo8WNu1my7G/lib/python3.14/site-packages/sphinx/cmd/build.py", line 337, in build_main
    app.build(args.force_all, args.filenames)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/james.garner@canonical.com/.cache/uv/archive-v0/4IXEopcRAOzo8WNu1my7G/lib/python3.14/site-packages/sphinx/application.py", line 378, in build
    self.builder.build_update()
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/home/james.garner@canonical.com/.cache/uv/archive-v0/4IXEopcRAOzo8WNu1my7G/lib/python3.14/site-packages/sphinx/builders/__init__.py", line 297, in build_update
    self.build(to_build,
    ~~~~~~~~~~^^^^^^^^^^
               summary=__('targets for %d source files that are out of date') %
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
               len(to_build))
               ^^^^^^^^^^^^^^
  File "/home/james.garner@canonical.com/.cache/uv/archive-v0/4IXEopcRAOzo8WNu1my7G/lib/python3.14/site-packages/sphinx/builders/__init__.py", line 318, in build
    updated_docnames = set(self.read())
                           ~~~~~~~~~^^
  File "/home/james.garner@canonical.com/.cache/uv/archive-v0/4IXEopcRAOzo8WNu1my7G/lib/python3.14/site-packages/sphinx/builders/__init__.py", line 462, in read
    for retval in self.events.emit('env-updated', self.env):
                  ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/james.garner@canonical.com/.cache/uv/archive-v0/4IXEopcRAOzo8WNu1my7G/lib/python3.14/site-packages/sphinx/events.py", line 109, in emit
    raise ExtensionError(
    ...<4 lines>...
    ) from exc
sphinx.errors.ExtensionError: Handler <function _automatic_redirects at 0x7286bf35e560> for event 'env-updated' threw an exception (exception: Manually configured redirects would be clobbered by automatic redirects:
 - howto/add_library_docs -> foo would be overwritten by howto/add_library_docs -> how-to/add-library-docs
 - howto/customize_integration_tests -> bar would be overwritten by howto/customize_integration_tests -> how-to/customize-integration-tests)

Extension error (automatic_redirects):
Handler <function _automatic_redirects at 0x7286bf35e560> for event 'env-updated' threw an exception (exception: Manually configured redirects would be clobbered by automatic redirects:
 - howto/add_library_docs -> foo would be overwritten by howto/add_library_docs -> how-to/add-library-docs
 - howto/customize_integration_tests -> bar would be overwritten by howto/customize_integration_tests -> how-to/customize-integration-tests)
error: Recipe `html` failed on line 14 with exit code 2
```

</details>

.docs/extensions/automatic_redirects.py
james-garner-canonical: Thanks for flagging this branch. Some digging discovered:
- This extension runs *before* `/index.html` shenanigans, including the 'point `foo/index.html` to `foo/` bit'.
- We can't cleanly mess with extra redirects for paths ending in `/index.html.md` (ends up with `index.html/index.html` nonsense).

I've tidied up to reflect this.

The preview RTD build indicates that these redirects don't apply to the `/index.html.md` files (mangled URLs just end up on a 'page not found' page). I'm OK with this, since `llms.txt` contains the correct canonical URLs.

.docs/extensions/automatic_redirects.py
james-garner-canonical: Turns out mixed paths are actually a potentially valid thing, like `category/canonical_interface_name/some-hyphenated-doc-name`. So I'll need to rethink this approach.

.docs/extensions/automatic_redirects.py
james-garner-canonical: The motivation for this is because we use `how-to` as the path for how to guides. Other docs (like `ops`) use `howto`. I figured with a canonical URL like `how-to/foo-bar`, we should support both `how_to` and `howto` as the categories. Then I wrote it in a way that's agnostic about whether the canonical form contains underscores or hyphens (or both). Maybe initially I should just aim for `all-hyphens` and `all_underscores`, because dropping them for the category is kind of scope creep (and also wants to live one logical level up from hyphens vs underscores, since we want `howto/foo-bar` *and* `howto/foo_bar` to redirect to `how-to/foo-bar` like `how_to/foo_bar` does.

james-garner-canonical: I switched to a generator based approach which handles the `howto` case cleanly (IMO).

james-garner-canonical:
Thanks for the review @tonyandrewmeyer. I've reworked the PR based on your feedback and some additional experimentation.

> Do you have some examples that would be easy to test this on?

You can try the redirects out in the preview build. For example, with this canonical URL https://canonical-juju-charmlibs--540.com.readthedocs.build/how-to/manage-libraries/ the following redirects all work:
- https://canonical-juju-charmlibs--540.com.readthedocs.build/howto/manage-libraries/
- https://canonical-juju-charmlibs--540.com.readthedocs.build/howto/manage_libraries/
- https://canonical-juju-charmlibs--540.com.readthedocs.build/how_to/manage_libraries/

Note that the redirects end up without `/index.html` trimmed, just like if you manually go to https://canonical-juju-charmlibs--540.com.readthedocs.build/how-to/manage-libraries/index.html

Note also that redirects *from* the `/index.html` suffixed pages just work:
- https://canonical-juju-charmlibs--540.com.readthedocs.build/howto/manage-libraries/index.html
- https://canonical-juju-charmlibs--540.com.readthedocs.build/howto/manage_libraries/index.html
- https://canonical-juju-charmlibs--540.com.readthedocs.build/how_to/manage_libraries/index.html

.docs/tests/test_automatic_redirects.py
```diff
@@ -0,0 +1,162 @@
...
+import typing
+
+import automatic_redirects
+import pytest
+
+if typing.TYPE_CHECKING:
+    import sphinx.application
+
+
+class TestrVariants:
```
tonyandrewmeyer: Is this a deliberate name? I assume "r" is for "redirect" but could we use the full word?

tonyandrewmeyer (approved):

.docs/extensions/automatic_redirects.py
tonyandrewmeyer: > The motivation for this is because we use `how-to` as the path for how to guides. Other docs (like `ops`) use `howto`.

It seems like there should be Canonical consistency for this...

> I figured with a canonical URL like `how-to/foo-bar`, we should support both `how_to` and `howto` as the categories.

but if there isn't (yet?) then I guess this is indeed r reasonable.

> Then I wrote it in a way that's agnostic about whether the canonical form contains underscores or hyphens (or both).

This felt a bit excessive, but it's not a blocker for me.

.docs/tests/test_automatic_redirects.py
james-garner-canonical: Just a sign that I wrote the name by hand 😅
```suggestion
class TestVariants:
```

.docs/extensions/automatic_redirects.py
```diff
@@ -0,0 +1,105 @@
...
+        lines = ['The following automatic redirects chain with manually configured redirects:']
+        lines.extend(f'{k} -> {v} -> {target[v]}' for k, v in redirects.items() if v in target)
+        raise ValueError('\n - '.join(lines))
+    target.update(redirects)
+
+
+def _build_redirects(found_docs: set[str]) -> dict[str, str]:
+    """Redirect underscored names to hyphenated ones and vice versa.
+
+    Categories area also separately aliased with both variants and with no separator.
```
james-garner-canonical:
```suggestion
    Categories are also separately aliased with both variants and with no separator.
```

.docs/extensions/automatic_redirects.py
dwilding: I'm not so sure we should support redirecting `howto`. The standard we should follow now is `how-to` (see [Sphinx Stack](https://github.com/canonical/sphinx-stack/tree/main/docs/how-to)), so Ops is wrong. I'm tempted to define a blanket redirect in Ops for `howto/*` to `how-to/*` - if we hadn't just added another layer of redirects with the latest move...

For Charmlibs, my preference is to let `howto` URLs be a 404. Unless `howto` vs `how-to` has been causing a lot of problems for people.

james-garner-canonical: Does the Sphinx stack using `how-to` really mean that existing sites using `howto` are now *wrong*?

A scan of my (now several month old) checkout of all Canonical's repositories shows the following:
```
Only 'howto':   35
Only 'how-to':  201
Both:           4
```
So `how-to` is overwhelmingly more popular. `howto/foo` going to the right place rather than 404-ing doesn't seem unreasonable, but I guess our traffic is low enough overall that it hardly matters, so I'll drop the `howto` redirects.

BTW the 'both' repos seem to be repositories that manage multiple sets of documentation, with some using `how-to` and others using `howto`.

PR #545, opened 2026-06-22
ci: exclude certain 'global' files from triggering all package tests

james-garner-canonical:
This PR adds the ability to exclude certain files from triggering all package tests. For instance, currently anything changed under `.github/` triggers all tests to run. This makes perfect sense when workflows are changed, but doesn't really provide any value if PR templates are what's changing.

The exclusions are a static list of full file paths (from the repository root) to avoid unintended matches.

This doesn't change `CODEOWNERS`, so the maintainers are still requested for review as appropriate.

Currently, the PR adds the `PULL_REQUEST_TEMPLATE` files only. But maybe it should add some more, like `.github/dependabot.yaml`. I'm open to suggestions.

I verified that the feature works by temporarily excluding the `get-changed.py` script itself from triggering global changes.

### DIFF
--- a/.github/get-changed.py
+++ b/.github/get-changed.py
@@ -37,6 +37,7 @@ _EXCLUSIONS = {
     '.github/PULL_REQUEST_TEMPLATE/adding-a-new-library.md',
     '.github/PULL_REQUEST_TEMPLATE/blank.md',
     '.github/PULL_REQUEST_TEMPLATE/migrating-a-library.md',
+    '.github/get-changed.py',
 }
 _REPO_ROOT = pathlib.Path(__file__).parent.parent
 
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (approved):
Maybe better to glob `.github/PULL_REQUEST_TEMPLATE*` to avoid needing to update this with new templates? 

Other candidates: `*.md` in the root, `LICENCE`, `check-conventional-pr-title.py`, dependabot and Zizmor config. Maybe `.example` unless that does have tests that are meant to run?

james-garner-canonical:
> Maybe better to glob `.github/PULL_REQUEST_TEMPLATE*` to avoid needing to update this with new templates?

I'll stick with hardcoded paths for simplicity of implementation and to keep things explicit for now.

> Other candidates: `*.md` in the root, `LICENCE`, `check-conventional-pr-title.py`, dependabot and Zizmor config. Maybe `.example` unless that does have tests that are meant to run?

- The root files like `LICENSE` and other markdown files are excluded by default (not in the `GLOBAL_FILES` set).
- I guess `check-conventional-pr-title.py` really doesn't need to trigger all CI, but I'm hesitant to add it since it's not strictly obvious just from the file path (you need to know what the script does -- which to be fair is pretty clear from the name). So I'll hold off on this one for now.
- I've added dependabot and Zizmor config, thanks.
- `.example` does have tests that are meant to run.

PR #550, opened 2026-06-23
ci: normalize extension for zizmor config

james-garner-canonical:
Per [the zizmor docs](https://docs.coderabbit.ai/tools/zizmor#configuration), the `.yaml` file extensions is perfectly acceptable. Per discussion elsewhere, it's nice to keep these normalized across the repo and our wider ecosystem, and the `.yaml` file extension is far more common (`charmcraft.yaml`, `interface.yaml`, `.github/workflows/*.yaml`, etc).


tonyandrewmeyer:
> Per [the zizmor docs](https://docs.coderabbit.ai/tools/zizmor#configuration), the `.yaml` file extensions is perfectly acceptable. 

This is true, except the link should be https://docs.zizmor.sh/configuration/

tonyandrewmeyer:
Changed in [1.19.0](https://github.com/zizmorcore/zizmor/releases/tag/v1.19.0).

tonyandrewmeyer (approved):
I'm very in favour of consistency. My issue here is that we consistently use `zizmor.yml` across Charm Tech (because until late last year it was the only supported default name).

Using `zizmor.yaml` across all Charm Tech repos would be most consistent with dependabot and our workflows. Using `zizmor.yaml` only in this repo would be worse than the status quo, in my opinion.

So if you open the same PR in all the other non-archived Charm Tech repos (other than charm-tech itself) I'm in favour of this. You can get me to review them all. If only here, then ideally consider my approval as changes-requested.

james-garner-canonical:
Not worth opening 14 more PRs for.

tonyandrewmeyer:
I believe it's only 9, and AI can do all the work. I'll do it and you can review.

PR #551, opened 2026-06-23
ci: exclude zizmor.yml from global change detection

james-garner-canonical:
This PR excludes `zizmor.yml` rather than `zizmor.yaml` from global change detection -- both extensions are supported (per https://docs.zizmor.sh/configuration/), but we'll stick with `zizmor.yml` in this repository until (if) all [charm tech repos](https://github.com/orgs/canonical/teams/charm-tech/repositories) migrate at the same time (per #550).


james-garner-canonical:
Closed in favour of #550.

PR #553, opened 2026-06-23
docs: update Charmlibs URLs for certificate_transfer, tls-certificates

dwilding:

I haven't bumped the package versions - I'll leave that up to the TLS team.


kayra1 (approved):
thanks

PR #555, opened 2026-06-24
docs: update Charmlibs URLs for Service Mesh team packages

dwilding:

I haven't bumped the package versions - I'll leave that up to the Service Mesh team.


adhityaravi (approved):
LGTM! And thanks very much! I'm not bumping the versions, it can be a part of the next code change

PR #556, opened 2026-06-24
docs: update Charmlibs URL for k8s_backup_target

dwilding:

I haven't bumped the package version - I'll leave that up to the Analytics team.


deusebio (approved):
LGTM

PR #558, opened 2026-06-24
docs: update Charmlibs URLs for sloth

dwilding:

I haven't bumped the package version - I'll leave that up to the Tracing and Profiling team.


mmkay (approved):
Thank you!

PR #560, opened 2026-06-25
ci: update workshop definitions

james-garner-canonical:
I initially defined `functional` actions for all the workshops defined in this repository. This seemed like a convenient way to wrap up the need for `sudo` and also the `just python=... functional` justfile variable syntax for use. With the switch to using a regular option for the Python version, like `just functional --python=...` I'm not convinced that the actions provide value. This PR drops the action definitions and updates the docs accordingly.

### DIFF
--- a/.workshop/jammy.yaml
+++ b/.workshop/jammy.yaml
@@ -2,7 +2,3 @@ name: jammy
 base: ubuntu@22.04
 sdks:
   - name: project-charmlibs
-actions:
-  functional: |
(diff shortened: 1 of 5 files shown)


tonyandrewmeyer (approved):
It's not clear to me whether project workshops are meant to provide actions that just wrap ones available via some task runner already, or if you are meant to just execute those. 

The docs say "For complex commands that you run often, define actions in your workshop definition to invoke them with workshop run". These fall under "run often" but are not complex given that the complexity is hidden inside the justfile, I guess.

So seems good to me.

PR #561, opened 2026-06-27
ci: add dependency-review-action on PRs

tonyandrewmeyer:
Scans dependency manifest changes introduced in pull requests against the OSV vulnerability database and the configured licence policy, and comments on the PR when a high-severity vulnerability is found so it can be addressed before merge.

### DIFF
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -1,0 +1,25 @@
+name: Dependency review
+
+on:
+  pull_request:
+    branches:
+      - main
+
+permissions: {}
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: Dependency review
+        uses: actions/dependency-review-action@v4
+        with:
+          fail-on-severity: high
+          comment-summary-in-pr: on-failure
--- a/.github/workflows/dependency-review.yaml
+++ b/.github/workflows/dependency-review.yaml
@@ -3,7 +3,8 @@ name: Dependency review
 on:
   pull_request:
     branches:
-      - main
+      - 'main'
+      - '*-maintenance'
 
 permissions: {}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yaml
```diff
@@ -0,0 +1,25 @@
+name: Dependency review
+
+on:
+  pull_request:
+    branches:
+      - main
```
james-garner-canonical: Why only against `main`?

.github/workflows/dependency-review.yaml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+        with:
+          persist-credentials: false
+      - name: Dependency review
+        uses: actions/dependency-review-action@v4
+        with:
+          fail-on-severity: high
+          comment-summary-in-pr: on-failure
```
james-garner-canonical: How helpful is having the CI result as a comment rather than as a CI result? I wonder if we'd be better off sticking with the default value of `never` here. This would mean we only need `contents: read` permissions, which would be nice.

tonyandrewmeyer: Ah, that's not meant to be there any more, after it was problematic in api_demo_server. I thought I fixed the branches I had, but I missed some.

tonyandrewmeyer:
```suggestion
permissions: {}

jobs:
  dependency-review:
    runs-on: ubuntu-latest
    permissions:
      contents: read
    steps:
      - name: Checkout repository
        uses: actions/checkout@v6
        with:
          persist-credentials: false
      - name: Dependency review
        uses: actions/dependency-review-action@v4
        with:
          fail-on-severity: high
```

.github/workflows/dependency-review.yaml
tonyandrewmeyer: If the question is why not also the `20.04-maintenance` branch, because I forgot about that.

If the question is why not all branches, then it's the recommended approach I found when researching this - the protection matters when introducing license violations or security issues to branches that are used for releasing, so a short-lived feature branch check is mostly noise. I could make it all branches if you preferred.

```suggestion
    branches:
      - 'main'
      - '*-maintenance'
```

james-garner-canonical: Your `suggestion` seems right to me. I'd suggest the same approach for Ops.

james-garner-canonical (approved):
Let's try this, it seems like a good idea to catch this kind of issue preemptively.

PR #562, opened 2026-06-29
docs: update migration guide to suggest commiting

### LINKED ISSUE #309: We should document how to organise commits to make library migration PRs easier to review

james-garner-canonical:
This PR updates the migration guide to document the suggested commit strategy for ease of review. This is also mentioned in the PR template (we correct a numbering error in the template in this PR too).

Resolves #309.

### DIFF
--- a/.docs/how-to/migrate.md
+++ b/.docs/how-to/migrate.md
@@ -85,6 +85,20 @@ In this case:
 
 You'll also want to check for any config files under the old `<interface name>` directory (for example, a `ruff.toml` file), and incorporate any applicable settings into your project's `pyproject.toml`.
 
+```{admonition} Committing as you go
+:class: important
+
+PRs into the monorepo are squash-merged, so the PR title becomes the final commit message on `main`.
+The commits on your branch are just for review, so the goal is simply to make the diff easy to follow.
+The most helpful pattern is to keep mechanical changes (copying files verbatim, fixing imports, running the formatter) in their own commits, separate from changes that need real thought.
+That way a reviewer can skip straight past the noise.
+
+A sequence that works well is: scaffolding, then a verbatim copy of the library, then import fixes, then linting and formatting.
+As you work through this guide we'll suggest good moments to commit, but treat these as a starting point -- split, reorder, or rename the commits however makes your change clearest (for example, you might give your docs their own copy-and-tidy commits).
+
+To start with, commit the output of `just init` now, before making any other changes.
+```
+
 ## Migrate your library's code
 
 This is the easy bit, since Charmhub-hosted libs are only a single module.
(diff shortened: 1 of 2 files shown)


.docs/how-to/migrate.md
```diff
@@ -85,6 +85,20 @@ In this case:
 
 You'll also want to check for any config files under the old `<interface name>` directory (for example, a `ruff.toml` file), and incorporate any applicable settings into your project's `pyproject.toml`.
 
+```{admonition} Committing as you go
+:class: important
```
dwilding: I don't think this works so well in note style. How about making it a section at the same level as "Interface libraries"? I'd call the section "Commit as you go"

.docs/how-to/migrate.md
```diff
@@ -85,6 +85,20 @@ In this case:
...
+
+PRs into the monorepo are squash-merged, so the PR title becomes the final commit message on `main`.
+The commits on your branch are just for review, so the goal is simply to make the diff easy to follow.
+The most helpful pattern is to keep mechanical changes (copying files verbatim, fixing imports, running the formatter) in their own commits, separate from changes that need real thought.
+That way a reviewer can skip straight past the noise.
+
+A sequence that works well is: scaffolding, then a verbatim copy of the library, then import fixes, then linting and formatting.
+As you work through this guide we'll suggest good moments to commit, but treat these as a starting point -- split, reorder, or rename the commits however makes your change clearest (for example, you might give your docs their own copy-and-tidy commits).
```
dwilding: A couple of suggestions for easier reading. Suggest using numbers for the sequence so that it stands out better.

```suggestion
A sequence that works well:

1. Scaffolding
2. Verbatim copy of the library
3. Fix imports
4. Linting and formatting

As you work through this guide we'll suggest good moments to commit, but treat these as a starting point -- split, reorder, or rename the commits however makes your change clearest. For example, you might give your docs their own copy-and-tidy commits.
```

.docs/how-to/migrate.md
```diff
@@ -108,6 +122,16 @@ Download a copy of the latest release of your library, and add it to your new pa
 ````
 `````
 
+Copy the module across verbatim, without making any changes to it yet. This gives reviewers a clean baseline to compare against the original Charmhub-hosted library.
+
+```{admonition} Committing as you go
```
dwilding: I think it makes sense to have these special tips as markers. Although I would call them "Commit as you go" to match the initial section.

The subsequent tip about adding dependencies feels out of place now and breaks the flow. Could it be added (indented) as part of step 5 below?

.docs/how-to/migrate.md
```diff
@@ -169,6 +196,13 @@ To speed things up, only build the reference docs for your library:
 just docs html <library path>
 ```
 
+```{admonition} Committing as you go
+:class: tip
+
+Linting and formatting touch a lot of lines but rarely need careful review, so they're worth keeping in a commit of their own.
+Include any `pyproject.toml` config or lint ignores you added to avoid unwanted changes here too.
```
dwilding: I think this would be fine

```suggestion
Include any `pyproject.toml` config changes or lint ignores.
```

.docs/how-to/migrate.md
dwilding: I did a test build to check that indenting the code block works. (Same with adding the dependencies tip as part of step 5)

````suggestion
6. Import the public API of your library to `__init__.py` and add the imported names to `__all__`, like this:

    ```python
    # immediately before or after from ._version
    # (imports are sorted alphabetically)
    from ._<name> import (
        # your library's public API
    )

    ...

    __all__ = [
        # the names we imported, as strings
    ]
    ```
````

dwilding (approved):
Nice improvements 🙂

.docs/how-to/migrate.md
james-garner-canonical: Good call, done.

.docs/how-to/migrate.md
james-garner-canonical: Nice, I didn't realise that was possible. Done (and for 5).

PR #563, opened 2026-06-29
ci: only print colours when appropriate

### LINKED ISSUE #523: Optimize CI output for agents and piping
The `just` quick start help uses ANSI color codes in its message, as do the messages for `just init` and `just init --interface`. We should detect if we're in an interactive session and adjust the output accordingly. For the `init` commands, interactivity can only be skipped if `--no-input` is passed to `cookiecutter`, but that's probably orthogonal to piping since input could get piped in ...

A good general approach might be the recipe suggested by https://bixense.com/clicolors/
```py
import os, sys

def has_colors():
    if os.environ.get('NO_COLOR'):
        return False
    elif os.environ.get('CLICOLOR_FORCE'):
        return True
    return sys.stdout.isatty()
```

james-garner-canonical:
This PR updates `just.py` to emit ANSI colour codes only when `FORCE_COLOR` is set or the output is a TTY.

Resolves #523.

### DIFF
--- a/.scripts/just.py
+++ b/.scripts/just.py
@@ -50,32 +50,59 @@ TEST_REQUIREMENTS = REPO_ROOT / 'test-requirements.txt'
 
 DEFAULT_PYTHON = '3.10'
 
-# ANSI escape codes for formatting the messages.
-BOLD = '\033[1m'
-NORMAL = '\033[0m'
-CYAN = '\033[36m'
-# Quick start message printed when called with no arguments.
-QUICK_START = f"""
-{BOLD}Charmlibs is Canonical's charm library monorepo{NORMAL}
 
-{BOLD}List all commands with {CYAN}just help{NORMAL}{BOLD}, or:{NORMAL}
-- Create a new package: {CYAN}just init{NORMAL} or {CYAN}just init --interface{NORMAL}
-- Add a dependency to a package: {CYAN}just add <package> <dependency>{NORMAL}
-- Lint, unit test, and build docs for a package: {CYAN}just check <package>{NORMAL}
-- Run {CYAN}ruff{NORMAL} for all packages: {CYAN}just fast-lint{NORMAL}
+class Colors:
+    """ANSI escape codes for formatting messages, empty strings when color is disabled.
 
-{BOLD}Run individual checks for a package:{NORMAL}
-- {CYAN}just lint <package>{NORMAL} (fix errors with {CYAN}just format <package>{NORMAL})
-- {CYAN}just unit <package>{NORMAL}
-- {CYAN}just functional <package>{NORMAL} (may require extra software like {CYAN}pebble{NORMAL})
+    Whether color is enabled follows https://bixense.com/clicolors/: `NO_COLOR` disables color,
+    `CLICOLOR_FORCE` forces it on, otherwise color is only used when stdout is an interactive
+    terminal (so piped or redirected output stays plain text).
+    """
 
-{BOLD}Run integration tests{NORMAL} (requires a Juju controller and a cloud):
-- Pack: {CYAN}just pack-k8s <package>{NORMAL} or {CYAN}just pack-machine <package>{NORMAL}
-- Run: {CYAN}just integration-k8s <package>{NORMAL} or {CYAN}just integration-machine{NORMAL}
+    def __init__(self) -> None:
+        if self._enabled():
+            self.bold = '\033[1m'
+            self.normal = '\033[0m'
+            self.cyan = '\033[36m'
+        else:
+            self.bold = self.normal = self.cyan = ''
 
-{BOLD}Build the docs: {CYAN}just docs{NORMAL}
-- For specific packages only: {CYAN}just docs html <packages>{NORMAL}
+    @staticmethod
+    def _enabled() -> bool:
+        """Return whether to emit ANSI color codes."""
+        if os.environ.get('NO_COLOR'):
+            return False
+        if os.environ.get('CLICOLOR_FORCE'):
+            return True
+        return sys.stdout.isatty()
+
+
+def _quick_start() -> str:
+    """Return the quick start message printed when called with no arguments."""
+    c = Colors()
+    bold, normal, cyan = c.bold, c.normal, c.cyan
+    return f"""
+{bold}Charmlibs is Canonical's charm library monorepo{normal}
+
+{bold}List all commands with {cyan}just help{normal}{bold}, or:{normal}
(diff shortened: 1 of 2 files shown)


.scripts/just.py
```diff
@@ -47,60 +47,86 @@
...
 
-def _register(fn: _F) -> _F:
-    RECIPES[fn.__name__.removeprefix('_').replace('_', '-')] = fn
-    return fn
+    @staticmethod
+    def _enabled() -> bool:
+        """Return whether to emit ANSI color codes."""
+        if os.environ.get('NO_COLOR'):  # https://no-color.org/ (present and non-empty)
+            return False
+        if 'FORCE_COLOR' in os.environ:  # https://force-color.org/ (present, empty string ok)
```
tonyandrewmeyer: My understanding is that this behaviour is inconsistent. Node will only accept 1,2,3 I think, for example. Even the site you link to here says that the empty string should not force colour. I'd prefer that both the empty string and "0" were excluded, but just the empty string would be ok.

.scripts/just.py
```diff
@@ -47,60 +47,86 @@
...
+        else:
+            self.bold = self.normal = self.cyan = ''
 
-def _register(fn: _F) -> _F:
-    RECIPES[fn.__name__.removeprefix('_').replace('_', '-')] = fn
-    return fn
+    @staticmethod
+    def _enabled() -> bool:
+        """Return whether to emit ANSI color codes."""
+        if os.environ.get('NO_COLOR'):  # https://no-color.org/ (present and non-empty)
```
tonyandrewmeyer:
```
$ curl https://no-color.org
curl: (6) Could not resolve host: no-color.org
```

Doesn't look like [just me](https://downforeveryoneorjustme.com/no-color.org).

.scripts/just.py
```diff
@@ -47,60 +47,86 @@
...
     func = RECIPES.get(name)
     if func is None:
         print(f'Unknown recipe: {name!r}\n', file=sys.stderr)
-        print(QUICK_START)
+        print(_quick_start())
         return 2
     return func(args)
 
 
+# --- Main --------------------------------------------------------------------------------------
```
tonyandrewmeyer: Seems a little odd that this is *after* `def main()`.

.scripts/just.py
```diff
@@ -47,60 +47,86 @@
...
+{c.bold}List all commands with {c.cyan}just help{c.normal}{c.bold}, or:{c.normal}
+- Create a new package: {c.cyan}just init{c.normal} or {c.cyan}just init --interface{c.normal}
+- Add a dependency to a package: {c.cyan}just add <package> <dependency>{c.normal}
+- Lint, unit test, and build docs for a package: {c.cyan}just check <package>{c.normal}
+- Run {c.cyan}ruff{c.normal} for all packages: {c.cyan}just fast-lint{c.normal}
+
+{c.bold}Run individual checks for a package:{c.normal}
+- {c.cyan}just lint <package>{c.normal} (fix errors with {c.cyan}just format <package>{c.normal})
+- {c.cyan}just unit <package>{c.normal}
+- {c.cyan}just functional <package>{c.normal} (may need sudo and additional software installed)
```
tonyandrewmeyer, line 106: I assume the wording change here is a driveby?

.scripts/just.py
```diff
@@ -47,60 +47,86 @@
...
+            self.bold = self.normal = self.cyan = ''
 
-def _register(fn: _F) -> _F:
-    RECIPES[fn.__name__.removeprefix('_').replace('_', '-')] = fn
-    return fn
+    @staticmethod
+    def _enabled() -> bool:
+        """Return whether to emit ANSI color codes."""
+        if os.environ.get('NO_COLOR'):  # https://no-color.org/ (present and non-empty)
+            return False
```
tonyandrewmeyer: I'm not sure what the most correct thing here is, but at the site you link to below force will override no.

.scripts/just.py
```diff
@@ -47,60 +47,86 @@
...
+            self.bold = '\033[1m'
+            self.normal = '\033[0m'
+            self.cyan = '\033[36m'
+        else:
+            self.bold = self.normal = self.cyan = ''
 
-def _register(fn: _F) -> _F:
-    RECIPES[fn.__name__.removeprefix('_').replace('_', '-')] = fn
-    return fn
+    @staticmethod
```
tonyandrewmeyer, line 65: Can you get the quick start and help together? Either way, probably this could get a lru_cache or similar decorator to avoid the recalc?

tonyandrewmeyer (approved):
For what it's worth, I think `NO_COLOR` and `FORCE_COLOR` are better choices than `CLIFORCE_COLOR`. I don't think Juju has a force env (only arg) but Python uses FORCE_COLOR, I believe.

I assume the failing tests are unrelated.

.scripts/just.py
james-garner-canonical: An embarrassing hallucination generated when reordering the branches 🙈

Should be https://bixense.com/clicolors/

.scripts/just.py
james-garner-canonical: We never print both quick start and help. Since [os.environ is itself a caching layer generated once at import time](https://docs.python.org/3/library/os.html#os.environ), I don't think adding caching here is worth it.

.scripts/just.py
james-garner-canonical: Thanks for checking that! I'll update to more closely match the spec.

.scripts/just.py
tonyandrewmeyer: It presumably existed once (via https://github.com/jcs/no_color ?). A web search finds it, so maybe not even dead for that long?

james-garner-canonical: I suppose that makes the hallucination more forgivable :smile:

.scripts/just.py
james-garner-canonical: Thanks for checking this too. I was surprised when GLM said that an empty string should be treated as truthy, I should have double checked the site myself. Reading the site now, it's very clear that an empty string is falsey.

.scripts/just.py
james-garner-canonical: Wait https://bixense.com/clicolors/ links to https://no-color.org/ ...

.scripts/just.py
james-garner-canonical: I've gone with 'FORCE_COLOR' overriding 'NO_COLOR'. The example in the site is a bit subtle -- 'no' is checked first, but not eagerly returned, so 'force' presumably overrides. The interaction between the various options doesn't seem to be specced anywhere.

Issue #552, opened 2026-06-23
Publish test snaps for use in the charmlibs.snap functional tests


james-garner-canonical:
Currently and in the WIP 2.0, we use some snaps from the store in our functional tests. We do need to hit the store, and we do need snaps that use the full breadth of features that we support, but ideally we'd publish something as lightweight as possible, that we control.

james-garner-canonical:
Currently we only use custom local snaps or test snaps published by the snapd developers, so closing this for now.

PR #569, opened 2026-07-01
feat(snap): add snap apps module

james-garner-canonical:
This PR adds the snap library functionality using the `/v2/apps` endpoint, mirroring `snap start`, `snap stop` and `snap restart`.

### DIFF
--- a/snap/src/charmlibs/snap/_snapd_apps.py
+++ b/snap/src/charmlibs/snap/_snapd_apps.py
@@ -1,4 +1,4 @@
-# Copyright 2021 Canonical Ltd.
+# Copyright 2026 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -34,7 +34,7 @@ def start(snap: str, *services: str, enable: bool = False) -> None:
 
     Raises:
         AppNotFoundError: if the snap is not installed or the service is not found.
-        ChangeError: if the change fails after starting (for example, the service fails to start).
+        ChangeError: if the change fails (for example, the service fails to start).
     """
     names = [f'{snap}.{s}' for s in services] if services else [snap]
     data: dict[str, Any] = {'action': 'start', 'names': names}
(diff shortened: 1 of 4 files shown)


snap/src/charmlibs/snap/_snapd_apps.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def start(snap: str, *services: str, enable: bool = False) -> None:
+    """Start snap services.
+
+    Args:
+        snap: The name of the snap whose services to start.
+        services: Names of services within the snap to start. If omitted, all of the snap's
+            services are started.
+        enable: If ``True``, also enable the services to start automatically at boot.
```
tonyandrewmeyer, line 33: This seems a poor name, but I see it's the one in the CLI, so best to stick with it.

snap/tests/functional/test_snapd_apps.py
```diff
@@ -0,0 +1,266 @@
...
+    _snapd_apps.start(_SNAP, _SERVICE)  # Second start should not raise.
+
+
+def test_start_with_enable():
+    # start with enable=True re-enables a disabled service.
+    ensure_installed(_SNAP, classic=True)
+    _stop_and_disable()
+    assert not _service_is_enabled()
+    _snapd_apps.start(_SNAP, _SERVICE, enable=True)
+    # enabled persists even after kube-proxy exits quickly.
```
tonyandrewmeyer, line 97: There are a few comments that mention the snap specifically, but it's abstracted out into the global. Maybe the comments should be more generic too?

tonyandrewmeyer (approved):
Much more straightforward than the last one! Looks sensible to me, and seems to work fine.

snap/tests/functional/test_snapd_apps.py
james-garner-canonical: Well spotted. I'm going to leave them for now as I think they (some of them) document some important assumptions about the particular snap being used. They should all go away when we resolve #552.

snap/src/charmlibs/snap/_snapd_apps.py
james-garner-canonical: Yeah I'm not a fan either. Maybe we should revisit the decision to mirror the CLI names before the 2.0 release, once all the methods are in place and we can see how the whole API fits together.

PR #570, opened 2026-07-01
feat(snap): add snap interfaces module

james-garner-canonical:
This PR adds the snap library functionality using the `/v2/interfaces` endpoint, mirroring the snap CLI commands `snap connect` and `snap disconnect`.


snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,108 @@
...
+
+from typing import Any
+
+from . import _client, _errors
+
+# /v2/interfaces
+
+
+def connect(
+    plug_snap: str, plug: str, slot_snap: str | None = None, slot: str | None = None, /
```
tonyandrewmeyer: Why positional only? I think I would prefer keywords here, because it's awkwardly a pair of strs and then a one-or-two str 'pair'. Plug->slot works ok for people who normally read left-to-right, I think, but then you're relying on CLI familiarity or seeing the signature to know the order of the rest.

I feel like I would have gone with

```suggestion
    plug_snap: str, slot_snap: str | None = None, *, plug: str, slot: str | None = None,
```

Although with no default for plug, that is a little odd.

I think I'm ok with no kwonly, but I don't like positional only unless there's a motivation I'm missing.

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,108 @@
...
+    (the underlying ``interfaces-unchanged`` error is suppressed, mirroring the snap CLI).
+
+    Args:
+        plug_or_slot_snap: The snap providing the plug (explicit form) or the snap providing
+            the plug or slot to disconnect (two-argument form).
+        plug_or_slot: The plug on ``plug_or_slot_snap`` (explicit form) or the plug or slot
+            name to disconnect (two-argument form).
+        slot_snap: The snap providing the slot. Omit for the two-argument form.
+        slot: The slot on ``slot_snap``. May be omitted to match any slot on ``slot_snap``.
+        forget: If ``True``, also forget any manual connection preference, so the interface
```
tonyandrewmeyer: I'm not super familiar with snap interfaces. I assume people who are would understand this, but maybe we should help others out more? Is there a way to use the library to set a manual connection preference? Where would I go to learn more about what this means, as a library user?

Also, I think we would generally use `true` (meaning truthy) here, not `True`.

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,108 @@
...
+    plug_or_slot: str,
+    slot_snap: str | None = None,
+    slot: str | None = None,
+    /,
+    *,
+    forget: bool = False,
+) -> None:
+    """Disconnect a plug from a slot.
+
+    May be called in two forms:
```
tonyandrewmeyer: I really don't like this as a Python API style. Did this get discussed and resolved already when everything was in one PR?

kwonly would solve this, or two methods would solve this. I definitely need convincing that positional args can be entirely different objects based on the number provided.

tonyandrewmeyer (changes requested):
The code seems fine, but I'm not convinced about the API here, sorry. Am I forgetting this being discussed previously?

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: Connect and disconnect both have this basic problem. The CLI uses colon-separated strings for the snap and its plug/slot, like:
```
  snap connect [connect-OPTIONS] <snap>:<plug> [<snap>:<slot>]
  snap disconnect [disconnect-OPTIONS] <snap>:<plug> [<snap>:<slot>]
```

Translating this to Python, I thought that separate positional arguments made the most sense. But there are other options. We've resisted creating a reified `Snap` type elsewhere, so we should probably see how far we can get with that here (so no `Snap('foo').plug('bar').connect(Snap('baz').slot('bartholemew')`).

Sticking with the flat, functional design, a few ideas spring to mind:
- Four separate positional arguments (current implementation).
- Two colon-delimited-string positional arguments.
- Two positional arguments that can either be a string or a `tuple[str, str]`.

I do suspect that reducing to two arguments would be a net win.

---

For `connect`, the slot snap can be omitted entirely, or provided as a bare snap name (no slot specified).

`disconnect` supports a single snap form, but with different semantics: if a single colon-separated snap name is provided, whether it's a slot or plug is implicit. Actually, I'm now wondering if my implementation even gets this right -- in the single snap disconnect case, I treat it as the slot snap. I'll have to look into this further.

---

As to the arguments being positional-only, that's why -- hard to have a name that should be used as a keyword argument if the meaning of the argument changes based on the number of arguments.

tonyandrewmeyer: > We've resisted creating a reified `Snap` type elsewhere, so we should probably see how far we can get with that here

Yeah, I'm onboard with that, although I do feel like it would have avoided things I don't like in other PRs too.

> * Two colon-delimited-string positional arguments.

I see why you didn't go with this, and it does feel less Pythonic and more just exposing the snap CLI in a raw way. I do think it's interesting that the API doesn't work this way.

> * Two positional arguments that can either be a string or a `tuple[str, str]`.

I think this is worth considering.

* `connect('somesnap')` (is this possible?)
* `connect('vlc', 'audio-record')`
* `connect(('jhack', 'ssh-read'), 'snapd')` (maybe a bad example since I think snapd is the default, but it's the one I know)
* `connect('foo', ('bar', 'face'))`
* `connect(('foo', 'one'), ('bar', 'two'))` (I don't love this, but probably it is most rare?)

> As to the arguments being positional-only, that's why -- hard to have a name that should be used as a keyword argument if the meaning of the argument changes based on the number of arguments.

Yeah, which is, I would argue, evidence that it's not the right API.

james-garner-canonical: > > We've resisted creating a reified `Snap` type elsewhere, so we should probably see how far we can get with that here
>
> Yeah, I'm onboard with that, although I do feel like it would have avoided things I don't like in other PRs too.

I wouldn't be opposed to revisiting that decision. Maybe it's too late to do it before release, but we could always build a `Snap` class on top of the lower-level functional API later.

> > * Two colon-delimited-string positional arguments.
>
> I see why you didn't go with this, and it does feel less Pythonic and more just exposing the snap CLI in a raw way. I do think it's interesting that the API doesn't work this way.

Yeah the fact that the API explicitly accepts named 'plugs' and 'slots' arguments is a point in favour of handling that differently here.

> > * Two positional arguments that can either be a string or a `tuple[str, str]`.
>
> I think this is worth considering.
>
> * `connect('somesnap')` (is this possible?)
> * `connect('vlc', 'audio-record')`
> * `connect(('jhack', 'ssh-read'), 'snapd')` (maybe a bad example since I think snapd is the default, but it's the one I know)
> * `connect('foo', ('bar', 'face'))`
> * `connect(('foo', 'one'), ('bar', 'two'))` (I don't love this, but probably it is most rare?)

I'll explore this option further, I kind of like it too. With the two argument form, we could use `plug` and `slot` as the names too, and keyword arguments start to look nice again.

> > As to the arguments being positional-only, that's why -- hard to have a name that should be used as a keyword argument if the meaning of the argument changes based on the number of arguments.
>
> Yeah, which is, I would argue, evidence that it's not the right API.

It's not very common, though we do see this in stdlib constructs most developers would be familiar with, like `range` and `slice`. But I'm hopeful that the two argument approach solves things.

One thing I'm not sure about is how to spell things like `[dis]connect(plug=None, slot=(mysnap, myslot))`. Maybe like that?

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: We seem to be consistently using `True` in this library. Should we switch to `true` for bool-typed-but-we-only-test-the-truthiness variables arguments in this library? If so, I'll do it in a separate PR before release.

I've updated the docstring to explain the `forget` behaviour more thoroughly.

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: This has been reworked like `connect`. We now have two, optional, positional-or-keyword arguments (one of which must be provided for a valid call).

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+    """Normalise a plug or slot spec to a ``(snap, name)`` pair of strings."""
+    if spec is None:
+        return '', ''
+    if isinstance(spec, str):
+        return spec, ''
+    snap, name = spec  # ValueError if not a 2-item pair.
+    return snap, name
+
+
+def _raise_if_snaps_not_installed_or_system(plug_snap: str, slot_snap: str) -> None:
```
tonyandrewmeyer, line 161: It seems like this could probably go alongside the singular one in utils, even if it isn't used elsewhere at the moment, or the singular one could just take an iterable rather than have this helper?

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+from __future__ import annotations
+
+from typing import Any
+
+from . import _client, _errors, _utils
+
+# /v2/interfaces
+
+
+def connect(plug: tuple[str, str], slot: tuple[str, str] | str | None = None) -> None:
```
tonyandrewmeyer, line 26: What do you think about using a namedtuple instead of tuple?

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+    """Connect a snap's plug to a slot.
+
+    Connecting an already-connected plug and slot succeeds silently.
+
+    Args:
+        plug: The plug to connect, as a ``(snap, plug)`` pair. Both parts are required:
+            snapd cannot resolve a plug from the snap name alone.
+        slot: The slot to connect to. May be given as:
+
+            - ``None`` (the default) to auto-resolve both the slot's snap and name. Shorthand
```
tonyandrewmeyer: My preference would be to not call out the shorthand. That feels like an implementation detail of snapd that we're allowing to work rather than something people should consider using.

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+              for ``('', '')``. snapd searches for a system snap and matching interface slot
+              for the plug. Raises if the chosen snap has no matching slot, or more than one.
+            - a bare snap name, to auto-resolve the matching slot on that snap (again raising
+              if that snap has no matching slot, or more than one). Shorthand for ``(snap, '')``.
```
tonyandrewmeyer: Also here, although this one doesn't feel as bad.

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+            - a ``(snap, slot)`` pair, to name the slot explicitly. Either part may be an
```
tonyandrewmeyer: Although then this maybe needs to change, and only mention that snap can be the empty string (or false-y?).

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+              empty string to have snapd resolve it as above (an empty snap resolves to the
+              system snap; an empty slot resolves to the matching slot on the given snap).
+
+    Raises:
+        NotFoundError: if the plug snap or slot snap is not installed (the plug snap is checked
```
tonyandrewmeyer: I'm not sure we want or need to say which is checked first, given that either is an error.

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+            first). Not raised for the ``system``/``core`` slot aliases, which snapd serves
+            without the core snap.
+        APIError: if the plug is not fully specified (empty snap or plug name), if the named plug
+            or slot does not exist, if the plug and slot interfaces do not match, or if the slot
+            cannot be resolved unambiguously.
+        ChangeError: if the operation fails after starting (for example, an interface hook errors).
+
+    ::
```
tonyandrewmeyer, line 52: It seems a bit odd to start an example with no test at all, do we do that anywhere else? Also below.

snap/src/charmlibs/snap/_snapd_interfaces.py
```diff
@@ -0,0 +1,180 @@
...
+
+    At least one of ``plug`` or ``slot`` should name something to disconnect. Each is a
+    ``(snap, name)`` pair; unlike :func:`connect`, a bare snap name is not accepted on either
+    side, because snapd requires the plug or slot name to identify what to disconnect. An
+    ``APIError`` is raised if neither side is specified or if either side is partially specified.
+
+    An empty snap on either side means the system snap (mirroring :func:`connect`'s slot): for
+    example ``('', 'mount-observe')`` refers to ``mount-observe`` on ``snapd``/``core``.
+
+    Three forms are supported:
```
tonyandrewmeyer, line 95: Do you think it's worth a "if plug is None and slot is None" raise inside the method?

snap/tests/functional/test_snapd_interfaces.py
```diff
@@ -0,0 +1,434 @@
...
+def test_connect_slot_snap_not_installed_precedes_bad_plug():
+    # Library: the not-installed slot snap wins over the bad plug, surfaced as NotFoundError.
+    with pytest.raises(_errors.NotFoundError) as ctx:
+        _snapd_interfaces.connect((_SNAP, 'nonexistent-plug'), (_ABSENT_SNAP, ''))
+    assert ctx.value.kind == 'snap-not-found'
+    assert _ABSENT_SNAP in str(ctx.value.value)
+    assert 'nonexistent-plug' not in str(ctx.value.value)
+
+
+def test_connect_plug_snap_checked_before_slot_snap():
```
tonyandrewmeyer, line 412: Following on from the code, I'm unconvinced this needs to be defined behaviour.

tonyandrewmeyer (approved):
Thanks! I like how the API looks now. A few small comments / suggestions, but I'm happy with this landing. I appreciate all the work considering the feedback from last round.

snap/src/charmlibs/snap/_snapd_interfaces.py
tonyandrewmeyer: > We seem to be consistently using `True` in this library. Should we switch to `true` for bool-typed-but-we-only-test-the-truthiness variables arguments in this library? If so, I'll do it in a separate PR before release.

I like the idea that we're consistent in docs across our repos, but it's not a big deal if left as-is, I think. You could get David's opinion? Separate PR if done is definitely fine.

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: A user of this function could already pass a `NamedTuple` of the right shape, so I guess what you're asking is whether we should switch to a proper object with named fields instead of the whole `tuple[str, str] | str | None` mess entirely? Or would you still want to support one or more of those short hand forms as well?

That is, are you proposing one more possibly input form (already possible right now with `NamedTuple` on the user side), or collapsing everything into a single canonical object?

I actually think a canonical object has a lot going for it in terms of conceptual clarity. There are three things that make me hesitate:
1. What on earth do we call it (`Endpoint`?), and what on earth do we call its nicely named fields (`snap` and `plug_or_slot`? `plug_or_slot` seems like a definite no, but what else could it be?)?
2. It seems like a shame to add a named type to the API only for these two functions, and if we were to add a named `Snap` type, then that naturally suggests an entire object-oriented API, so the shape would end up being quite different, and library-wide, and I think that would best be considered as a higher-level interface built on top of the lower-level function-based interface.
3. Is it actually a win on the calling ergonomics side? With the recommended library import, you end up repeating `snap` a lot, and it's fairly verbose: `snap.connect(snap.Endpoint('mysnap', 'foo'), snap.Endpoint('anothersnap', 'bar'))`

tonyandrewmeyer: > A user of this function could already pass a `NamedTuple` of the right shape, so I guess what you're asking is whether we should switch to a proper object with named fields instead of the whole `tuple[str, str] | str | None` mess entirely? Or would you still want to support one or more of those short hand forms as well?

I was meaning we would define for people, but the argument would still be like it is now, but docs/examples would use the clearer name. Fair push that people can just do this themselves.

> 1. What on earth do we call it (`Endpoint`?), and what on earth do we call its nicely named fields (`snap` and `plug_or_slot`? `plug_or_slot` seems like a definite no, but what else could it be?)?

I was thinking something like this:

```python
class Connection(NamedTuple):
    snap: str
    slot: str | None = None  # maybe needs to be empty string not None?

...

def connect(plug: tuple[str, str], slot: tuple[str, str] | str | None = None) -> None: ...

...

left = Connection(snap='mysnap', slot='foo')
right = Connection(snap='theirsnap')
connect(left, right)
```

> 2. It seems like a shame to add a named type to the API only for these two functions, and if we were to add a named `Snap` type, then that naturally suggests an entire object-oriented API, so the shape would end up being quite different, and library-wide, and I think that would best be considered as a higher-level interface built on top of the lower-level function-based interface.

I'm less convinced about providing a type only for these two functions, but the "we might build a higher level interface" argument is compelling.

> 3. Is it actually a win on the calling ergonomics side? With the recommended library import, you end up repeating `snap` a lot, and it's fairly verbose: `snap.connect(snap.Endpoint('mysnap', 'foo'), snap.Endpoint('anothersnap', 'bar'))`

I was assuming you'd use these objects other than only in the call, like logging or dynamic construction or whatever. But that was a complete guess not based on looking at any charm snap usage.

No objections to dropping this, it was only a minor suggestion / question.

james-garner-canonical: Thanks for clarifying. `Connection` is a nice name (though it would be misleading to use `slot` as the field for both sides IMO, which is a pain). I'll leave this out for now, but noting it as yet another case where review points towards a higher-level, object-oriented API being a win.

james-garner-canonical: PS: Discussed the `NamedTuple` idea with Ben briefly today and his inclination was to leave it out for now too.

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: I had this originally. I dropped it because this means the same thing to snapd `disconnect(('', ''), ('', ''))` (we translate `None` to `('', '')`), and catching that seemed like going too far, and if we're going to let snapd handle it in one case why not both?

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: We do that in the conf module too. There are no other `::` example blocks in the package.

snap/tests/functional/test_snapd_interfaces.py
james-garner-canonical: Makes sense. I'll retain the test, because we manually probe for installation on error, so it would be good to know if snapd ever changes behaviour.

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: I actually intend to do a pre-release cleanup for something adjacent to this: how we handle `''` for the snap argument elsewhere where it's not meaningful. In one case, this leads to a redirect, because we end up with `//` in the query path, which lead me to an interesting discovery: `urllib`'s redirect handler doesn't trigger on its own, even if explicitly provided as a handler -- it depends on being called by the error handler for 3xx responses, so we have no redirect handling right now.

For this PR, I'll leave this helper in this module.

snap/src/charmlibs/snap/_snapd_interfaces.py
james-garner-canonical: I've reversed the order, so we describe the ``(snap, slot)`` form first. This ends up a lot terser. LMK if you think the empty string behaviour should still be redacted or rearranged.

tonyandrewmeyer (approved):
Re-approving :)

PR #571, opened 2026-07-01
feat(snap): add snap aliases module

james-garner-canonical:
This PR adds the snap library functionality using the `/v2/aliases` endpoint, mirroring the snap CLI commands `snap alias` and `snap unalias`.


snap/tests/functional/test_snapd_aliases.py
```diff
@@ -0,0 +1,166 @@
...
+
+import pytest
+
+from charmlibs.snap import _client, _errors, _snapd_aliases
+from conftest import ensure_installed, ensure_removed
+
+if typing.TYPE_CHECKING:
+    from collections.abc import Mapping
+
+_SNAP = 'lxd'
```
tonyandrewmeyer: I know there's an issue for having dedicated test snaps, but in the meantime, isn't lxd rather large to be using?

snap/tests/functional/test_snapd_aliases.py
```diff
@@ -0,0 +1,166 @@
...
+    """Remove the test alias if it exists, ignoring errors."""
+    try:
+        _snapd_aliases.unalias(_ALIAS)
+    except Exception:  # noqa: S110
+        pass
+
+
+def _alias_exists() -> bool:
+    info = _list_aliases().get(_SNAP, {}).get(_ALIAS)
+    return info is not None and info.get('command') == f'{_SNAP}.{_APP}'
```
tonyandrewmeyer: I guess strictly speaking this is the definition of it existing and working. I wonder if a functional test ought to check the higher level "I can run the command" rather than only "snap db has the value". But maybe that falls out of testing "snap library" and into testing snapd.

snap/tests/functional/test_snapd_aliases.py
```diff
@@ -0,0 +1,166 @@
...
+        _snapd_aliases.alias(_SNAP, 'nonexistent-app', _ALIAS)
+    _cleanup_alias()
+
+
+def test_alias_is_idempotent():
+    # Calling alias() again with the same snap, app, and alias name succeeds silently.
+    ensure_installed(_SNAP)
+    _cleanup_alias()
+    _snapd_aliases.alias(_SNAP, _APP, _ALIAS)
+    _snapd_aliases.alias(_SNAP, _APP, _ALIAS)  # Second call — no error.
```
tonyandrewmeyer, line 98: To really test idempotency this should check it exists after both calls. Otherwise it maybe just worked the second time and failed the first.

snap/tests/functional/test_snapd_aliases.py
```diff
@@ -0,0 +1,166 @@
...
+    _cleanup_alias()
+
+
+def test_alias_reassigns_within_same_snap():
+    # Calling alias() with the same alias name but a different app of the same snap
+    # silently reassigns the alias — no error raised.
+    ensure_installed(_SNAP)
+    _cleanup_alias()
+    _snapd_aliases.alias(_SNAP, _APP, _ALIAS)
+    _snapd_aliases.alias(_SNAP, 'lxd', _ALIAS)  # Different app, same snap — no error.
```
tonyandrewmeyer: I think this should be testing that the alias exists (both times) not just that there is no error.

snap/src/charmlibs/snap/_snapd_aliases.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def unalias(alias_name: str) -> None:
+    """Remove an alias.
+
+    Args:
+        alias_name: The alias to remove.
+
+    Raises:
+        ChangeError: if the unalias change fails after starting.
```
tonyandrewmeyer: Technically correct, but maybe more friendly as

```suggestion
        ChangeError: if the alias removal fails after starting.
```

snap/src/charmlibs/snap/_snapd_aliases.py
```diff
@@ -0,0 +1,59 @@
...
+        APIError: if the alias does not exist (e.g. was never created, or the snap it
```
tonyandrewmeyer:
```suggestion
        APIError: if the alias does not exist (for example, was never created, or the snap it
```

snap/src/charmlibs/snap/_snapd_aliases.py
```diff
@@ -0,0 +1,59 @@
...
+from __future__ import annotations
+
+import logging
+
+from . import _client
+
+logger = logging.getLogger(__name__)
+
+
+def alias(snap: str, app: str, alias_name: str) -> None:
```
tonyandrewmeyer: What about dropping "alias"? It's there in the function name and pretty implied.

```suggestion
def alias(snap: str, app: str, name: str) -> None:
```

Or, simplifying in a different way, but maybe more explicit even though it keeps the duplication:

```suggestion
def alias(snap: str, app: str, alias: str) -> None:
```

(Maybe linters don't like the shadow there, but it's not like this is recursive or ever would be.)

snap/tests/functional/test_snapd_aliases.py
james-garner-canonical: Good call, switched to a light weight test snap like we did for conf.

snap/tests/functional/test_snapd_aliases.py
james-garner-canonical: Good point, added "`which` finds the alias" to the checks.

snap/src/charmlibs/snap/_snapd_aliases.py
james-garner-canonical: Linters don't seem to mind, so I went with `alias`.

tonyandrewmeyer (approved):
Thanks for addressing things from earlier. All looks good to me now.

PR #579, opened 2026-07-02
ci: hash-pin actions and drop zizmor config

tonyandrewmeyer:
## Summary

- Pin every third-party GitHub Actions reference to a full commit SHA, with the original tag kept as a trailing comment for readability.
- Drop `.github/zizmor.yaml`. It only existed to relax `unpinned-uses` to `ref-pin` for `actions/*`, `github/*`, and `pypa/*`; with hash pins in place the default policy is satisfied.

## Test plan

- [x] `zizmor .github/` locally: no findings.
- [x] CI zizmor job stays green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -27,7 +27,7 @@ jobs:
       interfaces: ${{ steps.interfaces.outputs.result }}
       publish: ${{ steps.publish.outputs.result }}
     steps:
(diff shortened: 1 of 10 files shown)


tonyandrewmeyer:
The tests consistently fail, but it doesn't seem related. Let me know if I need to do something here.

james-garner-canonical (approved):
Sorry, I think I just introduced conflicts by merging a Dependabot version bump PR. If you can resolve that in favour of the bump (but with the has), that would be great, thanks.

PR #584, opened 2026-07-10
feat(forward-auth): migrate `forward-auth` lib

natalian98:
This PR migrates the `forward-auth` interface library from [charmhub](https://charmhub.io/oauth2-proxy-k8s/libraries/forward_auth) into charmlibs.

## Library being migrated

- **Charmhub-hosted library:** 
  - **LIBAPI:** 0
  - **LIBPATCH:** 2
  - **LIBID:** e49a5628b4af489cb6e55a8f35974199
- **New `charmlibs` package:** `charmlibs.interfaces.forward_auth`

This is an interface library (`charmlibs.interfaces.<name>`, `just init --interface`).

## Migration status

This is a bug compatible migration of the Charmhub-hosted library, releasing as version 1.0.0.

Package:
- [x] Package initialised with `just init` or `just init --interface`.
- [x] Code migrated to `src/charmlibs/<name>/_<name>.py`.
- [x] Public API exported from `__init__.py` with `__all__`.
- [x] Charmhub lib docstring moved to `__init__.py` (this is rendered in the docs).
- [x] Charmhub-hosted `LIBAPI` and `LIBPATCH` version documented in the migrated module's docstring.
- [x] `LIBID`, `LIBAPI`, `LIBPATCH` removed (or retained with a note on why).
- [x] `PYDEPS` moved to `pyproject.toml` `dependencies` with appropriate constraints.
- [x] Library version set for release (typically `1.0.0`).

Repository metadata:
- [x] `.docs/reference/libs.yaml` updated with entries for new and old libs.
- [x] `CODEOWNERS` updated with a `/<package>/` entry for the owning team.

Tests and docs:
- [x] Unit tests migrated, plus functional and integration tests as appropriate.
- [x] Unnecessary files created by `just init` have been removed (including `test_version.py` and `tests/functional` and `tests/integration` if unused).
- [x] Diataxis docs (if any) migrated to `<package>/docs/`.

### Interface library specific items

- [x] Directory name exactly matches the interface name as written in `charmcraft.yaml`.
- [x] Interface metadata added (or updated if necessary), or an issue created and tracked to do this as a follow-up task.
- [ ] Testing package added under `interfaces/<name>/testing/` exporting `relation_for_provider` and `relation_for_requirer` if needed (see [how-to guide](https://canonical.com/juju/docs/charmlibs/how-to/provide-relation-data-for-charm-tests/)), or an issue created and tracked to do this as a follow-up task.

## Commit strategy

To make review easier, this PR is organized into focused, semantic commits:

  •  feat: migrate `forward_auth` interface: Unified migration of the upstream library. Contains the core scaffolding, verbatim baseline files (library source, tests, and documentation), updated imports, and linter-enforced adjustments in a single coherent step.
  •  chore: update `CODEOWNERS`: Registry of monorepo code ownership for the `forward_auth` interface directory.
  •  chore: update `forward_auth` lib metadata: Registering the package in the monorepo's global configuration indices (`interfaces/index.json `and `.docs/reference/libs.yaml`).


interfaces/forward_auth/src/charmlibs/interfaces/forward_auth/_forward_auth.py
```diff
@@ -0,0 +1,553 @@
...
+
+        if not isinstance(ingress_app_names, ForwardAuthRequirerConfig):  # pyright: ignore[reportUnnecessaryIsInstance]
+            raise TypeError(f'Unexpected type: {type(ingress_app_names)}')
+
+        try:
+            relation = self.model.get_relation(
+                relation_name=self._relation_name, relation_id=relation_id
+            )
+        except TooManyRelatedAppsError:
+            raise
```
james-garner-canonical: Here (and a few other places) this PR drops manually raising `TooManyRelatedAppsError` after catching, with the message "More than one relation is defined. Please provide a relation_id". The new message (from Ops) would be like "Too many remote applications on some-rel (2 > 1)".

If we can actually handle this with the charm passing a relation ID in this case, it seems like a shame to lose that directive for users. On the other hand, if the Ops message is fine, then this `try/except` is redundant and could just be:
```py
        relation = self.model.get_relation(
            relation_name=self._relation_name, relation_id=relation_id
        )
```

interfaces/forward_auth/src/charmlibs/interfaces/forward_auth/_forward_auth.py
```diff
@@ -0,0 +1,553 @@
...
+        if not relation or not relation.app:
+            return None
+
+        requirer_data = relation.data[relation.app]
+        if not requirer_data:
+            logger.info('No requirer relation data available.')
+            return
+
+        ingress_apps = str(requirer_data.get('ingress_app_names', ''))
```
james-garner-canonical: I believe this was `ingress_apps = requirer_data["ingress_app_names"]` in the Charmhub version.

The behaviour of `get` vs `[...]` is significat: previously we'd `KeyError` on a missing field and send the charm to error status on the relation changed event. Now we'd emit a custom event with an error message instead. This change seems reasonable, but should be documented, and you'll want to consider whether the Charmhub version should have the same behaviour for ease of user migration.

PS: The addition of `str` seems unnecessary since I believe this is the raw Juju relation data (already `str`).

interfaces/forward_auth/src/charmlibs/interfaces/forward_auth/_forward_auth.py
```diff
@@ -0,0 +1,553 @@
...
+
+    ret: dict[str, str] = {}
+    for k, v in data.items():
+        if isinstance(v, (list, dict)):
+            try:
+                ret[k] = json.dumps(v)
+            except (json.JSONDecodeError, TypeError) as e:
+                raise DataValidationError(f'Failed to encode relation json: {e}')
+        else:
+            ret[k] = str(v) if v is not None else ''
```
james-garner-canonical: On Charmhub this is `ret[k] = v`. I believe this change is to satisfy the type checker, since the input is `dict[str, Any]` and the output is `dict[str, str]`.

Since the schemas currently only permit string and array entries, there are no runtime changes right now, but a more conservative change to satisfy the type checker without (in principle) runtime effects would be to make the input type annotation something like `dict[str, list[Any] | str]`.

Ultimately it would probably be cleaner to just JSON encode every field, though we'd have to retain the bare value branch on the *load* side for backwards compatibility with the `decisions_address` field. This type of change is probably best left for a separate PR.

james-garner-canonical (commented):
Thanks for this library migration PR, Natalia, and thank you for being the first to try the new PR template for adding a new library -- I hope it was helpful, and please let me know if you have any feedback on it.

It's great to see this library being migrated. I left a few comments on things that jumped out to me, but I'll let @canonical/identity sign off on the actual code and docs migration.

A few 'book-keeping' things to address:
1. The first link in the PR description is wrong, should be https://charmhub.io/oauth2-proxy-k8s/libraries/forward_auth based on the changes in this PR.
2. The "Testing package added ... or an issue created" checkbox shouldn't be ticked as-is. Let's remove the line item from the PR description for now, as I have yet to add the first real testing package (for TLS Certificates interface).
3. The "Commit strategy" section doesn't match the actual commit history of the PR. That's perfectly OK, but please update the section to describe the actual commit strategy and how best to review the PR changes against upstream.
4. I see that the interface metadata had a Python schema, which this PR drops in favour of retaining the Charmhub-hosted lib's JSON schemas. Using the existing JSON schemas at runtime is the right approach for this migration, and I understand the benefit of having a single source of truth. However, I'd prefer to keep the Python schema for now, and take this opportunity to make it accurately match the runtime behaviour. I think the only necessary change would be dropping `| None` from `ingress_app_names`.

We have plans to make the Python schemas the single source of truth in future and have libraries use them for typed relation data validation (see for example how `k8s_backup_target` defines the schema [in the library](https://github.com/canonical/charmlibs/blob/main/interfaces/k8s_backup_target/src/charmlibs/interfaces/k8s_backup_target/_schema.py) and [symlinks to it for the interface metadata](https://github.com/canonical/charmlibs/blob/main/interfaces/k8s_backup_target/interface/v0/schema.py)). I can open a follow-up PR later in the cycle for this library, after we nail down the right approach.

interfaces/forward_auth/src/charmlibs/interfaces/forward_auth/_forward_auth.py
natalian98: Agreed and updated. I've modified `_dump_data` input type signature to `dict[str, list[Any] | str]`, which satisfies pyright while leaving the original  `ret[k] = v`. I'll keep the rest identical to the Charmhub version.

interfaces/forward_auth/src/charmlibs/interfaces/forward_auth/_forward_auth.py
natalian98: Good catch, thank you. I reverted this to use direct lookup `requirer_data['ingress_app_names']` and retain the KeyError and charm error status.

interfaces/forward_auth/src/charmlibs/interfaces/forward_auth/_forward_auth.py
natalian98: Fixed - I restored the original error catching

natalian98:
Thanks a lot @james-garner-canonical for the detailed review. I addressed all comments and updated the PR.

james-garner-canonical (approved):
Thanks for addressing my feedback, this looks good to go from my side. I've configured trusted publishing on PyPI so the package should publish cleanly on merge. We just need a review from `@canonical/identity` as `CODEOWNERS` for the migrated code and tests themselves.

nsklikas (approved):

natalian98:
@james-garner-canonical I rebased the branch and got an approval from my team. I think this PR is good to go.

james-garner-canonical:
> @james-garner-canonical I rebased the branch and got an approval from my team. I think this PR is good to go.

Thanks @natalian98, merging. For the other PRs, if you can please merge rather than rebase and avoid force pushing, then that will make it easier to see what has changed since previous reviews :pray:

PR #587, opened 2026-07-13
ci: add just check-codeowners

### LINKED ISSUE #548: Ensure CODEOWNERS entries for all packages
Every package directory should have an explicit code owner (not just falling back to the repository maintainers). It's currently easy to forget to add an entry, or to end up with an orphan entry when renaming a directory. CI should ensure that every directory has an entry, and that every entry corresponds to a real path in the repository.

The PR templates could drop the `CODEOWNERS` checklist item if we automate this.

james-garner-canonical:
This PR adds `just check-codeowners`, which ensures that:
1. All top-level files and directories have explicit owners (and likewise for children of `interfaces/`).
2. All codeowners entries correspond to a real, tracked file or directory.

CI currently fails due to four unowned interfaces. Blocked by landing PRs to add codeowners for these (TODO):
- `ingress` and `ingress_per_unit` should presumably be `@canonical/observability`
- `spark_service_account` and `zookeeper` seem to be `@canonical/data-bigdata`

Resolves #548.

### DIFF
--- a/.scripts/check_codeowners.py
+++ b/.scripts/check_codeowners.py
@@ -21,12 +21,15 @@
 
 """Validate the repository's CODEOWNERS file.
 
-Two checks are performed:
+Two checks are performed against every top-level path and every immediate child of `interfaces/`:
 
-1. Every package and interface directory has an explicit code owner, rather than
-   falling back to the repository maintainers via the catch-all `*` entry.
-2. Every path-based CODEOWNERS entry corresponds to a real path in the repository,
-   so renaming or removing a directory can't leave behind an orphan entry.
+1. Each path is owned, meaning either the path itself has a CODEOWNERS entry with an owner, or
+   (for a directory) every one of its immediate children is owned. This ensures ownership isn't
+   left to fall through to the repository maintainers via the catch-all `*` entry, while still
+   allowing a directory to be covered by per-child entries (such as an interface's `interface/`
+   plus `ruff.toml`).
+2. Every path-based CODEOWNERS entry corresponds to a real path in the repository, so renaming
+   or removing a directory can't leave behind an orphan entry.
 
 Exit with success (0) if both checks pass, otherwise print the problems to stdout
 and exit with failure (the number of problems found).
@@ -34,7 +37,6 @@ and exit with failure (the number of problems found).
 
 from __future__ import annotations
 
-import json
 import pathlib
 import subprocess
 import sys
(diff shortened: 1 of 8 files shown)


.scripts/check_codeowners.py
```diff
@@ -0,0 +1,130 @@
...
+    return {k: sorted(v) for k, v in children.items()}
+
+
+def check(
+    entries: dict[pathlib.Path, str], children: dict[pathlib.Path, list[pathlib.Path]]
+) -> list[str]:
+    """Return a list of problems found between the CODEOWNERS `entries` and the tracked tree.
+
+    Everything is checked against the tracked tree (`children`) rather than the filesystem, so
+    untracked and ignored paths are irrelevant. A path is a tracked directory iff it's a key in
```
tonyandrewmeyer, line 101: I'm always a bit iffy (heh) about using "iff". This is an internal script so it probably doesn't matter, but I wonder how widely people understand it versus thinking someone has typo'd "if".

.scripts/check_codeowners.py
```diff
@@ -0,0 +1,130 @@
...
+                f"No explicit CODEOWNERS for: /{path}/ (and its children aren't all owned)"
+            )
+    # Check that each entry points at a tracked path of the matching kind.
+    all_paths = {child for siblings in children.values() for child in siblings}
+    for target, pattern in entries.items():
+        if target not in all_paths:
+            problems.append(f'CODEOWNERS entry points at a missing path: {pattern}')
+        elif (target in children) != pattern.endswith('/'):
+            problems.append(
+                f'CODEOWNERS entry must have a trailing slash iff it is a dir: {pattern}'
```
tonyandrewmeyer, line 122: Ah, though this is public. I think the one above could be left in place but we could spell it out here.

.scripts/check_codeowners.py
```diff
@@ -0,0 +1,130 @@
...
+def get_tracked_files(root: pathlib.Path | str) -> list[pathlib.Path]:
+    """Return the repo-relative paths of all files tracked by git in `root`."""
+    output = subprocess.check_output(['git', 'ls-files', '-z'], cwd=root, text=True)
+    return [pathlib.Path(file) for file in output.split('\0') if file]
+
+
+def map_children(files: list[pathlib.Path]) -> dict[pathlib.Path, list[pathlib.Path]]:
+    """Map each tracked directory to its sorted immediate children (files and subdirectories).
+
+    The repository root is keyed as `pathlib.Path()` (i.e. `.`). A path is a directory iff it
```
tonyandrewmeyer, line 85: Are we avoiding Latin in this repo?

.scripts/check_codeowners.py
```diff
@@ -0,0 +1,130 @@
...
+
+def main() -> int:
+    """Run both CODEOWNERS checks, printing any problems and returning the problem count."""
+    entries = parse_codeowners((REPO_ROOT / 'CODEOWNERS').read_text())
+    children = map_children(get_tracked_files(REPO_ROOT))
+    problems = check(entries, children)
+    if problems:
+        print('\n'.join(problems))
+    else:
+        print('No problems found with CODEOWNERS :)')
```
tonyandrewmeyer, line 53: I like the old-school smiley to distinguish from the AI favoured emoji ✅

tonyandrewmeyer (approved):
Looks good, and the currently failing cases provide a nice example of what it will look like if you get it wrong.

.scripts/check_codeowners.py
james-garner-canonical: Apparently not very well -- some `ag` searching shows:
```json
{
  "e.g.": 102,
  "i.e.": 16,
  "etc.": 13,
  "verbatim": 5,
  "vice versa": 1,
  "ergo": 1
}
```
But on the bright side, no hits for:
```json
[
  "N.B.",
  "a priori",
  "ad hoc",
  "ad hominem",
  "bona fide",
  "cf.",
  "de facto",
  "de jure",
  "et al.",
  "et cetera",
  "et seq.",
  "ibid.",
  "in situ",
  "in vitro",
  "inter alia",
  "mutatis mutandis",
  "per se",
  "quid pro quo",
  "sic",
  "sine qua non",
  "status quo",
  "viz.",
  "vs."
]
```

.scripts/check_codeowners.py
james-garner-canonical: I'll split this into two types of problems: directory with no `/`, and non-directory with `/`.

PR #596, opened 2026-07-17
ci: drop redundant dependabot security config

james-garner-canonical:
This PR drops the redundant security updates section from our `.github/dependabot.yaml` (redundant with repo-level security scanning settings), silencing [this complaint from CI](https://github.com/canonical/charmlibs/pull/391/checks?check_run_id=87775878832).

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -25,15 +25,3 @@ updates:
       test-deps:
         patterns:
           - "*"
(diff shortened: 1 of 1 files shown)


tonyandrewmeyer (approved):
I think it helps to have an explanatory comment like in the other charm-tech repos, because dependabot's behaviour here is not obvious. But fine doing that in some later revamp.

PR #603, opened 2026-07-22
chore: make rollingops compatible w/ Python 3.10

imanenami:
## Description

Makes minor syntax changes to the `rollingops` lib to make it compatible with Python 3.10, and adoptable for charms on `jammy`.


rollingops/src/charmlibs/rollingops/_common/_utils.py
```diff
@@ -17,7 +17,7 @@
 import logging
 import subprocess
 from collections.abc import Callable
-from datetime import UTC, datetime
+from datetime import datetime, timezone
```
astrojuanlu: Petty request:


```suggestion
import datetime as dt
```

see https://github.com/python/cpython/pull/145315, https://adamj.eu/tech/2019/09/12/how-i-import-pythons-datetime-module/

sinclert-canonical:
This PR will allow us to start using the library in the MySQL 8.0 charm (see [closed PR](https://github.com/canonical/mysql-operators/pull/254)).

Highly appreciated if considered.

rollingops/src/charmlibs/rollingops/_common/_models.py
```diff
@@ -41,7 +41,17 @@
 logger = logging.getLogger(__name__)
 
 
-class OperationResult(StrEnum):
+class _StrEnum(str, Enum):
```
patriciareinoso: wdyt about doing something around

```
try:
    from enum import StrEnum as _StrEnum
except ImportError:
    from enum import Enum

    class _StrEnum(str, Enum):
        def __str__(self) -> str:
            return self.value
```
so that Python 3.11+ users use the real enum.StrEnum, while Python 3.10 uses the fallback ? 

it does not look too pretty tho and I do not know id that would work

imanenami: It's ok I guess, personally not a fan of import branching when the logic is simple.

tonyandrewmeyer: > wdyt about doing something around
> 
> ```
> try:
>     from enum import StrEnum as _StrEnum
> except ImportError:
>     from enum import Enum
> 
>     class _StrEnum(str, Enum):
>         def __str__(self) -> str:
>             return self.value
> ```
> 
> so that Python 3.11+ users use the real enum.StrEnum, while Python 3.10 uses the fallback ?

Drive-by +1 on this. It's very Pythonic, going back to true/name with NameError in 2.2, importing cModule as module, and so on.

imanenami:
Refactored all `datetime` imports based on @astrojuanlu's comment, also changed the `StrEnum` import to a conditional one.

rollingops/src/charmlibs/rollingops/_common/_utils.py
imanenami: Done a whole refactor here: https://github.com/canonical/charmlibs/pull/603/commits/9df3c13212d948528b2f26970e672978a10c5d4f

rollingops/src/charmlibs/rollingops/_common/_models.py
imanenami: Applied the change here: https://github.com/canonical/charmlibs/pull/603/commits/9df3c13212d948528b2f26970e672978a10c5d4f

PR #605, opened 2026-07-23
chore: ignore S311 in tests

james-garner-canonical:
This PR disables the S311 `ruff` security linting rule for library tests (https://docs.astral.sh/ruff/rules/suspicious-non-cryptographic-random-usage/). This rule complains about any use of the `random` module, under the auspices of it not being cryptographically secure. This isn't relevant for tests.

The inciting incident was trying to use `random` to add jitter to a retry in tests. I considered adding an in-line noqa just for that, but we'll never want this warning in tests, so let's just disable it at the repo level.

### DIFF
--- a/passwd/tests/functional/test_passwd.py
+++ b/passwd/tests/functional/test_passwd.py
@@ -75,7 +75,7 @@ def test_remove_nonexistent_user() -> None:
 
 def test_add_user_with_shell() -> None:
     """Verify we can add a user with a specific shell."""
-    assert passwd.add_user(username='bob', shell='/bin/sh')  # noqa: S604
+    assert passwd.add_user(username='bob', shell='/bin/sh')
     user_info = passwd.user_exists(user='bob')
     assert user_info is not None
     assert user_info.pw_shell == '/bin/sh'
--- a/passwd/tests/functional/test_passwd.py
+++ b/passwd/tests/functional/test_passwd.py
@@ -75,7 +75,7 @@ def test_remove_nonexistent_user() -> None:
 
 def test_add_user_with_shell() -> None:
     """Verify we can add a user with a specific shell."""
-    assert passwd.add_user(username='bob', shell='/bin/sh')
+    assert passwd.add_user(username='bob', shell='/bin/sh')  # noqa: S604
     user_info = passwd.user_exists(user='bob')
     assert user_info is not None
     assert user_info.pw_shell == '/bin/sh'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -143,6 +143,8 @@ ignore = [
     "S105",
     # Hard-coded password function argument.
     "S106",
+    # Complains about use of `random`.
+    "S311",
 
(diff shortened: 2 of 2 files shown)


tonyandrewmeyer (commented):
My concern here is that from what I understand, and some very quick local testing, a lower level pyproject.toml cannot *enable* a rule that is *ignored* at the higher level.

If the rules are enabled (or just not ignored) at the top level, then individual libraries can choose to have them apply or not. With this change, all libraries are *forced* to ignore them. That doesn't seem like a good change to me.

If I'm wrong about this please give me steps to reproduce because that would immediately change my mind on the PR.

What if the top level just didn't enable `S`? I think only `F` is on by default, or a bit more with preview, but not `S`. Then each library could continue to make its own decisions, but you'd still get your off-by-default behaviour.

For what it's worth, I feel some `S` rules are clearly incompatible with (pytest) tests, like "no assert", but there are lots of others I would prefer to have on for my testing code.

james-garner-canonical:
We shouldn't not enable S at the top-level, since that would make it opt-in for library code.

We can continue making more targeted ignores to S for testing. Would you consider the rule that triggered this PR to be suitable to ignore in testing for all libraries? https://docs.astral.sh/ruff/rules/suspicious-non-cryptographic-random-usage/

tonyandrewmeyer:
> We shouldn't not enable S at the top-level, since that would make it opt-in for library code.

Is that terrible? What's the repo position on linting, in terms of how much individual projects get to customise, what's on by default via the top level, on by default via templates, and so on?

In general, it feels like "here's a set of rules we like, adjust it however you like" is a nicer experience than "you must use these settings, and can customise these other ones". But maybe we are intending to be more opinionated than that with the code in this repo? I'm not sure I know what the positioning is.

> Would you consider the rule that triggered this PR to be suitable to ignore in testing for all libraries? https://docs.astral.sh/ruff/rules/suspicious-non-cryptographic-random-usage/

I don't particularly love that rule at all. There are so many reasonable reasons to use distributions in the random module that have nothing at all to do with security. I've never really been convinced that there are so many people improperly using the random module for security that it justifies tainting any use at all.

That said, I do think random is often a bad smell in tests, outside of fuzzing and hypothesis testing and the like. I think people often use random in a way that makes tests unintentionally non-deterministic when they meant to parametrise or have some unimportant fixed value. But even then I doubt I would want a lint rule to fire for it.

So I wouldn't object to turning it off for tests, and probably wouldn't object to turning it off for non-test code.

tonyandrewmeyer (approved):
I really appreciate the discussion and willingness to go another way with this, thanks!

james-garner-canonical:
> I really appreciate the discussion and willingness to go another way with this, thanks!

No worries, you raised a great point about how config inheritance works (easy to disable a top-level rule, not easy to opt back into a disabled one (need to opt out of inheritance or override the whole ignores table)), and that there are some `S` rules we would want in tests.

PR #606, opened 2026-07-23
test(snap): lighten store burden for functional tests

james-garner-canonical:
This PR makes two changes to the `charmlibs.snap` functional tests to reduce the burden on the snap store, as we were getting test failures due to rate limiting recently.

1. Cache channel lookups.
2. Implement a retry with exponential backoff and jitter (to hopefully decorelate parallel runs) for operations that hit the store.

Includes #605 temporarily to avoid linting failures.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -139,19 +139,14 @@ ignore = [
     # All documentation linting.
     "D",
 
-    # Hard-coded password string.
-    "S105",
-    # Hard-coded password function argument.
-    "S106",
+    # All security linting.
+    "S",
 
     # "Useless" expression.
     "B018",
 
     # Redefinition of unused (we import fixtures)
     "F811",
-
-    # Setting a permissive mask on a file or directory
-    "S103",
 ]
 
 [tool.ruff.lint.pydocstyle]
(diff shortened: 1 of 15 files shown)


pyproject.toml
```diff
@@ -139,19 +139,14 @@ ignore = [
     # All documentation linting.
     "D",
 
-    # Hard-coded password string.
```
tonyandrewmeyer, line 142: Looks like https://github.com/canonical/charmlibs/pull/605 leaking into this PR.

tonyandrewmeyer (approved):
I feel it's not great that our infra can't handle this level of use. It's nice to do a bit of caching and the retries are reasonable, though. Hopefully, this doesn't end up necessary at some point in the future.

I would like the leaked change to be removed (or the other PR merged first, effectively removing it) before merging.

pyproject.toml
james-garner-canonical: Yeah, per PR description:

> Includes https://github.com/canonical/charmlibs/pull/605 temporarily to avoid linting failures.

Will wait for #605.

tonyandrewmeyer: Oops, sorry! I read the rest of it. Maybe agent blindness making me ignore concluding lines in PR bodies.

tonyandrewmeyer: Although I don't like #605 and I do like this PR ... 😄

PR #608, opened 2026-07-24
ci: add TIOBE/TICS weekly workflow across monorepo packages

tonyandrewmeyer:
Weekly TICS scan. Iterates all published packages (apt, nginx_k8s, passwd, pathops, rollingops, snap, sysctl, systemd), runs 'just unit' and 'just combine-coverage' per package, merges per-package coverage.db files into a repo-root .report/coverage.xml, then invokes the TICS action on the whole repo with project=charmlibs.

### DIFF
--- a/.github/workflows/tiobe.yaml
+++ b/.github/workflows/tiobe.yaml
@@ -1,0 +1,69 @@
+name: TIOBE
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '15 3 * * 0'
+
+permissions: {}
+
+jobs:
+  TICS:
+    runs-on: [self-hosted, reactive, amd64, tiobe, noble]
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          persist-credentials: false
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@11f9893b081a58869d3b5fccaea48c9e9e46f990  # v8.3.2
+
+      # We could store per-package coverage from the regular CI runs and merge
+      # them here, but this is cheap to do fresh and keeps the TICS run
+      # self-contained.
+      - name: Generate coverage report across packages
+        run: |
+          set -euo pipefail
+          # Discover packages the same way the rest of CI does — dynamic so
+          # new packages get picked up automatically.
+          mapfile -t packages < <(
+            uv run --no-project --script .scripts/ls.py packages \
+              --exclude-examples --exclude-placeholders \
+            | jq -r '.[]'
+          )
+          for pkg in "${packages[@]}"; do
+            # Let `just` resolve the Python version per package (uses the higher
+            # of DEFAULT_PYTHON and the package's requires-python minimum).
+            uvx --from rust-just just unit "$pkg"
+            uvx --from rust-just just combine-coverage "$pkg"
+          done
+          mkdir -p .report
+          # Merge every per-package coverage-all-*.db into one repo-root
+          # coverage.xml for the TICS action to pick up.
+          mapfile -t dbs < <(find . -path '*/.report/coverage-all-*.db' -type f)
+          if [ ${#dbs[@]} -eq 0 ]; then
+            echo "No coverage databases produced." >&2
+            exit 1
+          fi
+          uv run --with coverage coverage combine --keep "${dbs[@]}"
+          uv run --with coverage coverage xml -o .report/coverage.xml
+
+      # Note that these are dependencies for the tools that the TIOBE action will
+      # run, which is done in the global environment, rather than a venv, as we
+      # do with the unit tests.
+      - name: Install TIOBE and project dependencies
+        run: |
+          set -euo pipefail
+          # Export the monorepo's lockfile so pylint/flake8 can resolve imports
+          # from every package during static analysis.
+          uv export --no-emit-project --frozen --no-hashes > requirements.txt
+          pip install flake8 pylint -r requirements.txt --break-system-packages
+
+      - name: TICS GitHub Action
+        uses: tiobe/tics-github-action@333078bc527399e5916a166bd1a3a788be5a01a9  # v3.12.0
+        with:
+          mode: qserver
+          viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
+          ticsAuthToken: ${{ secrets.TICSAUTHTOKEN }}
+          project: charmtech-charmlibs
+          installTics: true
--- a/.github/workflows/tiobe.yaml
+++ b/.github/workflows/tiobe.yaml
@@ -26,21 +26,16 @@ jobs:
           set -euo pipefail
           # Discover packages the same way the rest of CI does — dynamic so
           # new packages get picked up automatically.
-          mapfile -t packages < <(
-            uv run --no-project --script .scripts/ls.py packages \
-              --exclude-examples --exclude-placeholders \
-            | jq -r '.[]'
-          )
+          mapfile -t packages < <(.scripts/ls.py packages --exclude-examples --exclude-placeholders --no-json)
           for pkg in "${packages[@]}"; do
             # Let `just` resolve the Python version per package (uses the higher
             # of DEFAULT_PYTHON and the package's requires-python minimum).
             uvx --from rust-just just unit "$pkg"
-            uvx --from rust-just just combine-coverage "$pkg"
           done
           mkdir -p .report
-          # Merge every per-package coverage-all-*.db into one repo-root
+          # Merge every per-package coverage db into one repo-root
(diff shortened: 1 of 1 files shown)


.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,69 @@
...
+      # self-contained.
+      - name: Generate coverage report across packages
+        run: |
+          set -euo pipefail
+          # Discover packages the same way the rest of CI does — dynamic so
+          # new packages get picked up automatically.
+          mapfile -t packages < <(
+            uv run --no-project --script .scripts/ls.py packages \
+              --exclude-examples --exclude-placeholders \
+            | jq -r '.[]'
```
james-garner-canonical: `uv run` invocation is redundant with script shebang. `--no-json` flag should eliminate the need for `jq`.
```suggestion
          .scripts/ls.py packages --exclude-examples --exclude-placeholders --no-json
```

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,69 @@
...
+          )
+          for pkg in "${packages[@]}"; do
+            # Let `just` resolve the Python version per package (uses the higher
+            # of DEFAULT_PYTHON and the package's requires-python minimum).
+            uvx --from rust-just just unit "$pkg"
+            uvx --from rust-just just combine-coverage "$pkg"
```
james-garner-canonical: Only needed to combine unit and functional test coverage. Since we're only running unit tests, how about we use their output files?
```suggestion
```

That said, we could perhaps iterate over Python versions, as there may be branches that are only covered on specific versions.

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,69 @@
...
+          done
+          mkdir -p .report
+          # Merge every per-package coverage-all-*.db into one repo-root
+          # coverage.xml for the TICS action to pick up.
+          mapfile -t dbs < <(find . -path '*/.report/coverage-all-*.db' -type f)
```
james-garner-canonical: If we drop `combine-coverage`.
```suggestion
          mapfile -t dbs < <(find . -path '*/.report/coverage-*.db' -type f)
```

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+      # Note that these are dependencies for the tools that the TIOBE action will
+      # run, which is done in the global environment, rather than a venv, as we
+      # do with the unit tests.
+      - name: Install TIOBE and project dependencies
+        run: |
+          set -euo pipefail
+          # Export the monorepo's lockfile so pylint/flake8 can resolve imports
+          # from every package during static analysis.
+          uv export --no-emit-project --frozen --no-hashes > requirements.txt
```
james-garner-canonical: The monorepo lockfile doesn't actually install all the packages, it's just got the `fast-lint` stuff like `ruff` that doesn't need packages installed.

james-garner-canonical (changes requested):
Thanks for this. I've left a few comments.

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,68 @@
+name: TIOBE
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '15 3 * * 0'
```
james-garner-canonical: Why Sunday morning? My concern is that this means an email notification on Sunday -- how about Monday morning? Since we're calibrating it around NZ Monday, maybe it's best to make that explicit with a timezone.
```suggestion
    # min (0-59) / hour (0-23) / day (1-31) / month (1-12) / weekday (SUN-SAT)
    - cron: "14 3 * * MON"
      timezone: "Pacific/Auckland"
```

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,68 @@
...
+          mkdir -p .report
+          # Merge every per-package coverage db into one repo-root
+          # coverage.xml for the TICS action to pick up.
+          mapfile -t dbs < <(find . -path '*/.report/coverage-*.db' -type f)
+          if [ ${#dbs[@]} -eq 0 ]; then
+            echo "No coverage databases produced." >&2
+            exit 1
+          fi
+          uv run --with coverage coverage combine --keep "${dbs[@]}"
+          uv run --with coverage coverage xml -o .report/coverage.xml
```
james-garner-canonical: We're not actually installing anything from the bare `uv run` here. Using `uvx` like this would express the same thing, I think:
```suggestion
          uvx coverage combine --keep "${dbs[@]}"
          uvx coverage xml -o .report/coverage.xml
```

But maybe it's better to use the same pinned `coverage` version that we use in the repo:
```suggestion
          uv run --with-requirements=test-requirements.txt coverage combine --keep "${dbs[@]}"
          uv run --with-requirements=test-requirements.txt coverage coverage xml -o .report/coverage.xml
```
Hopefully in future we'll do this more nicely, and this can be updated at the same time.

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+      # Note that these are dependencies for the tools that the TIOBE action will
+      # run, which is done in the global environment, rather than a venv, as we
+      # do with the unit tests.
```
james-garner-canonical:
```suggestion
      # run, which is done in the global environment, rather than a venv.
```

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - name: Install TIOBE and project dependencies
+        run: |
+          set -euo pipefail
+          # The monorepo lockfile only pins the fast-lint tools, so export each
+          # package's own lockfile instead — that's what lets pylint/flake8
+          # resolve third-party imports during static analysis.
```
james-garner-canonical, line 55:
```suggestion
          # Install the dependencies of all libraries.
          # Works at time of writing, but will break if libraries ever have incompatible deps.
```

.github/workflows/tiobe.yaml
```diff
@@ -0,0 +1,68 @@
...
+            uv export --project "$pkg" --no-emit-project --frozen --no-hashes
+          done | sort -u > requirements.txt
+          pip install flake8 pylint -r requirements.txt --break-system-packages
+
+      - name: TICS GitHub Action
+        uses: tiobe/tics-github-action@333078bc527399e5916a166bd1a3a788be5a01a9  # v3.12.0
+        with:
+          mode: qserver
+          viewerUrl: https://canonical.tiobe.com/tiobeweb/TICS/api/cfg?name=default
+          ticsAuthToken: ${{ secrets.TICSAUTHTOKEN }}
```
james-garner-canonical, line 71: We should add a comment here as to why we use `secrets` without a dedicated environment (IIRC a limitation of the TICS GitHub Action?).

james-garner-canonical (changes requested):
Thanks for addressing my feedback. I have a few more comments I'd like addressed before merging.

james-garner-canonical:
PS: Thanks for doing this, this seems like a great baseline for the repository. We should be able to improve things in future by testing across multiple Python versions and perhaps including the functional test coverage too (as there will be some lines that are only covered by functional tests).

.github/workflows/tiobe.yaml
tonyandrewmeyer: Sunday because it's quiet so probably less ocean boiling. But that's a guess to be fair. And it would leave the email to be there ready for picking up Monday morning. But no objection to making it Monday and making it explicitly NZ, although lots of people in the repo are not located there, of course 😄.

.github/workflows/tiobe.yaml
tonyandrewmeyer: Ok, now?

.github/workflows/tiobe.yaml
tonyandrewmeyer: I did a modified version of this, is it ok?

tonyandrewmeyer: Ah, no I did this as is and did the next one modified :joy:.

.github/workflows/tiobe.yaml
tonyandrewmeyer: I did a modified version of this, is it ok?

.github/workflows/tiobe.yaml
tonyandrewmeyer: Limitation of how we are given the secret. Comment added.

.github/workflows/tiobe.yaml
james-garner-canonical: Thanks for explaining, I hadn't considered that angle for Sunday. It would be nice if we could figure out what can actually be done to minimise environmental impact with respect to scheduling.

The rationale for Pacific/Auckland is that it's probably the earliest timezone we have people working in, plus we're based there. But maybe we should just go all the way and use Pacific/Kiritimati?

james-garner-canonical (approved):
Thanks!

PR #609, opened 2026-07-24
feat(snap): validate and sanitize snap before hitting the API

james-garner-canonical:
This PR makes it an immediate `ValueError` to pass `snap=''` (apart from the special case of interfaces, where this has a special meaning).

Passing `snap=''` is one case where we could get a redirect from snapd, which we'd choke on, trying to parse it as JSON. This is because the `urllib` redirect handler only fires if the error handler calls it, and we don't use the error handler so we can handle errors manually. Rather than try to JSON decode redirects, we now check for them explicitly and raise a `BadResponseError`, as with our input sanitization we don't expect to hit them anymore. If snapd ever does start returning them, we'll quickly learn of it and can update the library accordingly.


tonyandrewmeyer (approved):
Interesting find! I like the choice to expect no redirects.

PR #610, opened 2026-07-24
feat(snap): clean error tracebacks

james-garner-canonical:
This PR reworks how `charmlibs.snap` raises errors when probing for if a snap is installed (which we do to provide cleaner, typed errors for operations that fail if the snap isn't installed, but which snapd doesn't return a nice typed error for). In particular, when probing in an error path, we want to avoid the dreaded and confusing chained traceback with "During handling of the above exception, another exception occurred". 

With this PR we have short tracebacks like:
```
Traceback (most recent call last):
  File "_snapd_conf.py", line 102, in get
    raise error from None
charmlibs.snap._errors.NotFoundError: snap not installed (myapp)
```

To facilitate using snapd's returned errors from our installation probe (get), we expand the `__str__` of `Error` in this PR to include `(value)` if it's a string and not already present in the error message.

### DIFF
--- a/snap/src/charmlibs/snap/_errors.py
+++ b/snap/src/charmlibs/snap/_errors.py
@@ -69,12 +69,9 @@ class Error(Exception):
         return str(self._value)
 
     def __str__(self) -> str:
-        # Surface `value` when it adds information the message doesn't already carry -- most
-        # importantly the snap name in snapd's terse 'snap not installed' error, whose message
-        # omits it. Only non-empty string values are appended, and only when not already present
-        # in the message: structured values (e.g. option-not-found's {'SnapName', 'Key'}) are
-        # already reflected in the message and would be noisy here. The raw value and the full
-        # detail remain available via the `value` property and `repr`.
+        # Surface `value` when it adds information the message doesn't already carry.
+        # Most useful for NotFoundError, with message 'snap not installed' and value '<snap>'.
+        # Skip OptionNotFoundError's non-string value, which is redundant with its message.
         value = self._value
         if isinstance(value, str) and value and value not in self._message:
             return f'{self._message} ({value})'
--- a/snap/src/charmlibs/snap/_errors.py
+++ b/snap/src/charmlibs/snap/_errors.py
@@ -68,6 +68,18 @@ class Error(Exception):
         """
         return str(self._value)
 
+    def __str__(self) -> str:
+        # Surface `value` when it adds information the message doesn't already carry -- most
+        # importantly the snap name in snapd's terse 'snap not installed' error, whose message
+        # omits it. Only non-empty string values are appended, and only when not already present
+        # in the message: structured values (e.g. option-not-found's {'SnapName', 'Key'}) are
+        # already reflected in the message and would be noisy here. The raw value and the full
+        # detail remain available via the `value` property and `repr`.
+        value = self._value
+        if isinstance(value, str) and value and value not in self._message:
+            return f'{self._message} ({value})'
+        return self._message
+
     def __repr__(self) -> str:
         return (
             f'{type(self).__module__}.{type(self).__name__}('
--- a/snap/tests/unit/test_errors.py
+++ b/snap/tests/unit/test_errors.py
@@ -85,5 +85,25 @@ def test_snap_error():
     assert 'the-value' in r
     assert '400' in r
     assert 'Bad Request' in r
-    # str() is just the message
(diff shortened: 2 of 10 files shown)


snap/src/charmlibs/snap/_errors.py
```diff
@@ -68,6 +68,18 @@ def value(self) -> object:
         """
         return str(self._value)
 
+    def __str__(self) -> str:
+        # Surface `value` when it adds information the message doesn't already carry -- most
+        # importantly the snap name in snapd's terse 'snap not installed' error, whose message
+        # omits it. Only non-empty string values are appended, and only when not already present
+        # in the message: structured values (e.g. option-not-found's {'SnapName', 'Key'}) are
```
tonyandrewmeyer:
```suggestion
        # in the message: structured values (such as option-not-found's {'SnapName', 'Key'}) are
```

snap/tests/unit/test_errors.py
```diff
@@ -85,5 +85,25 @@ def test_snap_error():
...
+    # A non-empty string value not already in the message is appended in parentheses -- this is
+    # what surfaces the snap name in snapd's terse 'snap not installed' error.
+    assert s('snap not installed', 'hello-world') == 'snap not installed (hello-world)'
+    # Not appended when it would be redundant: value already present in the message ...
+    assert s('snap "hello-world" is not installed', 'hello-world') == (
+        'snap "hello-world" is not installed'
+    )
+    # ... or empty ...
+    assert s('boom', '') == 'boom'
+    # ... or not a plain string (e.g. option-not-found's {'SnapName', 'Key'} dict, whose contents
```
tonyandrewmeyer:
```suggestion
    # ... or not a plain string (for example, option-not-found's {'SnapName', 'Key'} dict, whose contents
```

tonyandrewmeyer (approved):
This looks like a nice UX improvement.

It feels a bit over-tested to me, but we've had similar discussions before with the v2 work, and it's in line with the rest, so I'm making this comment and nothing else :)

I don't love designing for the walrus operator, but it's internal use and we've discussed that before too, so also just mentioning it and then shutting up.

snap/src/charmlibs/snap/_errors.py
james-garner-canonical: I've done a pass to drop overly verbose comments instead.

snap/tests/unit/test_errors.py
james-garner-canonical: Likewise here.

james-garner-canonical:
> This looks like a nice UX improvement.
>
> It feels a bit over-tested to me, but we've had similar discussions before with the v2 work, and it's in line with the rest, so I'm making this comment and nothing else :)

Thanks, I did a scan through the tests and cut down down an instance of over architecting, but happy with the current coverage.
 
> I don't love designing for the walrus operator, but it's internal use and we've discussed that before too, so also just mentioning it and then shutting up.

I think the comments may have overstated how designed for the walrus this is (corrected now). I originally went with for `error = ...` followed by `if error:`, and I'd keep the same design if we were stuck with that pattern. But `if error := ...:` is such a perfect fit that I changed to that.

---

The one place it comes back to bite us is that I'd like to write:
```py
    if keys is None and not config:
        # NOTE: snapd returns {} for an installed snap with no config and for a missing snap.
        # The CLI returns 'error: snap "foo" has no configuration' for a missing snap.
        # For symmetry with PUT (set/unset), we raise NotFoundError here for a missing snap.
        if error := _utils.check_installed_or_system(snap):
            raise error
```
But then ruff *insists* that this should be written like this ([SIM102](https://docs.astral.sh/ruff/rules/collapsible-if/)), which I don't love (but went with). What particularly annoys me is that interleaved comments exclude this from being automatically reformatted with `ruff check --fix --unsafe-fixes`, but don't exclude the rule from triggering in the first place.
```py
    if keys is None and not config and (error := _utils.check_installed_or_system(snap)):
        raise error
```
I decided to just live with this rule but use parentheses and a comment to make things clearer.
```py
    # Empty result when all config was requested: error if the snap isn't installed.
    if (not config) and (keys is None) and (error := _utils.check_installed_or_system(snap)):
        # NOTE: snapd returns {} for an installed snap with no config and for a missing snap.
        # The CLI returns 'error: snap "foo" has no configuration' for a missing snap.
        # For symmetry with PUT (set/unset), we raise NotFoundError here for a missing snap.
        raise error
```

PR #613, opened 2026-07-27
feat(snap): correctly handle channel and revision together

james-garner-canonical:
This PR corrects the handling of the `channel` and `revision` arguments when installing or refreshing a snap. The channel is not silently ignored by snapd when revision is specified as initially thought, rather the revision is verified to be available on that channel, and that channel is then tracked for future refreshes.

- `snap.install` and `snap.refresh` now allow `channel` and `revision` to both be specified.
- `snap.refresh` is updated to accept the `classic` argument symmetrically with `install`.
- The `ensure_revision` function is dropped and `ensure` now accepts `revision`, and has its internal logic updated accordingly.
- Revisions are always passed to snapd when specified, even when falsey (0 or '').
- Channel normalization is now branch aware.
- `snap.Info` now exposes snapd's `tracking-channel` as `Info.tracking` (the snap CLI's "Tracking" column), rather than snapd's `channel` as `Info.channel` (the non-canonicalised channel requested at install/refresh time, not exposed by the CLI).


snap/src/charmlibs/snap/_functions.py
```diff
@@ -17,85 +17,84 @@
 from . import _errors, _snapd_snaps, _utils
 
 
-def ensure_revision(snap: str, revision: int | str, *, classic: bool = False) -> object:
```
james-garner-canonical, line 20: Folded into `ensure`.

snap/src/charmlibs/snap/_functions.py
```diff
@@ -17,85 +17,84 @@
...
-
-
 def ensure(
-    snap: str, channel: str | None = None, *, classic: bool = False, update: bool = True
+    snap: str,
+    channel: str | None = None,
+    *,
+    revision: int | str | None = None,
+    classic: bool = False,
+    update: bool = True,
```
james-garner-canonical, line 26: Ignored if `revision` is not `None`.

snap/tests/functional/conftest.py
```diff
@@ -72,3 +74,25 @@ def ensure_removed(*snaps: str) -> None:
 def ensure_installed(*snaps: str, channel: str | None = None, classic: bool = False) -> None:
     for snap_name in snaps:
         retry_on_rate_limit(snap.ensure)(snap_name, channel=channel, classic=classic, update=False)
+
+
+@functools.cache  # Cached to avoid repeated store queries.
+def list_channels(snap: str) -> dict[str, _snapd.Info]:
```
james-garner-canonical, line 80: Moved from `test_snapd_snaps` for use in `test_functions` as well.

tonyandrewmeyer (approved):
Looks good -- I don't really have any specific comments.

What's the story with branches? I see a couple of cases in the tests, but otherwise it seems like they're not really part of the API. Is that the intention?

Nice addition of refresh being able to switch confinement. Things like that are ones that I am likely to miss in these reviews, by the way, because my snap knowledge isn't that detailed. Hopefully this is exactly where a comprehensive agent review will pay off.

I wonder about the value of overloads, mostly removed here. It seems like it's probably quite complex to try to describe the combinations, but also the right thing to keep it all in one method. I guess in practice people (or agents) are likely to know the combination they want, and trusting the API or the checks here to reject things at runtime is fine.

james-garner-canonical:
> What's the story with branches? I see a couple of cases in the tests, but otherwise it seems like they're not really part of the API. Is that the intention?

I believe they can be specified in the channel string and everything should just work. I forgot about them when implementing `ensure` originally, so they were mishandled there (I treated `/` in the channel as meaning `<channel>/<risk>`, but it could be `<risk>/<branch>`).

> Nice addition of refresh being able to switch confinement. Things like that are ones that I am likely to miss in these reviews, by the way, because my snap knowledge isn't that detailed. Hopefully this is exactly where a comprehensive agent review will pay off.

Yeah, being able to cross reference with the snapd source code is really helpful for this.

> I wonder about the value of overloads, mostly removed here. It seems like it's probably quite complex to try to describe the combinations, but also the right thing to keep it all in one method. I guess in practice people (or agents) are likely to know the combination they want, and trusting the API or the checks here to reject things at runtime is fine.

Luckily the overloads aren't needed anymore on these functions, since there are no longer forbidden argument combinations. But if we add support for cohorts some day, then they'd conflict with revisions so we'd bring the overloads back.

PR #614, opened 2026-07-28
feat(snap): consistently error on empty

james-garner-canonical:
This PR extends the empty-snap-name validation from #609 to every name-like field in the library. #609 specifically targeted snap names that end up in the endpoint path, like `/v2/snaps/<snap>`, but snap names also end up in the URL in comma-separated query strings.

This additional validation fixes a bug: `get()` should always raise `NotFoundError` if the snap isn't installed, which we enforce by checking the user's requested keys and probing for if the snap is installed when required. This is necessary because snapd returns `{}` for a non-installed snap if no particular keys are requested. We checked for `keys=None` and `keys=[]` but not `keys=['']` or `keys=[' ']`, which snapd interprets the same way, so `snap.get(not_installed_snap, [' '])` returns `{}` instead of raising.

---

The rule this PR applies is **validate client-side what the transport would mangle, and let snapd report what it passes through verbatim**. The `logs` names and `conf.get` keys are sent as one comma-separated query parameter, which snapd parses with `strutil.CommaSeparatedList` — splitting on commas, stripping whitespace, and discarding what's left empty. Anything that doesn't survive that unchanged silently becomes a different request: `logs('')` and `logs(' ')` return every snap's logs rather than failing, `logs('a,b')` queries two snaps, and `logs(' snapd ')` queries `snapd`. Everywhere else — JSON bodies and path segments — snapd sees values exactly as passed and reports them clearly, so those are left to snapd apart from empty and blank, which are always caller errors and which snapd reports late (as an `internal error` inside a `ChangeError`) or unhelpfully.

- **Empty** is rejected everywhere except the interface functions, where snapd treats `''` specially (find the system snap).
- **Blank** (whitespace-only, non-empty) is rejected everywhere, with no exemptions. It is never meaningful, not even in the interface functions, where an *empty* value selects the system snap or asks snapd to resolve that side of the connection.
- **Commas and surrounding whitespace** are additionally rejected for comma-separated query parameters.
- New coverage for fields that had none: alias `app` and `alias`, `unalias`, service names, and config keys.


tonyandrewmeyer (approved):
Looks good. I like the clear definition in the PR body. I've looked over what's here (although somewhat skimming the tests), and it makes sense against that definition. I haven't looked through code not in the PR to see if there are other things that should have also changed. I'm assuming that your review would have found that, or that the overall review you mentioned will find it.

PR #615, opened 2026-07-29
feat(snap): switch to `args: str | Iterable[str] | None` from `*args: str`

james-garner-canonical:
Following the decision made in #572 to use `keys: Iterable[str] | None` rather than `*keys: str`, where `None` means 'all' and `[]` means nothing, this PR revisits the other sites in `charmlibs.snap` where we use `*args`: `logs` and services (`start`, `stop`, `restart`).

In #572 we opted to make passing a `str` for `keys` a `TypeError`, despite `str` being `Iterable[str]`, to avoid silently splitting `keys='key'` into three keys: `'k', 'e', 'y'`. However, the ergonomics of being able to type `snap.logs(mysnap)` instead of being forced to type `snap.logs([mysnap])`, and `snap.start(mysnap, myservice)` rather than `snap.start(mysnap, [myservice])` justifies reversing that decision in this PR. Instead, we treat `keys='key'` as equivalent to `keys=['key']`.

The switch to handling an empty iterable differently from snapd, treating it as a request for nothing rather than silently falling through to a request for everything, necessitates some additional machinery for services, since we now follow conf in converting `AppNotFoundError` into `NotFoundError` for `snap.start('not-installed', 'service')`, and must implement the same semantics for `snap.start('not-installed', [])`.

### DIFF
--- a/snap/src/charmlibs/snap/_snapd_conf.py
+++ b/snap/src/charmlibs/snap/_snapd_conf.py
@@ -33,26 +33,27 @@ logger = logging.getLogger(__name__)
 
 # Getting one config value looks like get(s, [k])[k]. In future we could add a get_one(s) helper.
 # Get with keys=None returns the entire config, following the CLI (get_all is unnecessary).
-def get(snap: str, keys: Iterable[str] | None = None) -> dict[str, Any]:
+def get(snap: str, keys: str | Iterable[str] | None = None) -> dict[str, Any]:
     """Get snap configuration.
 
     Args:
         snap: The name of the snap to read configuration from.
-        keys: Configuration keys to read. Nested options may be accessed with dotted notation,
-            for example ``['server.port']``. If ``None``, the full config is returned as a
-            nested dict. If an empty iterable, an empty dict is returned if the snap is installed.
-            Must not be a bare string.
+        keys: Configuration keys to read, as a single key or an iterable of keys. Nested options
+            may be accessed with dotted notation, for example ``'server.port'``. If ``None``, the
+            full config is returned as a nested dict. If an empty iterable, an empty dict is
+            returned if the snap is installed.
 
     Returns:
         A dict mapping each requested key to its configured value. If all keys are requested
         (keys=None), the entire config is returned as a nested dict (empty if the snap has no
         configuration). If no keys are requested (keys=[]), an empty dict is returned if the snap
-        is installed. Each dotted key queried is returned as a top-level entry.
+        is installed. Each dotted key queried is returned as a top-level entry. A single key
+        passed as a bare string is no different from passing it in a list: the result is still a
+        dict, so reading one value looks like ``get(snap, key)[key]``.
 
     Raises:
         ValueError: if the snap name is empty, blank, or is not a single path segment, or if a
             key is empty, blank, or contains a comma or surrounding whitespace.
-        TypeError: if ``keys`` is a string (must be a non-string iterable of strings, or ``None``).
         NotFoundError: if the snap is not installed. Never raised for ``system`` or ``core``,
             whose configuration is served whether or not the core snap is installed.
         OptionNotFoundError: if a requested key has no value stored in the snap's configuration.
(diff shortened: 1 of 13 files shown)


snap/tests/unit/test_utils.py
```diff
@@ -251,25 +255,60 @@ def test_other_special_characters_are_encoded(self, snap: str, expected: str):
...
+    # 'one value or several' argument normalises it through here, so the rule is stated once.
+    @pytest.mark.parametrize(
+        ('values', 'expected'),
+        [
+            ('abc', ['abc']),
+            ('', ['']),  # Emptiness is the checks' business, not this function's.
+            ([], []),
+            (['a', 'b'], ['a', 'b']),
+            (('a', 'b'), ['a', 'b']),
+            ({'a': 1, 'b': 2}, ['a', 'b']),  # A mapping iterates its keys.
```
tonyandrewmeyer, line 267: The comment feels a little obvious.

snap/tests/unit/test_utils.py
```diff
@@ -251,25 +255,60 @@ def test_other_special_characters_are_encoded(self, snap: str, expected: str):
...
+        ],
+    )
+    def test_as_list(self, values: str | Iterable[str], expected: list[str]):
+        assert _utils.as_list(values) == expected
+
+    def test_generator_is_materialised(self):
+        # The values are iterated more than once -- to validate them and then to send them -- so
```
tonyandrewmeyer: I'm not clear on why we are testing this. It seems like it's just testing the expected behaviour any time you pass a generator to something that wants an iterable.

snap/src/charmlibs/snap/_utils.py
```diff
@@ -153,14 +150,55 @@ def parse_timestamp(timestamp: str) -> datetime.datetime:
...
+
 #############################################################
 # Rejecting values snapd can't use, before making a request #
 #############################################################
 
 
+def raise_if_not_path_segment(snap: str) -> None:
+    """Raise ValueError if a snap name would not be a single, canonical URL path segment.
+
+    Percent-encoding alone is not enough to guarantee that:
```
tonyandrewmeyer, line 179: A bunch of this feels to me more like comments than docstring, but since we aren't exposing it, fine leaving it as-is.

snap/src/charmlibs/snap/_snapd_conf.py
```diff
@@ -112,30 +113,31 @@ def get(snap: str, keys: Iterable[str] | None = None) -> dict[str, Any]:
...
 
     Unsetting a key that is not currently set is a no-op and does not raise.
 
     Args:
         snap: The name of the snap to unset configuration on.
-        keys: Configuration keys to unset. Nested options may be addressed with dotted
-            notation, for example ``['server.port']``. Must not be a bare string.
-            An empty iterable is still passed to snapd, and may trigger the snap's config hook.
+        keys: Configuration keys to unset, as a single key or an iterable of keys. Nested options
+            may be addressed with dotted notation, for example ``'server.port'``. An empty
```
tonyandrewmeyer, line 157: It seems like unset is now the odd one out in handling an empty iterable, since it still does an API call. It's called out so users are aware, but I wonder if it would be more consistent if it also short-circuited.

tonyandrewmeyer (approved):
Overall, I like the str|Iterable[str] choice a lot. The convenience for the one-item case is there, and it seems like that would be the most common charm case. And being explicit about passing an iterable rather than varargs just feels cleaner to me, *especially* when there is some other argument (like the snap name) first.

tonyandrewmeyer:
Oh, like the other PR, I'm assuming you'll resolve the conflicts and re-request review if there is a material change.

snap/src/charmlibs/snap/_snapd_conf.py
james-garner-canonical: It *is* consistent with `set(mysnap, {})` but agree that that's not obvious.

`get` and `start`/etc do make an API call on the short-circuit (to `snaps` to check if the snap is installed). `set` and `unset` are different in that their API call not only validates the snap is installed, it gives the snap a chance to complain (e.g. "you can't set config here").

On balance I think this is the right way to go, but happy to consider further if you'd still lean towards short-circuiting.

snap/src/charmlibs/snap/_utils.py
james-garner-canonical: Good point, I'll leave it since it's not public, but a good candidate for a future cleanup

snap/src/charmlibs/snap/_snapd_conf.py
james-garner-canonical: Merging for now, but happy to revisit this before release.

snap/tests/unit/test_utils.py
james-garner-canonical: Dropped.

snap/tests/unit/test_utils.py
james-garner-canonical: Good point, dropped.

PR #616, opened 2026-07-29
feat(snap): add get_one and rename info to list_one

james-garner-canonical:
This PR's primary purpose is to rename `snap.info` to `snap.list_one`, reflecting the CLI semantics that we're following -- `snap.list_one(mysnap)` queries local information about an installed snap (erroring if it's not installed), like `snap list mysnap`. It shouldn't be named `snap.info`, as `snap info` queries the snap store about available snaps instead. This PR doesn't add `snap.list`, as charms likely don't need to query arbitrary lists of snaps. This defers adding support for `snap list --all`, which includes information about inactive snap revisions that are on the system.

To further reduce ambiguity, the `snap.Info` class is renamed to `snap.InstalledInfo`. If we ever added `snap.info` to query the store, it could return a `snap.StoreInfo`, which would have a different structure (e.g. no `tracked` channel, instead a list of available channels). This continues to keep `snap.Snap` free if we add an object-oriented API in future.

Accompanying `list_one`, this PR adds the long-discussed `get_one` accompaniment to `get`. There aren't any other functions in the library that should have a `*._one` version.

### DIFF
--- a/snap/src/charmlibs/snap/__init__.py
+++ b/snap/src/charmlibs/snap/__init__.py
@@ -18,7 +18,7 @@ Use :func:`ensure` to ensure that a snap is installed, optionally on a specific 
 or revision.
 
 Manually manage snap installation with :func:`install`, :func:`refresh`, and :func:`remove`.
-Use :func:`info` to query the current state of an installed snap.
+Use :func:`list_one` to query the current state of an installed snap.
 
 Also manage:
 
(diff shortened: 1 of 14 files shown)


tonyandrewmeyer (approved):
Looks good to me, and the motivation makes sense as per the discussion earlier today.

PR #617, opened 2026-07-29
test(snap): drop heavy snaps from functional tests

james-garner-canonical:
This PR reworks the snap functional tests to use lightweight testing snaps instead of heavy snaps like LXD. The test snaps are either:
- defined locally and fake packed with `mksquashfs`
- snapd's own test snaps, which are published to the store
- in one case, the hello-world snap

### DIFF
--- a/snap/tests/functional/conftest.py
+++ b/snap/tests/functional/conftest.py
@@ -7,10 +7,13 @@
 from __future__ import annotations
 
 import functools
+import json
 import logging
 import random
 import time
 import typing
+import uuid
+from pathlib import Path
 from typing import Any
 
 from charmlibs import snap
@@ -23,6 +26,22 @@ if typing.TYPE_CHECKING:
 
     _P = ParamSpec('_P')
     _T = TypeVar('_T')
+
+# Snaps built from source by setup.sh before the suite runs. Tests prefer these to store snaps
+# wherever they need some capability of a snap rather than a particular snap, so that the suite
+# neither downloads hundreds of megabytes nor depends on a real-world snap keeping its shape.
+SNAPS_DIR = Path(__file__).parent / 'snaps'
+
+# Snaps used by the suite that declare no base of their own, and so are held by the core snap.
+# snapd refuses to remove core while any of them is installed ("snap is being used by snaps ..."),
+# so the tests that remove core must remove these first. Listed here because the snaps are
+# installed by one module and removed by another; a snap that declares a base doesn't belong.
+BASE_LESS_SNAPS = (
+    'hello-world',
+    'test-snap',
+    'test-snapd-classic-confinement',
+    'test-snapd-with-configure',
+)
 
 # Enable debug logging from snap library during tests.
 handler = logging.StreamHandler()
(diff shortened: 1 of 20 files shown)


snap/tests/functional/conftest.py
```diff
@@ -76,6 +87,107 @@ def ensure_installed(*snaps: str, channel: str | None = None, classic: bool = Fa
...
+    no base either but are not held by it. snapd is asked rather than a list being kept here on
+    purpose: which snaps are installed depends on what every other module has done, so a
+    hand-maintained list goes stale as soon as a test starts using a new snap -- and it fails far
+    from the change, as an unrelated core test reporting 'snap is being used by snaps ...'.
+    """
+    snaps = _client.get('/v2/snaps')
+    assert isinstance(snaps, list)
+    snaps = typing.cast('list[dict[str, Any]]', snaps)
+    return [
+        s['name'] for s in snaps if s.get('base') is None and s.get('type') not in _BASE_LIKE_TYPES
```
tonyandrewmeyer: I'm pretty far out of my expertise here, but couldn't this remove a snap of type kernel which seems bad? What about an allowlist instead like

```
return [s['name'] for s in snaps if s.get('type') == 'app' and s.get('base') in (None, 'core')]
```

snap/tests/functional/test_snapd_interfaces_system.py
```diff
@@ -30,7 +30,7 @@
 
 from charmlibs import snap
 from charmlibs.snap import _client, _errors, _snapd_interfaces
-from conftest import ensure_installed, ensure_removed
+from conftest import ensure_installed, ensure_removed, remove_core_blockers
 from test_snapd_local import SNAPS_DIR, install_local
```
tonyandrewmeyer: Missed moving this to import from conftest.

snap/tests/functional/snaps/test-other-service-snap-1.0/bin/daemon
```diff
@@ -0,0 +1,23 @@
...
+#
+# It then ticks slowly forever. Staying alive is the point: a daemon that exits turns every
+# subsequent start into a systemd restart, which is what forced the tests to reset the unit's
+# failure counter between cases when a real snap's service was used.
+set -e
+
+i=0
+while [ "$i" -lt 60 ]; do
+    i=$((i + 1))
+    echo "test-service-snap startup entry $i"
```
tonyandrewmeyer: This and the tick would be better if they had "other" in the name so they aren't exactly like the non-other snap.

snap/tests/functional/test_alive.py
```diff
@@ -17,7 +17,7 @@
 def test_snap_store_reachable():
     # GET /v2/find hits the snap store; a non-empty result confirms both snapd
     # and the store are reachable.
-    result = _client.get('/v2/find', query={'q': 'hello-world'})
+    result = _client.get('/v2/find', query={'q': 'test-snapd-tools'})
```
tonyandrewmeyer: Wouldn't this be better as "name" not "q" like in some other places? If I understand correctly, "q" is a search that could end up finding something else, but "name" is a filter for exactly the name.

tonyandrewmeyer (approved):
I really like this change, thanks! The squashfs stuff is pretty new to me but looks good.

A couple of small comments, but good to go from what I can tell.

I'm assuming you'll resolve the conflicts and nothing material will change.

snap/tests/functional/conftest.py
james-garner-canonical: Good call, fixed.

snap/tests/functional/test_alive.py
james-garner-canonical: Good call, changed.

PR #618, opened 2026-07-30
feat(snap): narrow snap-not-found to NotInstalledError and NotInStoreError

james-garner-canonical:
This PR's most substantial change is making `snap-not-found -> NotFoundError` private and always narrowing this to one of two sub-classes: `NotInStoreError`, or `NotInstalledError` (also mapped from `snap-not-installed`).

Additionally, this PR adds an explicit "not installed" check to the error path of `refresh`, and moves the check for `hold` to the error path (it was an eager pre-check previously).

The documentation for the `classic` argument to `ensure`, `install`, and `refresh` has also been improved, and the semantics of `install` returning falsy when the snap is already installed are now clearly documented (the snap is already installed but on who knows what channel or revision).


snap/src/charmlibs/snap/_errors.py
```diff
@@ -135,16 +151,22 @@ class AppNotFoundError(APIError):
...
     """
 
 
-class NotInstalledError(APIError):
-    """Raised via the API when a snap is not installed on the system."""
+class NotInstalledError(_NotFoundError):
+    """Raised when a snap is not installed on the system."""
+
+
+class NotInStoreError(_NotFoundError):
```
james-garner-canonical, line 165: Hmm, maybe `NotAvailableError`? It *is* funny how similar `InStore` and `Installed` sound though ...

snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -268,6 +282,15 @@ def refresh(
         _client.post(path, body=data)
     except _errors._NoUpdatesAvailableError:
         return False
+    except _errors.APIError as e:
```
tonyandrewmeyer, line 285: Every failed refresh now fires an extra GET /v2/snaps/{snap} before re-raising unchanged. So an extra API call plus if the check fails that swallows the original error. What about something like this?

```python
  except _errors.APIError as e:
      if e.kind and type(e) is not _errors._NotFoundError:
          raise  # snapd classified it; the probe can't add anything.
      if error := _utils.check_installed(snap):
          raise error from None
      if type(e) is _errors._NotFoundError:
          raise _errors.NotInStoreError._from(e) from None
      raise
```

Similar in hold below.

snap/src/charmlibs/snap/_errors.py
```diff
@@ -70,13 +76,23 @@ def value(self) -> object:
...
         # Surface `value` when it adds information the message doesn't already carry.
-        # Most useful for NotFoundError, with message 'snap not installed' and value '<snap>'.
+        # Most useful for _NotFoundError, with message 'snap not installed' and value '<snap>'.
         # Skip OptionNotFoundError's non-string value, which is redundant with its message.
         value = self._value
         if isinstance(value, str) and value and value not in self._message:
             return f'{self._message} ({value})'
         return self._message
 
+    @classmethod
```
tonyandrewmeyer: Very much a nit, but I think having str and repr next to each other and _from either above or below both would be cleaner.

snap/tests/unit/test_not_found.py
```diff
@@ -0,0 +1,102 @@
...
+        _CALLS[name]()
+    except _NotFoundError as e:
+        assert type(e) is not _NotFoundError, (
+            f'{name} let the base _NotFoundError escape: narrow it to NotInstalledError or'
+            f' NotInStoreError, depending on which the operation needed.'
+        )
+
+
+@pytest.mark.parametrize('name', sorted(_CALLS))
+def test_narrowing_preserves_snapds_own_fields(name: str, mock_client: MockClient):
```
tonyandrewmeyer: This and the next one, if the call stops raising for some reason the test will just pass, because it'll never hit the asserts in the except block. Maybe add an else to the try/except that asserts `name =='remove', "did not raise"` kind of thing?

tonyandrewmeyer (approved):
I think the main change is reasonable. If we get requests because charms are frequently catching both we can always go public later.

Rest looks good too. A couple of small comments.

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: I'm reluctant to do this because I'm not sure how seriously snapd guarantees "didn't set `kind` on the error".

snap/src/charmlibs/snap/_errors.py
james-garner-canonical: Good call.

snap/tests/unit/test_not_found.py
james-garner-canonical: Good call, I've rewritten this test suite.

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: I'm going to retain the 'probe first on error' approach because of this concern. I think it's OK because:
1. The probe is cheap, especially compared to the refresh itself (only queries local snapd state).
2. If the probe does fail, the exception is chained so the original is still exposed with a traceback containing "During handling of the above exception, another exception occurred".

PR #619, opened 2026-07-31
feat(snap): rename `ensure` -> `ensure_installed`

james-garner-canonical:
This PR renames `ensure` to `ensure_installed` because it's a better name, matches the convention of `pathops.ensure_contents`, and the 1.0 library has an `ensure` function with subtly different semantics, so it would be nice to avoid accidental collisions with that.

### DIFF
--- a/snap/src/charmlibs/snap/__init__.py
+++ b/snap/src/charmlibs/snap/__init__.py
@@ -14,8 +14,8 @@
 
 """Opinionated library for performing snap operations, targeted at use in charm code.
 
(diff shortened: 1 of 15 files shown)


tonyandrewmeyer (approved):
I agree with the change as discussed at our meeting earlier today - it's unfortunate the the name is a little long now, but the motivation makes sense, and it's not excessively long, particularly for something that probably won't be called many times (and is easily aliased locally).

Everything here seems to be the expected mechanical changes.

PR #620, opened 2026-07-31
fix(snap): correct timestamp timezone parsing, document InstalledInfo

james-garner-canonical:
The goal of this PR is to document the user-facing `InstalledInfo` class, returned by `snap.list_one`. Doing so lead to the following changes:
- Add a `__repr__` (and test that it round-trips).
- Drop unnecessary `normalize_channel` call leftover from when we incorrectly used `'channel'` (whatever the user asked for) rather than `'tracking-channel'` (snapd's normalised value).
- Fix our Python 3.10 timestamp parsing (`fromisoformat` doesn't support snapd's format until 3.11) -- it previously always assumed the timezone was UTC, whereas snapd's timestamps are timezone aware (and reflect where the daemon is running).

I've left Claude's verbose comments for now, as I think they're somewhat helpful in the context of PR review, but I intend to trim them down before merging to avoid comment drift and make the code easier to follow.

### DIFF
--- a/snap/src/charmlibs/snap/_snapd_snaps.py
+++ b/snap/src/charmlibs/snap/_snapd_snaps.py
@@ -46,7 +46,7 @@ class InstalledInfo:
     ):
         self._name = name
         self._classic = classic
-        self._tracking = _utils.normalize_channel(tracking)
+        self._tracking = tracking
         self._revision = str(revision)
         self._version = version
         self._hold = _utils.parse_timestamp(hold) if isinstance(hold, str) else hold
--- a/snap/src/charmlibs/snap/_snapd_snaps.py
+++ b/snap/src/charmlibs/snap/_snapd_snaps.py
@@ -108,13 +108,22 @@ class InstalledInfo:
 
         A held snap is not automatically refreshed, but can be manually refreshed.
 
-        A snap that is held forever is reported by snapd as held for 200+ years,
-        which is hopefully sufficient.
+        Snapd has no distinct value for an indefinite hold, such as the one :func:`hold` places
+        by default: it records the hold as ending after Go's maximum duration, so this is a
+        timestamp roughly 292 years after the hold was placed.
         """
         return self._hold
 
     def __repr__(self) -> str:
-        hold = self.hold if self.hold is None else f"'{self.hold}'"
+        # The hold is rendered as a timestamp string rather than as a datetime repr, which keeps
+        # the repr readable and still lets it be evaluated to reconstruct an equal object: this
+        # is the format __init__ accepts, in the UTC form _utils.parse_timestamp reads on every
+        # supported Python (its fallback for 3.10 doesn't handle an offset).
+        if self.hold is None:
+            hold = None
+        else:
+            utc = self.hold.astimezone(datetime.timezone.utc)
+            hold = f'{utc.isoformat(timespec="microseconds").removesuffix("+00:00")}Z'
         return (
             f'{type(self).__name__}('
             f'{self.name!r}'
@@ -122,7 +131,7 @@ class InstalledInfo:
             f', tracking={self.tracking!r}'
             f', revision={self.revision!r}'
             f', version={self.version!r}'
-            f', hold={hold}'
+            f', hold={hold!r}'
             ')'
         )
 
--- a/snap/src/charmlibs/snap/_snapd_snaps.py
+++ b/snap/src/charmlibs/snap/_snapd_snaps.py
@@ -33,6 +33,8 @@ logger = logging.getLogger(__name__)
 
 
 class InstalledInfo:
+    """Information about an installed snap."""
+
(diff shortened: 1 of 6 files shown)


snap/src/charmlibs/snap/_snapd_snaps.py
```diff
@@ -87,16 +91,41 @@ def tracking(self) -> str:
 
     @property
     def revision(self) -> str:
+        """The snap's revision, as a string.
+
+        Note that locally installed snaps have revisions in the form 'x<N>'.
```
tonyandrewmeyer, line 96: Is that a literal "x"?

snap/tests/functional/test_timestamps.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Snapd marshals a Go time.Time, which encoding/json writes with the RFC3339Nano layout
+# "2006-01-02T15:04:05.999999999Z07:00". The 9s mean trailing zeros are trimmed from the
+# fractional seconds, all the way to no fraction at all for a whole second; the Z07:00 means
+# 'Z' when the time is UTC and an offset such as '+13:00' when it isn't.
+_RFC3339_NANO = re.compile(
+    r'\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}'  # Date and time, always written in full.
+    r'(?:\.\d{1,9})?'  # Fractional seconds, if any survived the trimming.
+    r'(?:Z|[+-]\d{2}:\d{2})'  # Timezone: 'Z' for UTC, otherwise an offset.
+)
```
tonyandrewmeyer, line 42: I think this would probably be clearer with re.VERBOSE, but no objection to leaving it this way.

snap/tests/functional/test_timestamps.py
```diff
@@ -0,0 +1,154 @@
...
+
+
+def test_log_timestamp_format():
+    # /v2/logs timestamps come from journald rather than from snapd's own clock, so they're a
+    # separate format risk: snapd converts them with .UTC(), and journald records microseconds.
+    ensure_installed_local(_SERVICE_SNAP)
+    result = _client.get_logs(query={'n': 5})
+    entries = typing.cast('list[dict[str, Any]]', result)
+    if not entries:
+        pytest.skip('snapd reported no log entries to check the timestamp format of')
```
tonyandrewmeyer: Should this perhaps be a fail not a skip? I'm not sure if people notice skips as much as they should, and so if the ticker snap stops producing logs this would silently (well, other than the skip notice) start skipping.

snap/tests/unit/test_snapd_snaps.py
```diff
@@ -105,6 +106,85 @@ def test_extra_fields_ignored(self):
         assert info.name == 'hello-world'
 
 
+def _public_fields(cls: type) -> list[str]:
```
tonyandrewmeyer, line 109: Isn't all this stuff below covered by a ruff rule?

snap/tests/unit/test_utils.py
```diff
@@ -373,6 +376,180 @@ class TestResolveChannel:
     def test_resolve(self, channel: str, tracking: str, expected: str):
         assert _utils.resolve_channel(channel, tracking) == expected
 
+
+UTC = datetime.timezone.utc
+NZDT = datetime.timezone(datetime.timedelta(hours=13))
+NZST = datetime.timezone(datetime.timedelta(hours=12))
+CHATHAM = datetime.timezone(datetime.timedelta(hours=12, minutes=45))
```
tonyandrewmeyer, line 383: ❤️

tonyandrewmeyer (approved):
I like the datetime parsing approach (although it's regrettable that in the 2020s this sort of thing is still needed).

I'm on board with having the "indefinite (actually just ages) hold" still be a datetime. I wonder if people will want a "def effectively_indefinite(dt: date time.datetime): return (datetime.datetime.now() - dt) > 100 years" kind of thing, but it seems simple enough for them to write themselves for now.

Some small comments, but looks good to me. Quite a lot of additions for something that you'd think would be simple!

snap/tests/unit/test_snapd_snaps.py
james-garner-canonical: The tests these helpers build up to are: (1) all public fields are in the repr, and (2) the repr round-trips (e.g. the timestamp representation parses back to the same datetime). I don't think ruff covers these.

snap/tests/functional/test_timestamps.py
james-garner-canonical: Good catch, changing to `assert entries` since the snap is designed to produce some for us.

snap/src/charmlibs/snap/_snapd_snaps.py
james-garner-canonical: Yeah, like 'x1'.

PR #621, opened 2026-07-31
feat(snap): add `SocketNotFound` + additional error coverage; document error behaviour

james-garner-canonical:
This PR documents the `charmlibs.snap` error behaviour, and:
- adds the `SocketNotFound` error type as a subclass of `ConnectionError`.
- replaces `assert` statements in library code with `raise BadResponseError`.
- wraps the call to `InstalledInfo._from_dict` to raise `BadResponseError` on failure.
- wraps the client's `response.read()` call to convert errors while reading into library errors.
- adds additional error branch coverage to the `opener.open(...)` call to convert errors into library errors.

I've left Claude's verbose comments for now, as I think they're somewhat helpful in the context of PR review, but I intend to trim them down before merging to avoid comment drift and make the code easier to follow.

### DIFF
--- a/snap/tests/unit/test_client.py
+++ b/snap/tests/unit/test_client.py
@@ -140,11 +140,8 @@ _FAILURES = [
 
 @contextlib.contextmanager
 def _snapd_that(*behaviours: str) -> Iterator[str]:
-    """Serve one connection per behaviour on a real unix socket, and yield its path.
-
-    The socket lives directly under the system temp dir rather than pytest's tmp_path, because
-    a unix socket path has to fit in sockaddr_un (108 bytes) and pytest's paths are long.
-    """
+    """Serve one connection per behaviour on a real unix socket, and yield its path."""
+    # pytest's tmp_paths are too long -- a unix socket path has to fit in sockaddr_un (108 bytes).
     with tempfile.TemporaryDirectory() as tmp:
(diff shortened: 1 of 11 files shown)


snap/tests/unit/test_client.py
```diff
@@ -54,6 +64,103 @@ def _fake_response(
...
+    ('reset_mid_body', builtins.ConnectionResetError, 'reading'),
+]
+
+
+@contextlib.contextmanager
+def _snapd_that(*behaviours: str) -> Iterator[str]:
+    """Serve one connection per behaviour on a real unix socket, and yield its path.
+
+    The socket lives directly under the system temp dir rather than pytest's tmp_path, because
+    a unix socket path has to fit in sockaddr_un (108 bytes) and pytest's paths are long.
```
tonyandrewmeyer: Interesting! This is a bit fragile anyway, but I guess we're assuming that most of the time this will be `/tmp` so quite short.

This does seem like a comment not docstring, though.

tonyandrewmeyer (approved):
I'm not entirely sure that this extra work is needed - are issues common enough that the charm wouldn't want to just error out and retry? It seems like a lot of work is needed to try to be exhaustive.

However, since the work has been done, I don't object to including it.

Definitely +1 on trimming down the comments before merging.

PR #622, opened 2026-07-31
feat(snap)!: private error value and kind

james-garner-canonical:
This PR makes the `value` and `kind` fields returned by snapd errors private on our Python error types. Since they're not longer part of the public `Error` contract, they move to private fields on `APIError` (errors snapd returns), which means that other `Error` subclasses don't need to populate artificial `kind` and `value` fields.

Additionally, `BadResponseError` has the `response` field added for the bad response itself -- this was previously exposed as `value`. (But maybe this should be private and just surfaced in the `repr`?).

---

Why wouldn't we want snapd's `value` and `kind` to be public? Because they're very inconsistent -- the same logical error (the snap isn't installed) can have many different snapd kinds depending on which snapd endpoint is raising the error. Conversely, the library resolves these to a single Python error type based on a combination of mapping kind -> error, endpoint specific knowledge, and probing the current state of the snap.

We should make these attributes private before the 2.0 release, becaase we can always commit to them being part of the public API later, but we can't easily make them private later.

Why might we want to make them public in future? `value`, for instance, might hold useful information for a particular error -- for example, being able to query exactly the snap name for a  `NotInstalledError` might be useful. However, if this becomes a feature worth delivering, we can do better than just surfacing `value` -- we can add, for example, a `snap` field to `NotInstalledError` (populated from `value` or call-site knowledge), or `snap`, `revision` and `channel` fields to `RevisionNotAvailable`.

### DIFF
--- a/snap/src/charmlibs/snap/__init__.py
+++ b/snap/src/charmlibs/snap/__init__.py
@@ -58,7 +58,8 @@ never retried: it usually means snapd is not installed on the system.
 
 A :class:`BadResponseError` is raised if the snapd API returns a response the library does not
 understand. Callers will not be able to resolve this error directly, and should report it to the
-library maintainers.
+library maintainers, along with :attr:`BadResponseError.response` -- what snapd sent that the
+library could not read.
 """
 
 from ._errors import (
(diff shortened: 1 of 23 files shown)


tonyandrewmeyer (approved):
Rationale makes sense to me.

After that the changes are almost all the mechanical ones that I'd expect, so looks good, thanks!

